PR #50, opened 2026-02-22
fix(backend): set hatchling package path to app/ for wheel builds

nodesagar:
## Summary

When a new contributor tries to install the backend locally using `pip install -e "."` or `uv sync`, the build fails with a `ValueError` because Hatchling cannot auto-discover the source code.

## Problem

The project name in `pyproject.toml` is `dataloom-backend`. Per Python packaging standards, hyphens are normalized to underscores in package names. Hatchling follows this convention and looks for a source directory named `dataloom_backend/` during auto-discovery. Since the backend source code actually lives in `app/`, the auto-discovery fails with:

```
ValueError: Unable to determine which files to ship inside the wheel...
The most likely cause of this is that there is no directory that matches
the name of your project (dataloom_backend).
```

## Fix

Added an explicit package path configuration to `dataloom-backend/pyproject.toml`:

```toml
[tool.hatch.build.targets.wheel]
packages = ["app"]
```

This tells Hatchling exactly where the source code is, allowing both editable installs and standard wheel builds to succeed. This fix benefits both `pip install -e "."` and `uv sync` workflows, since both rely on Hatchling's build backend to resolve the package.

## How to verify

```bash
cd dataloom-backend
python3 -m venv env
source env/bin/activate
pip install -e "."   # should now succeed
```

## Tested on

- macOS (Apple Silicon, arm64), Python 3.14.2
- Verified `pip install -e "."` completes successfully and the wheel builds without errors

### DIFF
--- a/dataloom-backend/pyproject.toml
+++ b/dataloom-backend/pyproject.toml
@@ -32,9 +32,6 @@ select = ["E", "F", "I", "UP", "B", "SIM"]
 [tool.ruff.format]
 quote-style = "double"
 
-[tool.hatch.build.targets.wheel]
-packages = ["app"]
-
 [build-system]
 requires = ["hatchling"]
 build-backend = "hatchling.build"


nodesagar:
@ivantha I carefully reviewed it again and removed duplicate [tool.hatch.build.targets.wheel] table.

ivantha:
Closing as the fix already exists on main branch.

PR #51, opened 2026-02-22
fix: clear unapplied logs on revert to prevent stale replays on save

### LINKED ISSUE #49: [Bug] Revert doesn't clear unapplied logs — saving after revert replays reverted transformations
**Describe the bug**
Reverting a dataset doesn't clear the unapplied logs from the database. So if you revert and then save, all the old transformations get replayed on the original CSV again — basically undoing your revert.

**To Reproduce**

1. Upload a CSV and open the dataset
2. Add a row 
3. Go to File → View Checkpoints → click Revert
4. Table goes back to original 
5. Now go to File → Save Dataset, type any commit message
6. The row you reverted is back. The save replayed it.

**Expected behavior**
After reverting, saving should have nothing to apply. The reverted changes shouldn't come back.

The revert endpoint (`POST /{dataset_id}/rever`t in `datasets.py`) copies the original CSV back to the working copy, but it doesn't touch the `user_logs` table. The log entries from before the revert are still sitting there with `applied = False.`
Then when you hit save, `save_dataset` does this:

`logs = db.query(models.DatasetChangeLog).filter(
    models.DatasetChangeLog.dataset_id == dataset_id,
    models.DatasetChangeLog.applied == False
).order_by(models.DatasetChangeLog.timestamp).all()
`

The fix is just deleting the unapplied logs during revert, before the commit:

`db.query(models.DatasetChangeLog).filter(
    models.DatasetChangeLog.dataset_id == dataset_id,
    models.DatasetChangeLog.applied == False
).delete()
`

**Environment:**
- OS: macOS
- Browser: Chrome
- Python: 3.13

**Additional context**
Any other relevant information.

Premkumar-2004:
## Description
Revert resets the CSV but doesn't remove unapplied logs from the DB. So when you save after reverting, those old logs get replayed and the reverted changes come back.

Fixed by deleting unapplied logs in the revert endpoint before committing.
Fixes #49 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing
Uploaded a CSV, made changes, reverted, then saved confirmed the reverted changes don't reappear.
## Checklist

- [x] Tests pass locally

### DIFF
--- a/backend/app/api/endpoints/datasets.py
+++ b/backend/app/api/endpoints/datasets.py
@@ -628,10 +628,14 @@ async def revert_to_checkpoint(dataset_id: int, db: Session = Depends(database.g
     # Load the original dataset
     original_path = dataset.file_path.replace('_copy.csv', '.csv')
     df = pd.read_csv(original_path)
-    print("CSV on which applied", original_path)
 
-    # Simply save the original dataset as _copy.csv
     save_dataframe_to_csv(df, dataset.file_path)
+
+    # clear unapplied logs so save doesn't replay them
+    db.query(models.DatasetChangeLog).filter(
+        models.DatasetChangeLog.dataset_id == dataset_id,
+        models.DatasetChangeLog.applied == False
+    ).delete()
 
     db.commit()
 


ivantha (changes requested):
Could you please rebase? The branches are stable now.

ivantha (commented):
## ◎ Argus Review

This PR correctly identifies and addresses a real data-integrity bug: unapplied change logs surviving a revert and being replayed on the next save. The fix is logically sound in intent, but the implementation introduces a **non-atomic sequence** between file I/O and the database mutation that leaves a window for the very bug it is trying to fix, as well as a race condition under concurrent access. Additionally, the deletion scope is overbroad and lacks safeguards (no `synchronize_session`, no error handling). The removal of the stray debug `print` statement is a welcome clean-up. Overall, the approach is on the right track but needs hardening before merging.

ivantha:

ivantha:
Thank you for the contribution! This fix targets the old pre-restructure backend/ directory, and the underlying issue has already been addressed in #167. Please feel free to contribute to the current codebase structure.

PR #60, opened 2026-02-22
fix: block @ prefix in query sanitization to prevent RCE via in-scope…

Premkumar-2004:
## Description

The query sanitization misses the @ prefix which pandas uses to access local variables. So something like `@pd.io.common.os.system('echo PWNED') `slips through the blocklist and runs on the server because pd is in scope. The `local_dict={"__builtins__": {}}` doesn't help since` pd` isn't a builtin.
Added `@ `to the blocked patterns and tests for it.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Tested that `@pd.io.common.os.system(...) `gets blocked after the fix. Checked that normal queries like `col1 > 10 `still work.

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist

- [ ] Tests pass locally


Premkumar-2004:
@charithccmc @Pheewww @ivantha please have a look on this pr

ivantha (changes requested):
Please rebase.

ivantha:
## ◎ Argus Review

This PR addresses a legitimate RCE vulnerability in pandas `df.query()` by adding `@` to the blocklist of dangerous patterns, and simultaneously performs a large-scale codebase restructure (monorepo reorganization, new services/utils layers, full test suite). The core security fix is correct and well-tested, but the overall query sanitization strategy remains a **blocklist/denylist** approach which is fundamentally insufficient for arbitrary code execution prevention. Several additional high-severity issues exist: missing file-size validation, a weak path-traversal check using string comparison, the misleading `local_dict={"__builtins__": {}}` call which provides no actual sandboxing, and a SQLite test database committed to the repository. The architectural restructure is generally positive but introduces its own regressions (Alembic running on every startup, PostgreSQL-specific migrations used against a SQLite test DB, `window.confirm`/`window.prompt` still used despite new dialog components being added).

### Findings (13)

#### 🟠 Blocklist approach is fundamentally insufficient for query injection prevention
**Severity:** high | **Category:** security | **File:** `dataloom-backend/app/utils/security.py:86`

The `_DANGEROUS_PATTERNS` list is a **denylist/blocklist**. This is inherently weaker than an allowlist for security-critical validation. Known bypass vectors not covered include:

- `globals()`, `locals()`, `vars()`, `dir()` — built-in functions that expose namespaces (not blocked)
- Unicode confusables or encoding tricks that may survive regex but be interpreted by the expression evaluator
- `getattr` is not blocked (though most `getattr`-based gadgets need `__import__` which is caught)
- Future pandas versions or numexpr backends may introduce new evaluation paths

The redundant individual patterns (`r'__import__'`, `r'__builtins__'`, `r'__class__'`, etc.) are all already matched by the catch-all `r'__\w+__'` at the end, which adds maintenance confusion.

```python
# NOT blocked:
df.query("globals()")
df.query("vars()")
df.query("locals()")
```

**Suggestion:**
```
Adopt an **allowlist** approach: define a grammar of permitted constructs (column names, comparison operators, literals, boolean connectives) and reject anything outside it. Alternatively, validate that every token in the expression is either a known column name, a numeric/string literal, or a permitted operator. A dedicated expression parsing library is safer than rolling your own regex blocklist.

```python
import ast

ALLOWED_OPERATORS = {ast.Compare, ast.BoolOp, ast.UnaryOp, ast.Num, ast.Str, ast.Constant, ast.Name}

def validate_query_ast(query: str, df_columns: list[str]) -> None:
    try:
        tree = ast.parse(query, mode='eval')
    except SyntaxError as e:
        raise HTTPException(400, f"Invalid query syntax: {e}")
    for node in ast.walk(tree):
        if type(node) not in ALLOWED_OPERATORS:
            raise HTTPException(400, "Query contains disallowed expressions")
        if isinstance(node, ast.Name) and node.id not in df_columns:
            raise HTTPException(400, f"Unknown column: {node.id}")
```
```

---

#### 🟠 `local_dict={"__builtins__": {}}` provides no actual sandbox
**Severity:** high | **Category:** security | **File:** `dataloom-backend/app/services/transformation_service.py:278`

The call `df.query(query_string, local_dict={"__builtins__": {}})` is misleading. In pandas, `local_dict` supplements — it does **not replace** — the expression evaluator's namespace. Passing `{"__builtins__": {}}` merely adds a key named `__builtins__` with an empty-dict value to the local lookup dictionary; it does not clear Python builtins from the evaluation context. The comment in the PR description acknowledges this is ineffective ("doesn't help since `pd` isn't a builtin"), but the call remains, giving a false sense of security.

```python
# Current misleading code:
return df.query(query_string, local_dict={"__builtins__": {}})

# locals() inside the query evaluator are still accessible;
# pd, np, and other in-scope names remain reachable via @
# (now blocked) but the dict itself doesn't sandbox anything.
```

**Suggestion:**
```
Remove `local_dict={"__builtins__": {}}` since it is a no-op security measure and creates false confidence. The real protection should come from the pattern validation. Optionally, look into using the `numexpr` engine explicitly (`engine='numexpr'`) which has a more restricted evaluation environment, but this requires `numexpr` to be installed and its security properties must be independently verified.

```python
def advanced_query(df: pd.DataFrame, query_string: str) -> pd.DataFrame:
    validate_query_string(query_string)
    query_string = query_string.replace("'", '"').strip()
    for col in df.columns:
        if not col.isidentifier():
            query_string = query_string.replace(col, f'`{col}`')
    logger.debug("Executing query: %s", query_string)
    return df.query(query_string)  # rely on validation, not a fake sandbox
```
```

---

#### 🟠 Path traversal check uses fragile string prefix comparison
**Severity:** high | **Category:** security | **File:** `dataloom-backend/app/utils/security.py:55`

The path traversal defense in `resolve_upload_path` uses:

```python
if not str(target).startswith(str(upload_dir)):
    raise HTTPException(status_code=400, detail="Invalid file path")
```

This is vulnerable to a path confusion attack. If `upload_dir` resolves to `/var/app/uploads`, a target resolving to `/var/app/uploads2/evil` would **pass** the check because `'/var/app/uploads2/evil'.startswith('/var/app/uploads')` is `True`. Using `.resolve()` prevents `..` traversal, but the string prefix check is still incorrect.

**Suggestion:**
```
Use `Path.is_relative_to()` (Python 3.9+), which performs a proper path ancestry check:

```python
target = (upload_dir / filename).resolve()

# Correct check (Python 3.9+)
if not target.is_relative_to(upload_dir):
    raise HTTPException(status_code=400, detail="Invalid file path")

# For Python < 3.9 compatibility:
if upload_dir not in target.parents and target != upload_dir:
    raise HTTPException(status_code=400, detail="Invalid file path")
```
```

---

#### 🟠 File size validation missing despite `max_upload_size_bytes` setting
**Severity:** high | **Category:** security | **File:** `dataloom-backend/app/utils/security.py:35`

`validate_upload_file` checks file extension but **never validates file size**, despite the application having a `max_upload_size_bytes` setting (defaulting to 10 MB) and the `.env.example` documenting it. An attacker could upload a multi-gigabyte CSV, causing the server to exhaust memory when pandas reads it with `pd.read_csv()`.

```python
def validate_upload_file(file: UploadFile) -> None:
    settings = get_settings()
    ext = Path(file.filename).suffix.lower()
    if ext not in settings.allowed_extensions:
        raise HTTPException(...)  
    # ← No size check here!
```

**Suggestion:**
```
Read the file size from the upload and enforce the limit:

```python
def validate_upload_file(file: UploadFile) -> None:
    settings = get_settings()
    ext = Path(file.filename).suffix.lower()
    if ext not in settings.allowed_extensions:
        raise HTTPException(
            status_code=400,
            detail=f"File type '{ext}' not allowed."
        )
    # Seek to end to get size, then reset
    file.file.seek(0, 2)
    size = file.file.tell()
    file.file.seek(0)
    if size > settings.max_upload_size_bytes:
        raise HTTPException(
            status_code=413,
            detail=f"File too large. Max: {settings.max_upload_size_bytes} bytes"
        )
```
```

---

#### 🟠 SQLite test database committed to repository
**Severity:** high | **Category:** security | **File:** `dataloom-backend/test.db:1`

The file `dataloom-backend/test.db` — a binary SQLite database — has been committed to the repository. This is problematic for several reasons:

1. **Data leak**: The database may contain actual test data, schema information, or seed data that should not be public.
2. **Repository pollution**: Binary files bloat the repo and create merge conflicts.
3. **Inconsistent test isolation**: Tests are supposed to create/destroy the DB via fixtures (`setup_database`), so committing a DB state breaks reproducibility.
4. **`.gitignore` gap**: The `.gitignore` only ignores `*.sqlite3`, not `*.db` files.

**Suggestion:**
```
1. Remove `test.db` from the repository: `git rm --cached dataloom-backend/test.db`
2. Add `*.db` to `dataloom-backend/.gitignore`:
```
# SQLite databases
*.sqlite3
*.db
```
3. Ensure the `setup_database` fixture in `conftest.py` creates and tears down the database per-test, which it does — so committed state is never needed.
```

---

#### 🟡 Test suite uses SQLite but Alembic migrations are PostgreSQL-specific
**Severity:** medium | **Category:** bug | **File:** `dataloom-backend/tests/conftest.py:8`

Tests configure a SQLite database (`sqlite:///./test.db`) but the application's `lifespan` in `main.py` runs `alembic upgrade head` on startup. The Alembic migrations contain PostgreSQL-specific operations such as:

- `postgresql_using='gen_random_uuid()'`
- `sa.text('gen_random_uuid()')`
- UUID column type (`sa.Uuid`)

When the `TestClient(app)` context manager is entered (which triggers the lifespan), Alembic will attempt to run these migrations against the SQLite engine and **fail**. The tests would break in a fresh environment.

Separately, using `TestClient` as a context manager inside a fixture that's called per-test means Alembic migration runs on every test invocation — very slow.

**Suggestion:**
```
Override the lifespan in tests to skip Alembic and use `SQLModel.metadata.create_all()` directly:

```python
from contextlib import asynccontextmanager
from unittest.mock import patch

@pytest.fixture
def client(db):
    @asynccontextmanager
    async def test_lifespan(app):
        yield  # skip alembic, DB already set up by setup_database fixture

    app.dependency_overrides[get_db] = lambda: (yield db)
    with patch.object(app, 'router') as _:  # or patch lifespan
        with TestClient(app) as c:
            yield c
    app.dependency_overrides.clear()
```
Or more cleanly, separate the lifespan logic from migration logic and make migrations opt-in via an env var.
```

---

#### 🟡 Alembic migrations run on every application startup
**Severity:** medium | **Category:** performance | **File:** `dataloom-backend/app/main.py:20`

The lifespan function runs `alembic upgrade head` every time the application starts:

```python
@asynccontextmanager
async def lifespan(app):
    from alembic import command
    alembic_cfg = Config("alembic.ini")
    command.upgrade(alembic_cfg, "head")  # Runs on EVERY startup
```

In production with multiple Gunicorn/Uvicorn workers, multiple processes will simultaneously attempt to run migrations, risking race conditions and DB locks. This also adds latency to every deployment restart and makes the application responsible for infrastructure concerns.

**Suggestion:**
```
Run migrations as a separate step in the deployment pipeline (CI/CD or Docker entrypoint), not inside the application startup:

```dockerfile
# In Dockerfile or docker-compose command:
CMD ["sh", "-c", "alembic upgrade head && uvicorn app.main:app --host 0.0.0.0 --port 4200"]
```

Or use an environment variable to gate migration execution:

```python
import os
if os.getenv("RUN_MIGRATIONS", "false").lower() == "true":
```
```

---

#### 🟡 Missing lower-bound check for `col_index` in `change_cell_value`
**Severity:** medium | **Category:** bug | **File:** `dataloom-backend/app/services/transformation_service.py:227`

The boundary check in `change_cell_value` only guards against an upper-bound overflow:

```python
if row_index >= len(df) or col_index >= len(df.columns) + 1:
    raise TransformationError("Row or column index out of bounds")

column_name = df.columns[col_index - 1]  # 1-based from frontend
```

If `col_index = 0` is supplied (which passes the check since `0 < len(df.columns) + 1`), then `col_index - 1 = -1`, and Python's negative indexing means `df.columns[-1]` silently returns the **last column** instead of raising an error. This is a silent data corruption bug.

**Suggestion:**
```
Add a lower-bound check:

```python
if row_index >= len(df) or col_index < 1 or col_index > len(df.columns):
    raise TransformationError(
        f"Row index {row_index} or column index {col_index} out of bounds"
    )
column_name = df.columns[col_index - 1]
```
```

---

#### 🔵 Return value of `validate_query_string` silently discarded
**Severity:** low | **Category:** maintainability | **File:** `dataloom-backend/app/services/transformation_service.py:274`

`validate_query_string` is typed to return `str` (the validated query), but the call site in `advanced_query` discards the return value:

```python
# In advanced_query():
validate_query_string(query_string)  # return value thrown away
query_string = query_string.replace("'", '"').strip()  # uses original
```

If someone were to update `validate_query_string` to also sanitize/normalize the query (e.g., strip dangerous characters), the caller would never see those changes. The signature implies transformation but the contract is only validation.

**Suggestion:**
```
Either use the return value, or change the return type annotation to `None` to accurately reflect that it only validates:

```python
# Option A: use the return value
query_string = validate_query_string(query_string)

# Option B: align signature with usage
def validate_query_string(query: str) -> None:  # raises on invalid
    ...
    # don't return the query
```
```

---

#### 🔵 `window.confirm` and `window.prompt` still used despite new dialog components
**Severity:** low | **Category:** bug | **File:** `dataloom-frontend/src/Components/MenuNavbar.jsx:83`

The PR adds `ConfirmDialog` and `InputDialog` components as React-based replacements for native browser dialogs. However, `MenuNavbar.jsx` still uses:

```javascript
const commitMessage = prompt("Enter a commit message for this save:");
// ...
if (window.confirm("Are you sure you want to revert to this checkpoint?")) {
```

Native `window.confirm()` and `prompt()` block the browser's event loop, can't be styled, behave differently across environments (e.g., they're suppressed in iframes), and are generally considered bad UX practice. The new components were created but not integrated.

**Suggestion:**
```
Replace `prompt`/`confirm` with the new `InputDialog`/`ConfirmDialog` components using state:

```jsx
const [saveDialogOpen, setSaveDialogOpen] = useState(false);
const [revertDialogOpen, setRevertDialogOpen] = useState(false);

// In render:
<InputDialog
  isOpen={saveDialogOpen}
  message="Enter a commit message:"
  onSubmit={async (msg) => { setSaveDialogOpen(false); await saveProject(projectId, msg); }}
  onCancel={() => setSaveDialogOpen(false)}
/>
<ConfirmDialog
  isOpen={revertDialogOpen}
  message="Revert to this checkpoint?"
  onConfirm={async () => { setRevertDialogOpen(false); await handleRevert(pendingCheckpointId); }}
  onCancel={() => setRevertDialogOpen(false)}
/>
```
```

---

#### 🔵 Hardcoded credentials in docker-compose.yml and no frontend Dockerfile
**Severity:** low | **Category:** security | **File:** `docker-compose.yml:14`

The `docker-compose.yml` has two concerns:

1. **Hardcoded credentials**: `POSTGRES_PASSWORD: postgres` and the backend `DATABASE_URL` embed a plaintext password. While acceptable for local dev, this gets committed to version control and risks being cargo-culted into production.

2. **Missing frontend Dockerfile**: `frontend: build: dataloom-frontend` references a build context but no `Dockerfile` exists in `dataloom-frontend/`. Running `docker-compose up` will fail with `failed to solve: failed to read dockerfile`.

**Suggestion:**
```
1. Use environment variable substitution for secrets:
```yaml
environment:
  POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-postgres}  # default for dev only
```

2. Add a `dataloom-frontend/Dockerfile`:
```dockerfile
FROM node:20-slim AS builder
WORKDIR /app
COPY package*.json ./
RUN npm ci
COPY . .
RUN npm run build

FROM nginx:alpine
COPY --from=builder /app/dist /usr/share/nginx/html
EXPOSE 3200
```
```

---

#### ℹ️ Positive: `@` prefix correctly blocks the described RCE attack vector
**Severity:** info | **Category:** security | **File:** `dataloom-backend/app/utils/security.py:86`

The core fix is correct. In pandas `df.query()`, the `@` prefix is the mechanism for referencing local variables from the calling scope. The attack `@pd.io.common.os.system('echo PWNED')` works because `pd` (the pandas module) is a local variable in `advanced_query()`, and `@pd` accesses it via the expression evaluator. Adding `r'@'` to the blocklist definitively prevents this class of attack.

The accompanying tests in `test_security.py` (`test_at_prefix_blocked` and `test_at_prefix_read_csv_blocked`) are well-written and verify both the stated attack vector and a path-traversal variant. The test class structure and use of `pytest.raises` is idiomatic.

**Suggestion:**
```
No change needed for the fix itself. Consider adding a code comment explaining **why** `@` is blocked (not just that it is), to help future maintainers:

```python
_DANGEROUS_PATTERNS = [
    # `@var` in pandas query accesses local variables from the caller's
    # scope, allowing e.g. `@pd.io.common.os.system(...)` to execute
    # arbitrary shell commands since `pd` is always in scope.
    r'@',
    ...
]
```
```

---

#### ℹ️ Redundant individual dunder patterns shadowed by catch-all `__\w+__`
**Severity:** info | **Category:** maintainability | **File:** `dataloom-backend/app/utils/security.py:86`

The `_DANGEROUS_PATTERNS` list includes explicit patterns for `__import__`, `__builtins__`, `__class__`, `__subclasses__`, and `__globals__`, all of which are already caught by the final catch-all pattern `r'__\w+__'`. This creates dead code and may give a false impression that these specific patterns need to be listed explicitly.

```python
_DANGEROUS_PATTERNS = [
    r'@',
    r'__import__',      # ← already matched by r'__\w+__'
    r'__builtins__',    # ← already matched
    r'__class__',       # ← already matched
    r'__subclasses__',  # ← already matched
    r'__globals__',     # ← already matched
    ...
    r'__\w+__',  # Catch-all — makes all above redundant
]
```

**Suggestion:**
```
Simplify the list by removing redundant patterns, and add a comment explaining the catch-all:

```python
_DANGEROUS_PATTERNS = [
    r'@',            # local variable access in pandas query (@var)
    r'__\w+__',     # any dunder access (includes __import__, __class__, etc.)
    r'\bexec\b',
    r'\bos\b\s*\.',
    r'\bsys\b\s*\.',
    r'\blambda\b',
    r'\bopen\b\s*\(',
    r'\bcompile\b\s*\(',
]
```
```

---


*◎ Generated by [Argus](https://github.com) — AI-Powered PR Review*

File: dataloom-backend/app/services/transformation_service.py [hunk 1]
```diff
@@ -0,0 +1,343 @@
...
+def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, aggfunc: str = "sum") -> pd.DataFrame:
+    """Create a pivot table from the DataFrame.
+
+    Args:
+        df: Source DataFrame.
+        index: Comma-separated column names for the pivot index.
+        value: Column to aggregate.
+        column: Optional column to pivot on.
+        aggfunc: Aggregation function name.
+
```
ivantha, line 267: **[CRITICAL]** `local_dict` parameter is not valid for `pandas.DataFrame.query()` and provides no protection

The call `df.query(query_string, local_dict={"__builtins__": {}})` passes `local_dict` as a `**kwargs` argument to pandas `query()`. In pandas 2.x, `DataFrame.query()` does **not** accept `local_dict` as a controlling parameter for builtins — it is either silently ignored or forwarded incorrectly to the underlying evaluator. This means the apparent security mitigation provides **zero protection**.

Pandas query expressions are evaluated in the calling frame's scope; the legitimate protection must come from the blocklist in `validate_query_string`, not from `local_dict`. The PR description incorrectly cites this as a defense layer, giving a false sense of security.

**Suggestion:**
```
Remove the `local_dict` kwarg entirely — it is misleading noise. Rely solely on `validate_query_string` for protection:
```python
return df.query(query_string)
```
Add a comment explaining that security is enforced by `validate_query_string` before this call.
```

File: dataloom-backend/app/utils/security.py [hunk 2]
```diff
@@ -0,0 +1,119 @@
...
+
+# Patterns that could be used for code injection via df.query()
+_DANGEROUS_PATTERNS = [
+    r'@',
+    r'__import__',
+    r'__builtins__',
+    r'__class__',
+    r'__subclasses__',
+    r'__globals__',
+    r'\bexec\b',
```
ivantha, line 88: **[HIGH]** Blocking bare `@` breaks legitimate pandas local-variable reference syntax

The pattern `r'@'` blocks **all** occurrences of `@` in query strings. While this prevents the reported RCE vector (`@pd.io.common.os.system(...)`), it also completely breaks the pandas query feature of referencing local variables with `@` prefix (e.g., `df.query("age > @min_age")`). This is documented, intentional pandas functionality.

Furthermore, the blocklist approach can still be bypassed via lesser-known pandas/numexpr evaluation tricks. A more targeted pattern such as `r'@[A-Za-z_]'` (which matches `@` followed by an identifier character) would block the injection while still allowing users to leverage `@variable` references — though the safest approach is to whitelist allowed query patterns or use a proper query parser.

**Suggestion:**
```
Use a more targeted pattern and document the tradeoff:
```python
# Blocks @identifier access (e.g., @pd.io.common.os) while still allowing bare @ if needed
r'@[A-Za-z_]',
```
Alternatively, if local variable references are not a needed feature, document that this feature is intentionally disabled and why.
```

File: dataloom-backend/app/utils/security.py [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+
+    # Defense-in-depth: verify resolved path is inside upload_dir
+    if not str(target).startswith(str(upload_dir)):
+        raise HTTPException(status_code=400, detail="Invalid file path")
+
+    return target
+
+
+# Patterns that could be used for code injection via df.query()
+_DANGEROUS_PATTERNS = [
```
ivantha, line 81: **[HIGH]** Path traversal check uses string prefix matching instead of proper path comparison

The check `if not str(target).startswith(str(upload_dir))` is vulnerable to path prefix collision. For example, if `upload_dir` is `/app/uploads`, a path like `/app/uploads_secret/evil.csv` would pass the check. The correct way is to use `Path.is_relative_to()` (Python 3.9+) which compares actual path components.

Additionally, the check does not account for symlinks — a symlink inside the upload directory could point outside it.

**Suggestion:**
```
```python
# Python 3.9+
if not target.is_relative_to(upload_dir):
    raise HTTPException(status_code=400, detail="Invalid file path")

# For older Python:
if upload_dir not in target.parents and target != upload_dir:
    raise HTTPException(status_code=400, detail="Invalid file path")
```
Also consider resolving symlinks with `target.resolve()` before the check.
```

File: dataloom-backend/app/utils/security.py [hunk 4]
```diff
@@ -0,0 +1,119 @@
...
+        )
+
+
+def resolve_upload_path(filename: str) -> Path:
+    """Resolve a filename to a safe path within the upload directory.
+
+    Defense-in-depth: after constructing the path, verifies the resolved
+    absolute path doesn't escape the upload directory.
+
+    Args:
```
ivantha, line 56: **[HIGH]** File size is not validated despite `max_upload_size_bytes` setting existing

`validate_upload_file()` checks the file extension but **never checks the file size**. The `Settings` class defines `max_upload_size_bytes = 10_485_760` (10 MB) and the `.env.example` documents it, but the validation function ignores it entirely. A malicious user can upload an arbitrarily large CSV file, potentially exhausting disk space or triggering OOM when pandas loads it.

**Suggestion:**
```
```python
def validate_upload_file(file: UploadFile) -> None:
    settings = get_settings()
    ext = Path(file.filename).suffix.lower()
    if ext not in settings.allowed_extensions:
        raise HTTPException(status_code=400, detail=f"File type '{ext}' not allowed.")
    
    # Check size by reading and seeking back
    file.file.seek(0, 2)  # Seek to end
    size = file.file.tell()
    file.file.seek(0)    # Reset
    if size > settings.max_upload_size_bytes:
        raise HTTPException(status_code=413, detail=f"File exceeds maximum size of {settings.max_upload_size_bytes} bytes")
```
```

File: dataloom-backend/app/services/file_service.py [hunk 5]
```diff
@@ -0,0 +1,45 @@
...
+    original_path = resolve_upload_path(safe_name)
+
+    with open(original_path, "wb+") as f:
+        shutil.copyfileobj(file.file, f)
+
+    copy_path = Path(str(original_path).replace('.csv', '_copy.csv'))
+    shutil.copy2(original_path, copy_path)
+
+    logger.info("Stored upload: original=%s, copy=%s", original_path, copy_path)
+    return original_path, copy_path
```
ivantha, line 33: **[HIGH]** `.csv` replacement in path can corrupt paths containing `.csv` in directory names

`Path(str(original_path).replace('.csv', '_copy.csv'))` replaces **all** occurrences of `.csv` in the full path string, not just the file extension. If any parent directory in the path is named something like `csv_data` or `my.csv.files`, this would corrupt the path unpredictably.

Example: `/app/data.csv_archive/file.csv` → `/app/data_copy.csv_archive/file_copy.csv`

**Suggestion:**
```
Use `Path` stem manipulation which only affects the filename:
```python
copy_path = original_path.with_stem(original_path.stem + '_copy')
```
This correctly converts `/uploads/abc123_data.csv` to `/uploads/abc123_data_copy.csv` without touching parent directories.
```

File: docker-compose.yml [hunk 6]
```diff
@@ -0,0 +1,33 @@
+services:
+  db:
+    image: postgres:16
+    restart: unless-stopped
+    environment:
+      POSTGRES_DB: dataloom
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+    ports:
```
ivantha, line 9: **[MEDIUM]** Hardcoded weak database credentials in docker-compose.yml

The `docker-compose.yml` hardcodes `POSTGRES_PASSWORD: postgres` and `POSTGRES_USER: postgres`. While this file is for local development, it establishes a bad pattern and developers may inadvertently use the same credentials in staging or production environments.

**Suggestion:**
```
Use environment variable substitution with a `.env` file (which is git-ignored):
```yaml
environment:
  POSTGRES_DB: dataloom
  POSTGRES_USER: ${POSTGRES_USER:-postgres}
  POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:?Postgres password is required}
```
Document in the README that developers should create a local `.env` file with secure credentials.
```

File: dataloom-frontend/src/Components/MenuNavbar.jsx [hunk 7]
```diff
@@ -0,0 +1,291 @@
...
+  };
+
+  const handleMenuClick = (formType) => {
+    setShowFilterForm(false);
+    setShowSortForm(false);
+    setShowDropDuplicateForm(false);
+    setShowAdvQueryFilterForm(false);
+    setShowPivotTableForm(false);
+    setShowLogs(false);
+    setShowCheckpoints(false);
```
ivantha, line 100: **[MEDIUM]** `window.confirm` and `prompt` still used despite new dialog components existing

The PR introduces `ConfirmDialog` and `InputDialog` components in `src/Components/common/` specifically to replace browser native dialogs. However, `MenuNavbar.jsx` still calls `prompt()` for the commit message and `window.confirm()` for the revert confirmation. This is inconsistent with the new architecture and the native dialogs are blocking, inaccessible, and cannot be styled.

**Suggestion:**
```
Replace the native dialog calls with the new components:
```jsx
const [confirmRevert, setConfirmRevert] = useState(false);
const [showCommitInput, setShowCommitInput] = useState(false);

// In JSX:
<InputDialog
  isOpen={showCommitInput}
  message="Enter a commit message for this save:"
  onSubmit={async (msg) => { setShowCommitInput(false); await doSave(msg); }}
  onCancel={() => setShowCommitInput(false)}
/>
<ConfirmDialog
  isOpen={confirmRevert}
  message="Are you sure you want to revert to this checkpoint?"
  onConfirm={doRevert}
  onCancel={() => setConfirmRevert(false)}
/>
```
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 8]
```diff
@@ -0,0 +1,182 @@
...
+    commit_message: str,
+    db: Session = Depends(database.get_db),
+):
+    """Save project changes as a checkpoint.
+
+    Replays all pending transformations from the change log onto the original
+    file and creates a checkpoint record marking the save point.
+    """
+    project = get_project_or_404(project_id, db)
+
```
ivantha, line 100: **[MEDIUM]** Save endpoint has a race condition between applying logs and creating checkpoint

In the `/save` endpoint, unapplied logs are fetched and replayed, then `create_checkpoint()` is called which separately re-fetches unapplied logs to mark them applied. Between these two operations (in the same request but without a DB-level lock), a concurrent transformation request could insert a new log entry. That new entry would be marked as applied under this checkpoint without having been replayed onto the CSV file, corrupting the checkpoint's state.

Additionally, after `save_csv_safe(df, original_path)`, the working copy at `project.file_path` (the `_copy.csv`) is not updated to match. Subsequent operations on the copy will work from stale data.

**Suggestion:**
```
Wrap the entire operation in a database transaction with row-level locking:
```python
# Lock the project row to prevent concurrent modifications
db.execute(text("SELECT 1 FROM projects WHERE project_id = :id FOR UPDATE"), {"id": project_id})

# ... fetch logs, replay, save CSV ...

# Update both original and copy
save_csv_safe(df, original_path)
shutil.copy2(original_path, project.file_path)  # Keep copy in sync
```
```

File: dataloom-backend/app/main.py [hunk 9]
```diff
@@ -0,0 +1,62 @@
...
+
+@asynccontextmanager
+async def lifespan(app):
+    """Application startup/shutdown lifecycle."""
+    from alembic import command
+    from alembic.config import Config
+
+    alembic_cfg = Config("alembic.ini")
+    command.upgrade(alembic_cfg, "head")
+
```
ivantha, line 28: **[MEDIUM]** Auto-running Alembic migrations on startup is unsafe for production

Running `command.upgrade(alembic_cfg, "head")` in the FastAPI lifespan startup hook means every replica that starts will attempt to run migrations. In a multi-instance production deployment (e.g., Kubernetes with multiple pods), this creates a race condition where multiple instances try to apply the same migrations simultaneously, potentially causing deadlocks or duplicate constraint errors.

Additionally, if a migration is destructive and fails midway, the application will be left in a half-migrated state.

**Suggestion:**
```
Run migrations as a separate pre-deployment step (e.g., a Kubernetes init container or CI/CD step):
```bash
# In your deployment pipeline or Dockerfile CMD
uv run alembic upgrade head && uvicorn app.main:app ...
```
Or use a dedicated management command. If auto-migration is truly desired, add a distributed lock (e.g., using the database itself) to prevent concurrent migration runs.
```

File: dataloom-backend/alembic/versions/b2c3d4e5f6a7_seed_sample_projects.py [hunk 10]
```diff
@@ -0,0 +1,153 @@
...
+
+
+def _write_csv(path: Path, headers: list[str], rows: list[list[str]]) -> None:
+    with open(path, "w", newline="") as f:
+        writer = csv.writer(f)
+        writer.writerow(headers)
+        writer.writerows(rows)
+
+
+def upgrade() -> None:
```
ivantha, line 110: **[MEDIUM]** Seed migration writes CSV files to filesystem during Alembic migration

The `upgrade()` function in the seed migration uses `_write_csv()` to write files to the upload directory during a database migration. This is problematic for several reasons:
1. The upload directory path is environment-specific and may not exist or be writable in all deployment contexts (e.g., CI, Docker build time).
2. Database migrations are intended to be reversible and schema-only; file system side effects make the migration non-idempotent.
3. If the upload directory is ephemeral (mounted volume, S3), the files will not persist across restarts but the DB records will remain, causing 404 errors when accessing seed projects.

**Suggestion:**
```
Separate seed data from schema migrations. Create a dedicated management command or seed script:
```python
# dataloom-backend/app/management/seed.py
def seed_projects(db: Session, upload_dir: Path) -> None:
    for sample in SAMPLE_PROJECTS:
        # write CSVs and create DB records
        ...
```
Call this during initial setup, not during migration.
```

File: dataloom-frontend/src/api/projects.js [hunk 11]
```diff
@@ -0,0 +1,66 @@
...
+ * Save the current project state as a checkpoint.
+ * @param {string} projectId - The project ID.
+ * @param {string} commitMessage - Description of changes.
+ * @returns {Promise<Object>} Updated project response.
+ */
+export const saveProject = async (projectId, commitMessage) => {
+  const response = await client.post(
+    `/projects/${projectId}/save?commit_message=${encodeURIComponent(commitMessage)}`
+  );
+  return response.data;
```
ivantha, line 52: **[MEDIUM]** Query parameters built via string concatenation instead of params object

`saveProject` constructs the URL as a template string: `` `/projects/${projectId}/save?commit_message=${encodeURIComponent(commitMessage)}` ``. While `encodeURIComponent` is used, this approach is fragile. If the API adds more query parameters later, or if the commit message contains characters like `+` or `%` that `encodeURIComponent` doesn't handle the same way as form encoding, subtle bugs can arise. Using axios's `params` option is the idiomatic and safer approach.

**Suggestion:**
```
```javascript
export const saveProject = async (projectId, commitMessage) => {
  const response = await client.post(
    `/projects/${projectId}/save`,
    null,
    { params: { commit_message: commitMessage } }
  );
  return response.data;
};
```
```

File: dataloom-backend/tests/conftest.py [hunk 12]
```diff
@@ -0,0 +1,66 @@
+"""Test configuration and fixtures for the DataLoom backend tests."""
+
+import csv
+import pytest
+from pathlib import Path
+from fastapi.testclient import TestClient
+from sqlmodel import SQLModel, Session, create_engine
+from app.database import get_db
+from app.main import app
+
```
ivantha, line 10: **[LOW]** `test.db` SQLite file committed to the repository

A binary SQLite database file `dataloom-backend/test.db` was committed to the repository (visible in the diff as a new binary file). Test database files should never be committed — they are generated at test runtime, are binary, cause unnecessary diff noise, and may contain sensitive test data.

**Suggestion:**
```
Add `test.db` to `dataloom-backend/.gitignore`:
```gitignore
# SQLite (local dev)
*.sqlite3
test.db
```
Then delete the committed file with `git rm --cached dataloom-backend/test.db`.
```

File: dataloom-backend/app/utils/security.py [hunk 13]
```diff
@@ -0,0 +1,119 @@
...
+    r'\bcompile\b\s*\(',
+    r'__\w+__',  # Catch-all for dunder attributes
+]
+
+
+def validate_query_string(query: str) -> str:
+    """Validate a pandas query string against known injection patterns.
+
+    Blocks dangerous Python constructs that could be exploited through
+    pandas df.query(), which internally uses expression evaluation.
```
ivantha, line 102: **[LOW]** Redundant specific dunder patterns before catch-all `__\w+__` pattern

The `_DANGEROUS_PATTERNS` list contains specific dunder patterns (`__import__`, `__builtins__`, `__class__`, `__subclasses__`, `__globals__`) that are all already covered by the catch-all `r'__\w+__'` pattern at the end. The catch-all makes the specific ones redundant, adding unnecessary complexity and giving the false impression that specific patterns have been individually considered and the catch-all is a bonus.

Furthermore, `r'\bos\b\s*\.'` won't catch `os` accessed as part of a longer expression like `module.os.path`.

**Suggestion:**
```
Simplify to:
```python
_DANGEROUS_PATTERNS = [
    r'@[A-Za-z_]',        # @variable access (e.g., @pd.io.common.os)
    r'__\w+__',          # Any dunder attribute/method
    r'\bexec\b',
    r'\bos\b\s*\.',
    r'\bsys\b\s*\.',
    r'\blambda\b',
    r'\bopen\b\s*\(',
    r'\bcompile\b\s*\(',
]
```
And add a comment explaining each pattern's purpose.
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py [hunk 14]
```diff
@@ -0,0 +1,129 @@
...
+    op.drop_constraint(None, 'user_logs', type_='foreignkey')
+    op.drop_constraint(None, 'user_logs', type_='foreignkey')
+    op.drop_constraint(None, 'checkpoints', type_='foreignkey')
+
+    # Step 2: Revert FK columns back to INTEGER
+    op.alter_column('user_logs', 'checkpoint_id',
+               existing_type=sa.Uuid(),
+               type_=sa.INTEGER(),
+               existing_nullable=True)
+    op.alter_column('user_logs', 'dataset_id',
```
ivantha, line 95: **[LOW]** Downgrade migration drops foreign keys with `None` as constraint name

In the `downgrade()` function, multiple `op.drop_constraint(None, ...)` calls attempt to drop foreign key constraints that were created with `op.create_foreign_key(None, ...)` in the upgrade. Alembic generates auto-names for constraints created with `None`, but dropping them by `None` relies on Alembic correctly resolving the auto-generated name, which may not be deterministic across different versions of Alembic or PostgreSQL. This can cause the downgrade to fail silently or error.

**Suggestion:**
```
Assign explicit constraint names during creation and use those names for deletion:
```python
# In upgrade:
op.create_foreign_key('fk_checkpoints_project_id', 'checkpoints', 'datasets', ['dataset_id'], ['dataset_id'])

# In downgrade:
op.drop_constraint('fk_checkpoints_project_id', 'checkpoints', type_='foreignkey')
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 15]
```diff
@@ -0,0 +1,343 @@
...
+        index = action_details['row_params']['index']
+        if index < 0 or index >= len(df):
+            raise TransformationError(f"Row index {index} out of range")
+        return df.drop(index)
+
+    elif action_type == 'addCol':
+        index = action_details['col_params']['index']
+        column_name = action_details['col_params']['name']
+        return add_column(df, index, column_name)
+
```
ivantha, line 320: **[INFO]** `apply_logged_transformation` uses different `delRow` logic than `delete_row()`

The `apply_logged_transformation` function has its own inline implementation of `delRow` rather than calling the existing `delete_row()` function:
```python
elif action_type == 'delRow':
    index = action_details['row_params']['index']
    if index < 0 or index >= len(df):
        raise TransformationError(...)
    return df.drop(index)  # Note: no reset_index!
```
Meanwhile, `delete_row()` uses `df.drop(index).reset_index(drop=True)`. The missing `reset_index` in the replay path means that after a save/replay, the DataFrame index may not be contiguous, which can cause subtle bugs in subsequent index-based operations.

**Suggestion:**
```
Reuse existing functions consistently:
```python
elif action_type == 'delRow':
    return delete_row(df, action_details['row_params']['index'])
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR bundles a legitimate security fix (blocking `@` prefix in pandas query strings to prevent RCE via `@pd.io.common.os.system()`) with a massive, undisclosed codebase restructuring that includes breaking API/schema changes, a database migration destroying existing relationships, and hundreds of new files. The security fix itself has two significant flaws: the `local_dict={"__builtins__": {}}` parameter passed to `df.query()` is not a valid security control and provides no protection, and the bare `@` blocklist completely disables the legitimate pandas local-variable reference syntax. Additionally, a committed `test.db` binary, missing file-size validation, a path traversal check using string prefix matching, and hardcoded Docker credentials are among the other issues found. The PR description dramatically understates the scope of changes, making it unsafe to merge without a thorough independent review of the restructuring component.

### ⚠️ Breaking Changes

- **[HIGH]** API routes changed from `/datasets/` to `/projects/`: **Previous behavior**: All dataset CRUD and transformation endpoints were mounted under `/datasets/` prefix.

**New behavior**: All endpoints are now under `/projects/` prefix. For example: `POST /datasets/upload` → `POST /projects/upload`, `POST /datasets/{id}/transform` → `POST /projects/{id}/transform`.

**Who is affected**: Any existing API clients, frontend code, integration tests, or external consumers that call the old `/datasets/` endpoints will receive 404 errors after this change.

**Migration path**: Update all API calls to use the `/projects/` prefix. The frontend code in this PR has been updated, but any external consumers (scripts, integrations, other services) must be updated manually.
- **[HIGH]** Response schema field `dataset_id` renamed to `project_id`: **Previous behavior**: All API responses included a `dataset_id` field (integer type) in `DatasetResponse` and `BasicQueryResponse`.

**New behavior**: Responses now use `project_id` (UUID type) in `ProjectResponse` and `BasicQueryResponse`.

**Who is affected**: Any API consumers parsing the response JSON — including the frontend (updated in this PR), mobile clients, scripts, or external integrations that read `dataset_id` from responses.

**Migration path**: Update all consumers to read `project_id` instead of `dataset_id`, and update code expecting integer IDs to handle UUID strings.
- **[CRITICAL]** Database schema: PKs changed from INTEGER to UUID, tables renamed: **Previous behavior**: `datasets.dataset_id` and `checkpoints.id` were INTEGER primary keys. The table was named `datasets`.

**New behavior**: Both PKs are now UUID type. The table `datasets` has been renamed to `projects`, `dataset_id` column renamed to `project_id`. Foreign key columns in `user_logs` and `checkpoints` also changed from INTEGER to UUID.

**Who is affected**: **All existing data** in production is affected. The migration converts integer IDs to UUIDs by calling `gen_random_uuid()`, which **discards all existing relationships** — existing foreign key values that were integers become new random UUIDs with no correspondence to original data. This is effectively a data reset for any deployment with existing data.

**Migration path**: Before applying, back up all data. Consider a data migration strategy that preserves relationships by mapping old integer IDs to deterministic UUIDs, or accept that this is a breaking reset of existing data.


### PR Description Issues

- **[HIGH]** PR description dramatically understates the scope of changes: The PR description frames this entirely as a narrow security patch ("Added `@` to the blocked patterns and tests for it"), but the diff contains a **massive restructuring** of the entire codebase:
- Complete backend rewrite with new project structure (`dataloom-backend/`)
- Complete frontend rewrite (`dataloom-frontend/`)
- Database schema migration (INTEGER → UUID, table renames)
- Breaking API route changes (`/datasets/` → `/projects/`)
- New Dockerfile, docker-compose, Alembic migrations, seed data
- New test infrastructure (pytest fixtures, vitest, 400+ lines of tests)
- New context API, toast system, error boundaries, hooks

A reviewer looking only at the description would be completely unprepared for the actual scope. This makes the PR very difficult to review safely.
  > **Suggestion:** The PR should either: (1) be split into a focused security fix PR and a separate refactoring PR, or (2) the description should be updated to accurately describe the full scope:

```markdown
## Description
This PR contains two major changes:

### 1. Security Fix: Block `@` prefix in query sanitization
[existing description]

### 2. Major Refactoring: Codebase restructure
- Renamed backend from `backend/` to `dataloom-backend/`, frontend to `dataloom-frontend/`
- Renamed all "dataset" concepts to "project" throughout API, schema, and DB
- Changed DB primary keys from INTEGER to UUID
- Added Alembic migration system, Docker support, comprehensive test suite
- Breaking changes: `/datasets/` routes → `/projects/`, `dataset_id` → `project_id`

**⚠️ This PR contains breaking database schema changes that will affect existing deployments.**
```
- **[MEDIUM]** Unchecked checklist items indicate incomplete validation: The PR checklist shows:
- `- [ ] Existing tests pass` — unchecked
- `- [ ] Tests pass locally` — unchecked

For a security fix PR, having unconfirmed test status is a significant concern. Additionally, the description references `local_dict={"__builtins__": {}}` as a security mechanism, which (as noted in a separate finding) does not actually work with `pandas.DataFrame.query()`.
  > **Suggestion:** Before merging:
1. Run the full test suite and check the boxes if tests pass.
2. Correct the description's claim about `local_dict` — it does not restrict builtins in `df.query()`.
3. Link to the relevant GitHub issue that this PR fixes (the `Fixes #(issue number)` placeholder is not filled in).

dataloom-backend/tests/test_security.py
```diff
@@ -0,0 +1,67 @@
...
+            validate_query_string("obj.__globals__")
+
+    def test_at_prefix_blocked(self):
+        """@pd.io.common.os bypasses dunder checks"""
+        with pytest.raises(HTTPException):
+            validate_query_string("@pd.io.common.os.system('echo test')")
+
+    def test_at_prefix_read_csv_blocked(self):
+        with pytest.raises(HTTPException):
+            validate_query_string("@pd.read_csv('/etc/passwd')")
```
ivantha, line 67: **[INFO]** Security tests do not cover all bypass vectors

The test suite for `validate_query_string` covers the explicitly listed patterns but is missing tests for:
- Unicode/encoding bypasses (e.g., using Unicode lookalikes for `@`)
- Chained attribute access without `@` (e.g., via string methods in numexpr)
- Empty string input
- Very long query strings that might cause ReDoS on the regex patterns
- Case sensitivity edge cases (the `re.IGNORECASE` flag is used — are all patterns truly case-insensitive by nature?)

**Suggestion:**
```
Add edge case tests:
```python
def test_empty_query_passes(self):
    result = validate_query_string("")
    assert result == ""

def test_os_dot_blocked(self):
    with pytest.raises(HTTPException):
        validate_query_string("os.system('ls')")

def test_exec_blocked(self):
    with pytest.raises(HTTPException):
        validate_query_string("exec('import os')")
```
```

File: dataloom-backend/app/main.py [hunk 17]
```diff
@@ -0,0 +1,62 @@
...
+@app.exception_handler(TransformationError)
+async def transformation_error_handler(request: Request, exc: TransformationError):
+    return JSONResponse(status_code=400, content={"detail": str(exc)})
+
+
+app.add_exception_handler(AppException, app_exception_handler)
+
+app.include_router(projects.router, prefix="/projects", tags=["projects"])
+app.include_router(transformations.router, prefix="/projects", tags=["transformations"])
```
ivantha, line 57: **[HIGH]** API routes changed from `/datasets/` to `/projects/`

**Previous behavior**: All dataset CRUD and transformation endpoints were mounted under `/datasets/` prefix.

**New behavior**: All endpoints are now under `/projects/` prefix. For example: `POST /datasets/upload` → `POST /projects/upload`, `POST /datasets/{id}/transform` → `POST /projects/{id}/transform`.

**Who is affected**: Any existing API clients, frontend code, integration tests, or external consumers that call the old `/datasets/` endpoints will receive 404 errors after this change.

**Migration path**: Update all API calls to use the `/projects/` prefix. The frontend code in this PR has been updated, but any external consumers (scripts, integrations, other services) must be updated manually.

dataloom-backend/app/schemas.py
```diff
@@ -0,0 +1,207 @@
...
+
+class ProjectResponse(BaseModel):
+    """Response for project CRUD operations."""
+    filename: str
+    file_path: str
+    project_id: uuid.UUID
+    columns: list[str]
+    row_count: int
+    rows: list[list]
+
```
ivantha, line 178: **[HIGH]** Response schema field `dataset_id` renamed to `project_id`

**Previous behavior**: All API responses included a `dataset_id` field (integer type) in `DatasetResponse` and `BasicQueryResponse`.

**New behavior**: Responses now use `project_id` (UUID type) in `ProjectResponse` and `BasicQueryResponse`.

**Who is affected**: Any API consumers parsing the response JSON — including the frontend (updated in this PR), mobile clients, scripts, or external integrations that read `dataset_id` from responses.

**Migration path**: Update all consumers to read `project_id` instead of `dataset_id`, and update code expecting integer IDs to handle UUID strings.

dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py
```diff
@@ -0,0 +1,129 @@
...
+               existing_nullable=False,
+               postgresql_using='gen_random_uuid()')
+    op.alter_column('user_logs', 'dataset_id',
+               existing_type=sa.INTEGER(),
+               type_=sa.Uuid(),
+               existing_nullable=False,
+               postgresql_using='gen_random_uuid()')
+    op.alter_column('user_logs', 'checkpoint_id',
+               existing_type=sa.INTEGER(),
+               type_=sa.Uuid(),
```
ivantha, line 65: **[CRITICAL]** Database schema: PKs changed from INTEGER to UUID, tables renamed

**Previous behavior**: `datasets.dataset_id` and `checkpoints.id` were INTEGER primary keys. The table was named `datasets`.

**New behavior**: Both PKs are now UUID type. The table `datasets` has been renamed to `projects`, `dataset_id` column renamed to `project_id`. Foreign key columns in `user_logs` and `checkpoints` also changed from INTEGER to UUID.

**Who is affected**: **All existing data** in production is affected. The migration converts integer IDs to UUIDs by calling `gen_random_uuid()`, which **discards all existing relationships** — existing foreign key values that were integers become new random UUIDs with no correspondence to original data. This is effectively a data reset for any deployment with existing data.

**Migration path**: Before applying, back up all data. Consider a data migration strategy that preserves relationships by mapping old integer IDs to deterministic UUIDs, or accept that this is a breaking reset of existing data.

ivantha:

ivantha:
Thank you for identifying and fixing this RCE risk — blocking the @ prefix in pandasql queries is a genuinely important security improvement! However, this PR carries a 26k-line diff that largely duplicates the old backend/ tree. Please rebase on main, extract only the sanitization fix, and resubmit as a focused PR so it can be reviewed and merged quickly.

PR #101, opened 2026-02-26
Docker-fix Implemented

### LINKED ISSUE #61: [Bug] Docker setup breaks on fresh clone
## Describe the bug

Running docker-compose up --build on a fresh clone fails due to multiple Docker setup issues. The frontend service references a missing Dockerfile, the backend container crashes because uvicorn is not found at /app/.venv/bin/uvicorn, Postgres fails to start if port 5432 is already in use locally, and Alembic migrations fail on an empty database causing the backend to restart repeatedly.

These combined issues make first-time local setup very difficult.

---

## To Reproduce

	1.	Clone the repository.
	2.	Ensure Docker Desktop is running.
	3.	From the project root, run:

```
docker-compose down -v
docker-compose up --build
```
	4.	Observe the following:
	•	Frontend service fails (missing dataloom-frontend/Dockerfile)
	•	Backend container exits with:
```
exec /app/.venv/bin/uvicorn: no such file or directory
```
	•	If local Postgres is running, DB container fails due to port 5432 conflict.
	•	On fresh DB, Alembic migrations fail and backend restarts continuously.

---

## Expected behavior
	•	Docker Compose should build and start the backend successfully.
	•	Postgres should start without conflicting with local installations.
	•	Backend should become available at http://localhost:4200.
	•	Database schema should initialize cleanly on first run for local/dev usage.

---

## Environment
	•	OS: macOS
	•	Docker Desktop: latest
	•	Python version: 3.12
	•	Postgres (local): installed
	•	Browser: Chrome

---

## Additional context

- Suggested improvements:
	•	Remove frontend service from docker-compose.yml (or add missing Dockerfile).
	•	Simplify backend Dockerfile to use standard pip install + uvicorn.
	•	Change DB host port mapping to 5433:5432 to avoid local Postgres conflicts.
	•	Initialize tables on startup (SQLModel.metadata.create_all) for local/dev to avoid Alembic failures on fresh databases.
	•	Optionally make Alembic migrations conditional via an environment variable.

These issues appear to affect any new contributor attempting Docker-based setup, not just a local machine-specific problem.

---

Sharma-Ji-21:
## Description

This PR improves backend containerization, fixes Docker startup failures, resolves local PostgreSQL port conflicts, and makes database migrations safe for both fresh and existing databases.

### Summary of Changes

- Removed frontend service from `docker-compose.yml` and containerized backend only.
- Updated compose to build backend from `dataloom-backend/`.
- Mapped host Postgres port `5433 -> 5432` (container) to prevent local port 5432 conflicts.
- Replaced non-standard uv-based Docker setup with a simpler Dockerfile using `pip` and direct `uvicorn` execution.
- Fixed missing `/app/.venv/bin/uvicorn` startup failure.
- Made migration `34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids` defensive:
  - Added table existence checks using SQLAlchemy inspector.
  - Prevented `ALTER TABLE` execution on non-existing tables.
  - Ensured safe execution on both fresh (`down -v`) and existing databases.
- Verified correct migration order:
  - 34e620d988fc → a1b2c3d4e5f6 → b2c3d4e5f6a7

### Issues Resolved

- Port 5432 conflict on developer machines running local Postgres.
- Docker build failure due to missing uvicorn in container path.
- Backend restart loop caused by migration crash on fresh database.
- Non-portable Docker setup using custom toolchain.

Fixes #61 

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [ ] Documentation update

---

### Docker Build

- Verified successful build of `dataloom-backend` image.
- Confirmed `uvicorn` available in container PATH.

### Docker Compose

Tested both scenarios:

1. Existing database:
 ```bash
 docker-compose down
 docker-compose up --build
```

2. Fresh database:
```bash
docker-compose down -v
docker-compose up --build
```

## Verification

- Backend successfully reachable at:
http://localhost:4200
- Database accessible from host at:
localhost:5433
- No port binding errors.
- No missing uvicorn errors.
- No migration failures or restart loops.
- Existing tests pass
- New tests added
- Manual testing

---

## Checklist

- My code follows the project’s style guidelines
- I have performed a self-review
- I have added/updated documentation as needed
- My changes generate no new warnings
- Tests pass locally

---

### DIFF
--- a/dataloom-backend/.env.example
+++ b/dataloom-backend/.env.example
@@ -5,10 +5,5 @@ DATABASE_URL="postgresql://user:password@localhost:5432/dataloom"
 CORS_ORIGINS=["http://localhost:3200"]
 
 # Directory for uploaded CSV files
-UPLOAD_DIR=uploads
-
-# Maximum file upload size in bytes (default: 10MB)
-MAX_UPLOAD_SIZE_BYTES=10485760
-
-# Debug mode
-DEBUG=false
+# Local dev: use relative path 'uploads' (default)
+# Docker: use '/app/uploads' to match volume mount
--- a/dataloom-backend/Dockerfile
+++ b/dataloom-backend/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12-slim
+FROM python:3.12-slim AS builder
 
 WORKDIR /app
 
@@ -9,6 +9,12 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 COPY pyproject.toml uv.lock ./
 RUN mkdir -p app && touch app/__init__.py
 RUN uv pip install --system --no-cache .
+
+FROM python:3.12-slim
+
+WORKDIR /app
+
+COPY --from=builder /usr/local /usr/local
 
 # Copy application code
 COPY . .
--- a/dataloom-backend/Dockerfile
+++ b/dataloom-backend/Dockerfile
@@ -2,22 +2,23 @@ FROM python:3.12-slim
 
 WORKDIR /app
 
-# Install dependencies
-COPY pyproject.toml ./
-RUN pip install --no-cache-dir "fastapi>=0.115" "uvicorn[standard]>=0.34" \
-    "sqlmodel>=0.0.22" "alembic>=1.14" "psycopg2-binary>=2.9" \
-    "pydantic-settings>=2.0" "pandas>=2.1" "python-dotenv>=1.0" \
-    "python-multipart>=0.0.6"
+# Install uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
+
+# Install dependencies (layer cached separately from app code)
+COPY pyproject.toml uv.lock ./
+RUN mkdir -p app && touch app/__init__.py
+RUN uv pip install --system --no-cache .
 
 # Copy application code
 COPY . .
 
-# Create non-root user
-RUN groupadd -r appuser && useradd -r -g appuser appuser && \
+# Create non-root user with home directory
+RUN groupadd -r appuser && useradd -r -g appuser -m appuser && \
     chown -R appuser:appuser /app
 
 USER appuser
 
 EXPOSE 4200
 
-CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "4200"]
+CMD ["python", "-m", "uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "4200"]
--- a/dataloom-backend/Dockerfile
+++ b/dataloom-backend/Dockerfile
@@ -1,27 +1,23 @@
-FROM python:3.12-slim AS builder
-
-WORKDIR /app
-
-COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
-
-COPY pyproject.toml uv.lock* ./
-RUN uv sync --frozen --no-dev
-
 FROM python:3.12-slim
 
 WORKDIR /app
 
-RUN groupadd -r appuser && useradd -r -g appuser appuser
+# Install dependencies
+COPY pyproject.toml ./
+RUN pip install --no-cache-dir "fastapi>=0.115" "uvicorn[standard]>=0.34" \
+    "sqlmodel>=0.0.22" "alembic>=1.14" "psycopg2-binary>=2.9" \
+    "pydantic-settings>=2.0" "pandas>=2.1" "python-dotenv>=1.0" \
+    "python-multipart>=0.0.6"
 
-COPY --from=builder /app/.venv /app/.venv
+# Copy application code
 COPY . .
 
-RUN chown -R appuser:appuser /app
-
-ENV PATH="/app/.venv/bin:$PATH"
-
-EXPOSE 4200
+# Create non-root user
+RUN groupadd -r appuser && useradd -r -g appuser appuser && \
+    chown -R appuser:appuser /app
 
 USER appuser
+
+EXPOSE 4200
 
 CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "4200"]
--- a/dataloom-backend/app/main.py
+++ b/dataloom-backend/app/main.py
@@ -7,12 +7,15 @@ from contextlib import asynccontextmanager
 from pathlib import Path
 from app.api.endpoints import projects, user_logs, transformations
 from app.config import get_settings
+from app.database import engine
 from app.exceptions import AppException, app_exception_handler
 from app.services.transformation_service import TransformationError
 from app.utils.logging import setup_logging, get_logger
 from fastapi import FastAPI, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
+from sqlalchemy import inspect
+from sqlmodel import SQLModel
 
 logger = get_logger(__name__)
 
@@ -24,7 +27,12 @@ async def lifespan(app):
     from alembic.config import Config
 
     alembic_cfg = Config("alembic.ini")
-    command.upgrade(alembic_cfg, "head")
+    inspector = inspect(engine)
+    if not inspector.has_table("alembic_version"):
+        SQLModel.metadata.create_all(engine)
+        command.stamp(alembic_cfg, "head")
+    else:
+        command.upgrade(alembic_cfg, "head")
 
     settings = get_settings()
     setup_logging(settings.debug)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,8 +18,11 @@ services:
       - "4200:4200"
     environment:
       DATABASE_URL: postgresql://postgres:postgres@db:5432/dataloom
+    volumes:
+      - uploads:/app/uploads
     depends_on:
       - db
 
 volumes:
   pgdata:
+  uploads:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
     ports:
-      - "5432:5432"
+      - "5433:5432"
     volumes:
       - pgdata:/var/lib/postgresql/data
 
@@ -20,14 +20,6 @@ services:
       DATABASE_URL: postgresql://postgres:postgres@db:5432/dataloom
     depends_on:
       - db
-
-  frontend:
-    build: dataloom-frontend
-    restart: unless-stopped
-    ports:
-      - "3200:3200"
-    depends_on:
-      - backend
 
 volumes:
   pgdata:
--- a/dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py
+++ b/dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py
@@ -20,44 +20,6 @@ depends_on: Union[str, Sequence[str], None] = None
 
 def upgrade() -> None:
     """Upgrade schema."""
-    bind = op.get_bind()
-    inspector = sa.inspect(bind)
-
-    if "datasets" not in inspector.get_table_names() and "projects" not in inspector.get_table_names():
-        op.execute("CREATE EXTENSION IF NOT EXISTS pgcrypto")
-
-        op.create_table(
-            "datasets",
-            sa.Column("dataset_id", sa.Uuid(), primary_key=True, server_default=sa.text("gen_random_uuid()")),
-            sa.Column("name", sa.String(), nullable=False),
-            sa.Column("description", sa.String(), nullable=True),
-            sa.Column("file_path", sa.String(), nullable=False),
-            sa.Column("upload_date", sa.DateTime(), server_default=sa.text("now()"), nullable=True),
-            sa.Column("last_modified", sa.DateTime(), server_default=sa.text("now()"), nullable=True),
-        )
-        op.create_index(op.f("ix_datasets_dataset_id"), "datasets", ["dataset_id"], unique=False)
-
-        op.create_table(
-            "checkpoints",
-            sa.Column("id", sa.Uuid(), primary_key=True, server_default=sa.text("gen_random_uuid()")),
-            sa.Column("dataset_id", sa.Uuid(), sa.ForeignKey("datasets.dataset_id"), nullable=False),
-            sa.Column("message", sa.String(), nullable=False),
-            sa.Column("created_at", sa.DateTime(), server_default=sa.text("now()"), nullable=True),
-        )
-        op.create_index(op.f("ix_checkpoints_id"), "checkpoints", ["id"], unique=False)
-
-        op.create_table(
-            "user_logs",
-            sa.Column("change_log_id", sa.Integer(), primary_key=True),
-            sa.Column("dataset_id", sa.Uuid(), sa.ForeignKey("datasets.dataset_id"), nullable=False),
-            sa.Column("action_type", sa.String(length=50), nullable=False),
-            sa.Column("action_details", sa.JSON(), nullable=False),
-            sa.Column("timestamp", sa.DateTime(), server_default=sa.text("now()"), nullable=False),
-            sa.Column("checkpoint_id", sa.Uuid(), sa.ForeignKey("checkpoints.id"), nullable=True),
-            sa.Column("applied", sa.Boolean(), server_default=sa.text("false"), nullable=False),
-        )
-        return
-
     # Step 1: Drop all foreign keys that reference columns being altered
     op.drop_constraint(op.f('checkpoints_dataset_id_fkey'), 'checkpoints', type_='foreignkey')
     op.drop_constraint(op.f('user_logs_dataset_id_fkey'), 'user_logs', type_='foreignkey')
--- a/dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py
+++ b/dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py
@@ -20,6 +20,44 @@ depends_on: Union[str, Sequence[str], None] = None
 
 def upgrade() -> None:
     """Upgrade schema."""
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+
+    if "datasets" not in inspector.get_table_names() and "projects" not in inspector.get_table_names():
+        op.execute("CREATE EXTENSION IF NOT EXISTS pgcrypto")
+
+        op.create_table(
+            "datasets",
+            sa.Column("dataset_id", sa.Uuid(), primary_key=True, server_default=sa.text("gen_random_uuid()")),
+            sa.Column("name", sa.String(), nullable=False),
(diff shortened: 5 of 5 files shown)


ivantha (changes requested):
uv should be used to install the dependencies.

ivantha (changes requested):
1. Editing an existing migration
Modifying 34e620d988fc after it may have been applied to existing databases is risky practice. A new migration (or startup-time create_all as suggested in the issue) would be safer.

2. Dockerfile lost multi-stage build
The original multi-stage build kept the final image slim — the only bug was the CMD/PATH. Could we keep the staged build and just fix the relevant part?

Sharma-Ji-21:
Thank you @ivantha for the detailed feedback.
I’ve restored the original multi-stage Docker build and limited the changes to fixing the runtime PATH/CMD issue.
I’ve also reverted modifications to the existing migration and moved the fresh-database handling to application startup to keep migration history immutable.
Additionally, I removed hardcoded upload paths from docker-compose and documented the Docker configuration via .env.example to keep the setup environment-agnostic.
Please let me know if anything else should be adjusted.

dataloom-backend/app/main.py
```diff
@@ -24,7 +27,12 @@ async def lifespan(app):
     from alembic.config import Config
 
     alembic_cfg = Config("alembic.ini")
-    command.upgrade(alembic_cfg, "head")
+    inspector = inspect(engine)
+    if not inspector.has_table("alembic_version"):
+        SQLModel.metadata.create_all(engine)
+        command.stamp(alembic_cfg, "head")
+    else:
+        command.upgrade(alembic_cfg, "head")
```
ivantha: **[HIGH]** create_all() bypass creates schema divergence between fresh and migrated databases

When the `alembic_version` table is absent, the code calls `SQLModel.metadata.create_all(engine)` and then stamps the database at `head` without actually running any migrations. This means:

1. **Schema source-of-truth split**: A fresh deployment gets its schema from ORM model classes, while an existing deployment gets it from the migration chain. Any discrepancy between the two (e.g. a migration that adds an index, modifies a column default, or runs a data migration) will be silently skipped on fresh installs.
2. **Future migrations may fail**: If migration `a1b2c3d4e5f6` or `b2c3d4e5f6a7` assumes specific column constraints, nullability, or data that `create_all` doesn't honour, those migrations will behave incorrectly or fail on a stamped-but-not-migrated database.
3. **Root cause already fixed**: The PR description states that migration `34e620d988fc` was already made defensive with table-existence checks. If that is true, `command.upgrade(alembic_cfg, 'head')` on a fresh database should succeed without this bypass at all — making this block redundant and risky.

The Alembic-recommended pattern for fresh databases is simply to let `upgrade head` run all migrations in order, each of which should be idempotent.

**Suggestion:**
```
Remove the `create_all` bypass entirely and rely on Alembic for all cases:

```python
async def lifespan(app):
    from alembic import command

    alembic_cfg = Config("alembic.ini")
    command.upgrade(alembic_cfg, "head")  # works for both fresh and existing DBs

    settings = get_settings()
    setup_logging(settings.debug)
```

If the defensive migration is already correct, this is all that is needed. Also remove the now-unused `from app.database import engine`, `from sqlalchemy import inspect`, and `from sqlmodel import SQLModel` imports.
```

File: dataloom-backend/app/main.py [hunk 2]
```diff
@@ -24,7 +27,12 @@ async def lifespan(app):
     from alembic.config import Config
 
     alembic_cfg = Config("alembic.ini")
-    command.upgrade(alembic_cfg, "head")
+    inspector = inspect(engine)
+    if not inspector.has_table("alembic_version"):
+        SQLModel.metadata.create_all(engine)
+        command.stamp(alembic_cfg, "head")
```
ivantha: **[HIGH]** TOCTOU race condition in alembic_version check on startup

The pattern `check → act` (inspect whether `alembic_version` exists, then decide which path to take) is a classic Time-of-Check/Time-of-Use (TOCTOU) race. If two containers start simultaneously against a fresh database — common in autoscaling, rolling deploys, or CI environments — both may observe the table absent, both call `create_all()`, and then both attempt `stamp('head')`. Depending on timing this can produce duplicate-table errors or a corrupted Alembic version row.

Alembic itself provides advisory locking via `with_autogenerate_context` / migration environment hooks and is designed to be run from a single place (e.g. an init container or a deploy hook), precisely to avoid this.

**Suggestion:**
```
Run migrations as a one-shot step in a Kubernetes init container, an ECS task override, or a deploy hook — not inside the application's lifespan. If that is not feasible, use a distributed lock (e.g. a Postgres advisory lock) before the migration block:

```python
from sqlalchemy import text
with engine.connect() as conn:
    conn.execute(text("SELECT pg_advisory_lock(12345)"))
    conn.execute(text("SELECT pg_advisory_unlock(12345)"))
```
```

File: docker-compose.yml [hunk 3]
```diff
@@ -18,16 +18,11 @@ services:
       - "4200:4200"
     environment:
       DATABASE_URL: postgresql://postgres:postgres@db:5432/dataloom
+    volumes:
```
ivantha: **[MEDIUM]** Hardcoded database credentials in docker-compose.yml

The `DATABASE_URL` value (`postgresql://postgres:postgres@db:5432/dataloom`) embeds plaintext credentials that will be committed to source control. Even for a development-only compose file, this is problematic:
- It trains developers to treat hardcoded secrets as acceptable.
- If this compose file is ever used in a staging/production context the credentials are already leaked in git history.
- It does not match the `.env.example` pattern already established in the project.

**Suggestion:**
```
Use environment variable substitution and provide defaults only via `.env`:

```yaml
services:
  backend:
    environment:
      DATABASE_URL: postgresql://${POSTGRES_USER:-postgres}:${POSTGRES_PASSWORD:-postgres}@db:5432/${POSTGRES_DB:-dataloom}
```

And document in `.env.example`:
```
POSTGRES_USER=postgres
POSTGRES_PASSWORD=postgres
POSTGRES_DB=dataloom
```
```

File: docker-compose.yml [hunk 4]
```diff
@@ -18,16 +18,11 @@ services:
...
-    build: dataloom-frontend
-    restart: unless-stopped
-    ports:
-      - "3200:3200"
-    depends_on:
-      - backend
-
 volumes:
   pgdata:
+  uploads:
```
ivantha: **[MEDIUM]** UPLOAD_DIR environment variable not set for Docker — volume mount may be unreachable

The compose file mounts the `uploads` named volume at `/app/uploads`, but the `.env.example` comment says the app must be configured with `UPLOAD_DIR=/app/uploads` for Docker. However, no `UPLOAD_DIR` environment variable is set in the `backend` service's `environment` block. If the application defaults `UPLOAD_DIR` to `uploads` (a relative path), uploaded files will be written to the container's working directory (`/app/uploads` relative path resolves to the same location only by coincidence). If the default is `uploads` without a leading slash and the CWD is `/app`, writes will land in `/app/uploads` — but this is fragile and undocumented. If the app resolves the path relative to something other than `/app`, files will be lost on container restart.

**Suggestion:**
```
Explicitly set the environment variable in the compose service:

```yaml
services:
  backend:
    environment:
      DATABASE_URL: ...
      UPLOAD_DIR: /app/uploads
```

And restore the variable with its Docker value to `.env.example`:
```
# Docker: set to /app/uploads to match volume mount
UPLOAD_DIR=/app/uploads
```
```

Comment (ivantha) [inline docker-compose.yml]: **[MEDIUM]** Missing healthcheck on db service — backend may start before PostgreSQL is ready

`depends_on: db` only guarantees the container has *started*, not that PostgreSQL is *accepting connections*. On slower machines or first-boot (when initdb runs), the backend process will attempt to connect and run migrations while Postgres is still initialising. This results in the restart loop the PR is trying to fix — the symptom is shifted, not the root cause.

**Suggestion:**
```
Add a healthcheck to the `db` service and a `condition: service_healthy` dependency:

```yaml
services:
  db:
    image: postgres:15
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U postgres"]
      interval: 5s
      timeout: 5s
      retries: 10
    ...

  backend:
    depends_on:
      db:
        condition: service_healthy
```
```

File: dataloom-backend/Dockerfile [hunk 5]
```diff
@@ -2,26 +2,29 @@ FROM python:3.12-slim AS builder
...
+COPY pyproject.toml uv.lock ./
+RUN mkdir -p app && touch app/__init__.py
+RUN uv pip install --system --no-cache .
 
 FROM python:3.12-slim
 
 WORKDIR /app
 
-RUN groupadd -r appuser && useradd -r -g appuser appuser
+COPY --from=builder /usr/local /usr/local
```
ivantha: **[MEDIUM]** Copying entire /usr/local from builder bloats image and copies build tooling into runtime

`COPY --from=builder /usr/local /usr/local` copies the complete `/usr/local` tree from the builder stage into the runtime image. This includes:
- The `uv` binary itself (a build-time tool)
- Python interpreter headers and `pip`
- Any build artefacts produced during package compilation
- Potentially any other tool installed in that stage

The original approach of copying only `/app/.venv` was more surgical. Copying all of `/usr/local` increases the final image size, includes tools that have no business being in a production runtime, and widens the CVE exposure surface.

**Suggestion:**
```
Install into a virtual environment and copy only that, or copy only the site-packages directory:

```dockerfile
# Builder
FROM python:3.12-slim AS builder
WORKDIR /app
COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
COPY pyproject.toml uv.lock ./
RUN uv venv /app/.venv && uv sync --frozen --no-dev

# Runtime
FROM python:3.12-slim
WORKDIR /app
COPY --from=builder /app/.venv /app/.venv
COPY . .
ENV PATH="/app/.venv/bin:$PATH"
```

This is essentially the original pattern, which was correct.
```

File: dataloom-backend/Dockerfile [hunk 6]
```diff
@@ -2,26 +2,29 @@ FROM python:3.12-slim AS builder
...
-COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+# Install uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 
-COPY pyproject.toml uv.lock* ./
-RUN uv sync --frozen --no-dev
+# Install dependencies (layer cached separately from app code)
+COPY pyproject.toml uv.lock ./
+RUN mkdir -p app && touch app/__init__.py
+RUN uv pip install --system --no-cache .
```
ivantha: **[MEDIUM]** Project code installed in builder stage but immediately overwritten — --no-install-project missing

The builder stage runs:
```
RUN mkdir -p app && touch app/__init__.py
RUN uv pip install --system --no-cache .
```
This creates a stub `app/__init__.py` so that the local package (`dataloom-backend`) can be resolved during `pip install .`. The package (including that stub) is then installed into the system Python. In the runtime stage, `COPY . .` overwrites the actual application source.

While functional, this installs a stale/stub version of the package in site-packages AND copies the real source code on top — so Python may find both. Depending on how imports resolve (editable vs non-editable install, `sys.path` ordering), this can cause subtle import confusion.

Additionally, the two separate `RUN` commands (`mkdir` and `uv pip install`) each create a layer, making the image slightly larger than necessary.

**Suggestion:**
```
Use `--no-install-project` to install only dependencies, then copy the real source in the runtime stage without a conflicting installed copy:

```dockerfile
COPY pyproject.toml uv.lock ./
RUN uv pip install --system --no-cache --no-install-project .
```

This eliminates the need for the stub `app/__init__.py` trick entirely.
```

File: dataloom-backend/.env.example [hunk 7]
```diff
@@ -5,10 +5,5 @@ DATABASE_URL="postgresql://user:password@localhost:5432/dataloom"
 CORS_ORIGINS=["http://localhost:3200"]
```
ivantha: **[MEDIUM]** MAX_UPLOAD_SIZE_BYTES and DEBUG removed from .env.example with no replacement

The diff removes three environment variables from `.env.example`:
- `UPLOAD_DIR=uploads`
- `MAX_UPLOAD_SIZE_BYTES=10485760`
- `DEBUG=false`

`UPLOAD_DIR` is replaced by a comment, but `MAX_UPLOAD_SIZE_BYTES` and `DEBUG` are dropped entirely. If the application still reads these settings (which is likely given they were documented), new contributors will have no idea these variables exist, what their defaults are, or when to configure them. This is the primary purpose of `.env.example`.

**Suggestion:**
```
Restore the removed variables with their defaults and expand the comments to cover both local and Docker contexts:

```
# Directory for uploaded CSV files
# Local dev: relative path (default)
# Docker: must match volume mount path
UPLOAD_DIR=uploads

# Maximum file upload size in bytes (default: 10MB)
MAX_UPLOAD_SIZE_BYTES=10485760

# Enable debug mode (set to true for local development)
DEBUG=false
```
```

docker-compose.yml
ivantha: **[LOW]** Backend service missing restart policy

The `db` service presumably has a `restart: unless-stopped` policy (visible in the surrounding context), but the `backend` service has no restart policy defined. This means that if the backend crashes after the initial startup (e.g. an unhandled exception, OOM kill), Docker will not restart it automatically. In a developer environment this is minor; in any shared or staging environment it silently stops the service.

**Suggestion:**
```
Add a restart policy to the backend service:

```yaml
services:
  backend:
    restart: unless-stopped
    ...
```
```

File: dataloom-backend/Dockerfile [hunk 8]
```diff
@@ -2,26 +2,29 @@ FROM python:3.12-slim AS builder
...
 
-ENV PATH="/app/.venv/bin:$PATH"
+USER appuser
 
 EXPOSE 4200
 
-USER appuser
-
-CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "4200"]
+CMD ["python", "-m", "uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "4200"]
```
ivantha: **[LOW]** Missing newline at end of Dockerfile

The diff marks `\ No newline at end of file` on the final `CMD` line. POSIX requires text files to end with a newline, and many tools (linters, git diff viewers) will flag this. Some editors or shell scripts that read the file line-by-line may also behave unexpectedly.

**Suggestion:**
```
Add a trailing newline after the `CMD` instruction:

```dockerfile
CMD ["python", "-m", "uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "4200"]
```
(followed by a blank line or simply ensure the file ends with `\n`)
```

File: docker-compose.yml (hunk 4 above)
Comment (ivantha) [inline docker-compose.yml]: **[LOW]** Breaking change — port 5433 and removed frontend — not communicated in compose file

Two breaking changes are made silently:
1. The host-side Postgres port changes from `5432` to `5433`. Any developer `.env`, DBeaver/TablePlus connection profile, or CI script that connected to port `5432` will break without warning.
2. The `frontend` service is removed, so `docker-compose up` no longer starts the full stack. Developers expecting the previous behaviour will wonder why the UI is not available.

Neither change has an inline comment or a `CHANGELOG` / `README` update visible in the diff.

**Suggestion:**
```
Add inline comments to communicate the change intent:

```yaml
ports:
  # Host port changed from 5432 to 5433 to avoid conflict with local Postgres.
  # Update your DB client connection profiles and any local .env files.
  - "5433:5432"
```

Also add a note to the project `README` or `docker-compose.yml` header explaining that the frontend must now be started separately (e.g. `cd dataloom-frontend && npm run dev`).
```

File: dataloom-backend/Dockerfile [hunk 9]
```diff
@@ -2,26 +2,29 @@ FROM python:3.12-slim AS builder
...
+COPY --from=builder /usr/local /usr/local
 
-COPY --from=builder /app/.venv /app/.venv
+# Copy application code
 COPY . .
 
-RUN chown -R appuser:appuser /app
+# Create non-root user with home directory
+RUN groupadd -r appuser && useradd -r -g appuser -m appuser && \
+    chown -R appuser:appuser /app
```
ivantha: **[INFO]** Positive: non-root user correctly created and applied before container start

The PR correctly adds `-m` to `useradd` to create a home directory for `appuser`, sets ownership of `/app`, and places the `USER appuser` directive before `EXPOSE` and `CMD`. This is an improvement over the original ordering where `USER` appeared after `EXPOSE`. Running the application as a non-root user is a security best practice that limits the blast radius of a container escape.

**Suggestion:**
```
No change needed. Consider also adding `--no-log-init` to `useradd` to avoid potential log initialisation issues on some systems:
```dockerfile
RUN groupadd -r appuser && useradd -r -g appuser -m --no-log-init appuser && \
    chown -R appuser:appuser /app
```
```

File: dataloom-backend/Dockerfile (hunk 8 above)
Comment (ivantha) [inline dataloom-backend/Dockerfile]: **[INFO]** Positive: python -m uvicorn is more portable than direct binary invocation

Switching from `uvicorn` (direct binary lookup) to `python -m uvicorn` (module invocation via the interpreter) is a well-established best practice. It guarantees uvicorn is loaded from the same Python environment that the application runs in, avoiding PATH ambiguity issues — exactly the class of bug this PR was created to fix.

**Suggestion:**
```
No change needed. This is the recommended way to invoke uvicorn in Docker environments.
```

File: dataloom-backend/Dockerfile [hunk 10]
```diff
@@ -2,26 +2,29 @@ FROM python:3.12-slim AS builder
...
 WORKDIR /app
 
-COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+# Install uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 
-COPY pyproject.toml uv.lock* ./
-RUN uv sync --frozen --no-dev
+# Install dependencies (layer cached separately from app code)
+COPY pyproject.toml uv.lock ./
```
ivantha: **[INFO]** uv.lock changed from optional (glob) to required — verify lock file is committed

The original `COPY pyproject.toml uv.lock* ./` used a glob that made `uv.lock` optional (the `*` means zero or more characters after `uv.lock`). The new `COPY pyproject.toml uv.lock ./` requires the file to be present or the build will fail with a `file not found` error. This is generally desirable for reproducibility — but only if `uv.lock` is committed to the repository. If `uv.lock` is in `.gitignore`, the Docker build will now fail in CI.

**Suggestion:**
```
Verify that `uv.lock` is tracked in git (`git ls-files uv.lock`). If it is not, either commit it or revert to the glob form. Locking dependency versions in version control is strongly recommended.
```

ivantha (commented):
## ◎ Argus Review

This PR addresses real, legitimate Docker pain points — port 5432 conflicts, a missing uvicorn binary, and a crashing migration on fresh databases — and the core intent is solid. However, the workaround in `main.py` that bypasses Alembic in favour of `SQLModel.metadata.create_all()` for fresh databases introduces a schema-consistency risk that could silently diverge new deployments from the migration-managed schema. Credentials are hardcoded in docker-compose, the `.env.example` drops important variables without replacement, and the Dockerfile's broad `/usr/local` copy from the builder stage carries unnecessary bloat and tooling into the runtime image. The PR is a meaningful step forward but requires revisions — particularly to the migration logic — before merging to `main`.

ivantha:

ivantha:
Thank you for working on the Docker fix! This work has been superseded by #213 which provides a cleaner implementation. Feel free to follow up on that PR or open a new one if you see additional improvements.

PR #105, opened 2026-02-26
fix(context-menu): extract inline menu from Table.jsx and improve state management

### LINKED ISSUE #95: Refactor Table.jsx Context Menu
**Difficulty:** Medium
**Area:** frontend

## Description

The right-click context menu in `Table.jsx` is currently implemented inline within the table component, making the component large and harder to maintain. The menu logic (positioning, visibility toggling, click-outside handling, and action dispatch) is interleaved with the table's rendering and cell editing logic, creating a complex component with multiple responsibilities.

DataLoom already has a `useContextMenu` hook in `hooks/useContextMenu.js` that encapsulates some of the context menu state management. This issue extracts the context menu rendering and behavior into a dedicated `ContextMenu.jsx` component, leveraging the existing hook and improving the separation of concerns within the table component.

The refactored context menu should be a positioned overlay that appears at the cursor location on right-click, displays the appropriate action items, and closes when the user clicks outside of it or selects an action.

## Acceptance Criteria

- [ ] A standalone, reusable context menu component exists
- [ ] Context menu appears at the cursor position on right-click
- [ ] Context menu displays the correct action items for the clicked context
- [ ] Clicking a menu item executes the action and closes the menu
- [ ] Clicking outside the menu closes it
- [ ] Pressing Escape closes the menu
- [ ] Browser's default right-click menu is suppressed on table cells
- [ ] The table component no longer contains inline context menu rendering logic
- [ ] All context menu operations (add/delete row, add/delete/rename column) still work correctly
- [ ] The context menu hook is properly utilized for state management
- [ ] Menu is styled consistently with the rest of the application
- [ ] Menu position adjusts if it would overflow the viewport edge

## Resources

- `dataloom-frontend/src/Components/Table.jsx` — current inline context menu implementation
- `dataloom-frontend/src/hooks/useContextMenu.js` — existing context menu hook
- `dataloom-frontend/src/api/transforms.js` — `transformProject()` for menu actions

vikas-cse123:
## Description

This PR refactors the inline context menu implementation in Table.jsx into a standalone  ContextMenu component and integrates the existing  useContextMenu hook.

Closes #95

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
Manually tested the context menu by right-clicking on rows and columns.
Verified it opens at the correct position, closes on outside click, Escape, and scroll.
Confirmed all add/delete/rename operations work and the menu closes on both success and failure.

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/ContextMenu.jsx
+++ b/dataloom-frontend/src/Components/ContextMenu.jsx
@@ -1,11 +1,12 @@
-import { useEffect, useRef, useState, useLayoutEffect } from "react";
+import { useState, useEffect, useRef, useLayoutEffect } from "react";
 import PropTypes from "prop-types";
-
 const ContextMenu = ({ isOpen, position, contextData, onClose, actions }) => {
   const menuRef = useRef(null);
 
   const [adjustedPosition, setAdjustedPosition] = useState(position);
 
+  // useLayoutEffect (not useEffect) ensures clamping runs synchronously
+  // before the browser paints, preventing visible position jump.
   useLayoutEffect(() => {
     if (!isOpen) return;
 
@@ -35,28 +36,21 @@ const ContextMenu = ({ isOpen, position, contextData, onClose, actions }) => {
   }, [isOpen, position]);
 
   useEffect(() => {
-    if (!isOpen) {
-      setAdjustedPosition(position);
-    }
+    if (!isOpen) return;
 
     const handleClickOutside = (event) => {
-      if (menuRef.current && !menuRef.current.contains(event.target)) {
-        onClose();
-      }
+      if (menuRef.current && !menuRef.current.contains(event.target)) onClose();
     };
-
     const handleKeyDown = (event) => {
-      if (event.key === "Escape") {
-        onClose();
-      }
+      if (event.key === "Escape") onClose();
     };
-
-    const handleScroll = () => {
-      onClose();
-    };
+    const handleScroll = () => onClose();
 
     document.addEventListener("click", handleClickOutside);
     document.addEventListener("keydown", handleKeyDown);
+    // Close the context menu on any scroll event.
+    // Using capture phase ensures scrolling inside nested containers
+    // (like the table's overflow area) also closes the menu.
     window.addEventListener("scroll", handleScroll, true);
 
     return () => {
@@ -64,13 +58,15 @@ const ContextMenu = ({ isOpen, position, contextData, onClose, actions }) => {
       document.removeEventListener("keydown", handleKeyDown);
       window.removeEventListener("scroll", handleScroll, true);
     };
-  }, [isOpen, onClose, position]);
+  }, [isOpen, onClose]);
 
   if (!isOpen) return null;
 
   return (
     <div
       ref={menuRef}
+      role="menu"
+      aria-label="Context menu"
       className="fixed bg-white border border-gray-200 rounded-lg shadow-lg p-1 z-50"
       style={{
         top: adjustedPosition.y,
--- a/dataloom-frontend/src/Components/Table.jsx
+++ b/dataloom-frontend/src/Components/Table.jsx
@@ -1,12 +1,26 @@
 import ContextMenu from "./ContextMenu";
 import { useContextMenu } from "../hooks/useContextMenu";
-import { useEffect, useState } from "react";
+import { useState, useEffect } from "react";
 import { transformProject } from "../api";
 import { useProjectContext } from "../context/ProjectContext";
 import InputDialog from "./common/InputDialog";
 import Toast from "./common/Toast";
 import DtypeBadge from "./common/DtypeBadge";
-import proptypes from "prop-types";
+import PropTypes from "prop-types";
+
+const MenuButton = ({ children, onClick }) => (
+  <button
+    role="menuitem"
+    className="block w-full text-left text-sm text-gray-700 px-3 py-1.5 hover:bg-gray-100 rounded-md transition-colors duration-150 whitespace-nowrap"
+    onClick={onClick}
+  >
+    {children}
+  </button>
+);
+MenuButton.propTypes = {
+  children: PropTypes.node.isRequired,
+  onClick: PropTypes.func.isRequired,
+};
 
 const Table = ({ projectId, data: externalData }) => {
   const { columns: ctxColumns, rows: ctxRows, dtypes, updateData } = useProjectContext();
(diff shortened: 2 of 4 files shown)


ivantha (changes requested):
The InputDialog/Toast got swapped out for native prompt()/alert() which is a step back UX-wise, and the close logic (click outside, Escape) is duplicated between the hook and the component. Fix those and this should be good to go.

dataloom-frontend/src/Components/Table.jsx
```diff
@@ -287,51 +259,40 @@ const Table = ({ projectId, data: externalData }) => {
...
+                <MenuButton onClick={() => handleAddRow(data.rowIndex)}>Add Row</MenuButton>
+                <MenuButton onClick={() => handleDeleteRow(data.rowIndex)}>Delete Row</MenuButton>
+              </>
+            );
+          }
+        }}
+      />
 
       {inputConfig && (
         <InputDialog
```
ivantha: **[LOW]** `actions` render-prop returns `undefined` for unrecognised `data.type`

The `actions` function passed to `ContextMenu` handles `"column"` and `"row"` explicitly, but has no final fallback:

```js
actions={(data) => {
  if (!data) return null;
  if (data.type === "column") { return (...); }
  if (data.type === "row")    { return (...); }
  // implicit return undefined
}}
```

If `contextData` ever contains an unexpected `type` (e.g., a future `"header"` type, or a typo), `actions(contextData)` returns `undefined`. React renders `undefined` silently, but it also means `ContextMenu` renders a visible but empty box, which is confusing UX.

**Suggestion:**
```
Add an explicit `return null` at the end:

```js
actions={(data) => {
  if (!data) return null;
  if (data.type === "column") { return (...); }
  if (data.type === "row")    { return (...); }
  return null; // safe fallback
}}
```
```

File: dataloom-frontend/src/Components/Table.jsx [hunk 2]
```diff
@@ -361,4 +322,13 @@ Table.propTypes = {
...
   }),
 };
+
+const MenuButton = ({ children, onClick }) => (
+  <button
+    className="block w-full text-left text-sm text-gray-700 px-3 py-1.5 hover:bg-gray-100 rounded-md transition-colors duration-150 whitespace-nowrap"
+    onClick={onClick}
+  >
+    {children}
+  </button>
```
ivantha: **[LOW]** `MenuButton` declared after the component that uses it

`MenuButton` is a `const` arrow-function component defined **below** the `Table` component that references it. While this works at runtime (closures capture the binding by the time React calls the render function, after module initialisation is complete), it violates the common convention of declaring helpers **before** their first use. It also means that a naive static analysis pass may flag a reference-before-declaration issue, and it reduces readability for engineers scanning the file top-to-bottom.

**Suggestion:**
```
Move the `MenuButton` definition to the top of the file (or at least above `Table`), and optionally give it PropTypes:

```jsx
// Place before Table component definition
const MenuButton = ({ children, onClick }) => (
  <button
    className="block w-full text-left text-sm text-gray-700 px-3 py-1.5
               hover:bg-gray-100 rounded-md transition-colors duration-150 whitespace-nowrap"
    onClick={onClick}
  >
    {children}
  </button>
);

MenuButton.propTypes = {
  children: PropTypes.node.isRequired,
  onClick: PropTypes.func.isRequired,
};
```
```

Comment (ivantha) [inline dataloom-frontend/src/Components/Table.jsx]: **[LOW]** `MenuButton` has no PropTypes definition

Every other component in the diff (`ContextMenu`, `Table`) has PropTypes. `MenuButton` has none. While it is a small, internal helper, the inconsistency means a missing `onClick` or wrong `children` type would go undetected in development.

**Suggestion:**
```
Add PropTypes for `MenuButton`:

```js
MenuButton.propTypes = {
  children: PropTypes.node.isRequired,
  onClick: PropTypes.func.isRequired,
};
```
```

File: dataloom-frontend/src/Components/ContextMenu.jsx [hunk 3]
```diff
@@ -0,0 +1,100 @@
...
+      }
+    };
+
+    const handleScroll = () => {
+      onClose();
+    };
+
+    document.addEventListener("click", handleClickOutside);
+    document.addEventListener("keydown", handleKeyDown);
+    window.addEventListener("scroll", handleScroll, true);
```
ivantha: **[LOW]** Scroll listener uses capture phase — closes menu on any scroll anywhere

`window.addEventListener('scroll', handleScroll, true)` uses the **capture phase**, which means it intercepts scroll events on *all* elements before they reach their targets — including scrolling within the table's overflow container. This will close the context menu any time the user scrolls the table, even if the menu is anchored to a fixed position and would remain fully visible.

This may or may not be the desired UX, but the choice should be deliberate.

**Suggestion:**
```
If scroll-within-table should not close the menu, switch to bubble phase and scope to the scroll container:

```js
// bubble phase; only fires on the document scroll root
window.addEventListener("scroll", handleScroll);
```

Or pass a `scrollContainerRef` into `ContextMenu` and listen only on that element. If aggressive close-on-any-scroll is intentional, add a comment explaining the decision.
```

File: dataloom-frontend/src/Components/ContextMenu.jsx [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+    newX = Math.max(8, newX);
+    newY = Math.max(8, newY);
+
+    setAdjustedPosition({ x: newX, y: newY });
+  }, [isOpen, position]);
+
+  useEffect(() => {
+    if (!isOpen) {
+      setAdjustedPosition(position);
+    }
```
ivantha: **[LOW]** `adjustedPosition` reset in `useEffect` is redundant and increases complexity

The block:
```js
if (!isOpen) {
  setAdjustedPosition(position);
}
```
runs when the menu is closing and resets `adjustedPosition` to the raw (unadjusted) `position`. This reset is never visually seen because the component returns `null` immediately after. When the menu next opens, `useLayoutEffect` will correctly recalculate the clamped position before the first paint, making the pre-paint state irrelevant.

The reset only adds a state write on every close, coupling the position-reset concern with the listener-setup concern.

**Suggestion:**
```
Remove the reset — it is not needed:

```js
useEffect(() => {
  // Removed: if (!isOpen) { setAdjustedPosition(position); }
  // useLayoutEffect will compute the correct position on open.
  if (!isOpen) return;
  ...
}, [isOpen, onClose]);
```
```

PR #122, opened 2026-02-27
Added sample rows transform and preview table updates

### LINKED ISSUE #96: Add Sample Rows Transformation
**Difficulty:** Medium
**Area:** full stack

## Description

When working with large datasets, it is often useful to take a random sample of rows to quickly explore the data, test transformations on a smaller subset, or create a representative sample for analysis. DataLoom currently has no built-in way to randomly sample rows from a dataset.

This issue adds a "Sample Rows" transformation that uses pandas' `DataFrame.sample()` to return a random subset of rows. The user specifies the number of rows to sample, and optionally a random seed for reproducibility. This is a straightforward transformation that follows the existing pattern and provides a practical tool for working with large datasets.

## Acceptance Criteria

- [ ] A "sample" operation type is available
- [ ] Sampling returns the correct number of randomly selected rows
- [ ] Sampling with the same seed produces identical results (reproducibility)
- [ ] Requesting more rows than exist returns all rows without error
- [ ] Sample size of 0 or negative is rejected with a validation error
- [ ] The result has clean sequential row indices
- [ ] The sample rows form renders with a sample size input and optional seed input
- [ ] Form submits correctly and data table refreshes with sampled rows
- [ ] Operation is accessible from the side menu
- [ ] Transformation is logged in the project change log
- [ ] Sampling a large dataset (e.g., 10,000 rows down to 100) works quickly

## Resources

- `dataloom-backend/app/services/transformation_service.py` — existing transform functions
- `dataloom-backend/app/schemas.py` — `OperationType` enum
- `dataloom-backend/app/api/endpoints/transformations.py` — `_handle_basic_transform()`
- `dataloom-frontend/src/Components/forms/FilterForm.jsx` — reference form pattern
- `dataloom-frontend/src/Components/MenuNavbar.jsx` — side menu
- [pandas `sample` documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sample.html)

Sharma-Ji-21:
## Description

This change introduces a full “Sample Rows” workflow to DataLoom, spanning backend logic, API validation, logging, and frontend UX. The backend now supports sampling with an optional random seed for deterministic results, rejects invalid sample sizes, and preserves original row numbers so users can trace sampled rows back to their positions in the source dataset. The transformation is logged like other operations, and preview results are returned in the same response shape as existing transforms.

On the frontend, a dedicated Sample Rows form was added to the Transform menu. The preview table has been brought in line with the main dataset table: consistent typography, spacing, and header styling, plus both horizontal and vertical scrolling with a sticky header. Column headings preserve original casing (no forced uppercase), and the S.No column displays original row numbers from the sampled data instead of sequential preview indices.

Fixes #96 

---

## Type of Change

- [ ] Bug fix  
- [x] New feature  
- [ ] Breaking change  
- [ ] Documentation update  

---

## How Has This Been Tested?

- [ ] Existing tests pass  
- [ ] New tests added  
- [x] Manual testing  

Manual testing details:
- Sample Rows with valid size and seed from the UI.
- Verified reproducibility with the same seed.
- Verified behavior when sample size exceeds dataset length.
- Verified validation errors for zero/negative sample size.
- Confirmed preview table scrolling (both axes), spacing, and header casing.
- Confirmed S.No reflects original row numbers for sampled data.
- Verified transformation appears in logs.

---

## Screenshots

https://github.com/user-attachments/assets/4e585d7a-a47e-4a0e-931f-b27db6191f98

---

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/transformations.py
+++ b/dataloom-backend/app/api/endpoints/transformations.py
@@ -91,6 +91,12 @@ def _handle_basic_transform(df, transformation_input, project, db, project_id):
         p = transformation_input.cast_data_type_params
         return ts.cast_data_type(df, p.column, p.target_type), True
 
+    elif op == 'sample':
+        if not transformation_input.sample_rows_params:
+            raise HTTPException(status_code=400, detail="Sample parameters required")
+        p = transformation_input.sample_rows_params
+        return ts.sample_rows(df, p.sample_size, p.random_seed), False
+
     elif op == 'trimWhitespace':
         if not transformation_input.trim_whitespace_params:
             raise HTTPException(status_code=400, detail="Trim whitespace parameters required")
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -36,6 +36,7 @@ class OperationType(str, Enum):
     renameCol = 'renameCol'
     castDataType = 'castDataType'
     trimWhitespace = 'trimWhitespace'
+    sample = 'sample'
 
 
 class DropDup(str, Enum):
@@ -69,6 +70,7 @@ class ActionTypes(str, Enum):
     changeCellValue = 'changeCellValue'
     renameCol = 'renameCol'
     castDataType = 'castDataType'
+    sample = 'sample'
     trimWhitespace = 'trimWhitespace'
 
 
@@ -132,6 +134,12 @@ class CastDataTypeParams(BaseModel):
     target_type: DataType
 
 
+class SampleRowsParams(BaseModel):
+    """Parameters for sampling rows from a dataset."""
+    sample_size: int
+    random_seed: Optional[int] = None
+
+
 class TrimWhitespaceParams(BaseModel):
     """Parameters for trimming whitespace from columns."""
     column: str
@@ -192,6 +200,7 @@ class TransformationInput(BaseModel):
     change_cell_value: Optional[ChangeCellValue] = None
     rename_col_params: Optional[RenameColumnParams] = None
     cast_data_type_params: Optional[CastDataTypeParams] = None
+    sample_rows_params: Optional[SampleRowsParams] = None
     trim_whitespace_params: Optional[TrimWhitespaceParams] = None
 
 
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -383,6 +383,36 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
     return result.reset_index()
 
 
+def sample_rows(df: pd.DataFrame, sample_size: int, random_seed: int = None) -> pd.DataFrame:
+    """Sample rows from the DataFrame.
+
+    Args:
+        df: Source DataFrame.
+        sample_size: Number of rows to sample.
+        random_seed: Optional seed for reproducibility.
+
+    Returns:
+        DataFrame with sampled rows and reset indices.
+
+    Raises:
+        TransformationError: If sample_size is not positive.
+    """
+    if sample_size <= 0:
+        raise TransformationError("Sample size must be greater than 0")
+
+    # If sample_size is greater than dataset length, return all rows
+    n_rows = len(df)
+    actual_sample_size = min(sample_size, n_rows)
+
+    # Use random_seed if provided
+    kwargs = {'n': actual_sample_size, 'random_state': random_seed} if random_seed is not None else {'n': actual_sample_size}
+
+    sampled = df.sample(**kwargs)
+    sampled = sampled.copy()
+    sampled.insert(0, "__row_number", sampled.index + 1)
+    return sampled.reset_index(drop=True)
+
+
 def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_details: dict) -> pd.DataFrame:
     """Replay a logged transformation from its serialized form.
 
@@ -445,6 +475,11 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
         column = action_details['cast_data_type_params']['column']
         target_type = action_details['cast_data_type_params']['target_type']
         return cast_data_type(df, column, target_type)
+
+    elif action_type == 'sample':
+        sample_size = action_details['sample_rows_params']['sample_size']
+        random_seed = action_details['sample_rows_params'].get('random_seed')
+        return sample_rows(df, sample_size, random_seed)
 
     elif action_type == 'trimWhitespace':
         column = action_details['trim_whitespace_params']['column']
--- a/dataloom-frontend/src/Components/forms/SampleRowsForm.jsx
+++ b/dataloom-frontend/src/Components/forms/SampleRowsForm.jsx
@@ -1,0 +1,106 @@
+import { useState } from "react";
+import PropTypes from "prop-types";
+import { transformProject } from "../../api";
+import TransformResultPreview from "./TransformResultPreview";
+
+const SampleRowsForm = ({ projectId, onClose }) => {
+    const [sampleSize, setSampleSize] = useState("");
+    const [randomSeed, setRandomSeed] = useState("");
+    const [result, setResult] = useState(null);
+    const [loading, setLoading] = useState(false);
+    const [error, setError] = useState("");
+
+    const handleSubmit = async (e) => {
+        e.preventDefault();
+        setError("");
+
+        // Validate sample size
+        const size = parseInt(sampleSize, 10);
+        if (!sampleSize || isNaN(size) || size <= 0) {
+            setError("Sample size must be a positive integer");
+            return;
+        }
+
+        console.log("Submitting sample with parameters:", {
+            sample_size: size,
+            random_seed: randomSeed ? parseInt(randomSeed, 10) : null,
+        });
+        setLoading(true);
+        try {
+            const params = {
+                sample_size: size,
+            };
+            if (randomSeed) {
+                params.random_seed = parseInt(randomSeed, 10);
+            }
+
+            const response = await transformProject(projectId, {
+                operation_type: "sample",
+                sample_rows_params: params,
+            });
+            setResult(response);
+            console.log("Sample API response:", response);
+        } catch (err) {
+            setError(err.response?.data?.detail || "Error applying sample");
+            console.error("Error applying sample:", err.response?.data || err.message);
+        } finally {
+            setLoading(false);
+        }
+    };
+
+    return (
+        <div className="p-4 border border-gray-200 rounded-lg bg-white">
+            <form onSubmit={handleSubmit}>
+                <h3 className="font-semibold text-gray-900 mb-2">Sample Rows</h3>
+                {error && <div className="mb-4 p-2 bg-red-50 border border-red-200 text-red-700 rounded-md text-sm">{error}</div>}
+                <div className="flex flex-wrap mb-4">
+                    <div className="w-full sm:w-1/2 mb-2">
+                        <label className="block mb-1 text-sm font-medium text-gray-700">Sample Size:</label>
+                        <input
+                            type="number"
+                            value={sampleSize}
+                            onChange={(e) => setSampleSize(e.target.value)}
+                            placeholder="e.g., 100"
+                            className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+                            required
+                        />
+                    </div>
+                    <div className="w-full sm:w-1/2 mb-2 pl-2">
+                        <label className="block mb-1 text-sm font-medium text-gray-700">Random Seed (Optional):</label>
+                        <input
+                            type="number"
+                            value={randomSeed}
+                            onChange={(e) => setRandomSeed(e.target.value)}
+                            placeholder="e.g., 42"
+                            className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+                        />
+                    </div>
+                </div>
+                <div className="flex justify-between">
+                    <button
+                        type="submit"
+                        className="bg-blue-500 hover:bg-blue-600 text-white px-4 py-2 rounded-md font-medium transition-colors duration-150"
+                        disabled={loading}
+                    >
+                        Sample Rows
+                    </button>
+                    <button
+                        type="button"
+                        onClick={onClose}
+                        className="bg-white border border-gray-300 text-gray-700 hover:bg-gray-50 px-4 py-2 rounded-md font-medium transition-colors duration-150"
+                    >
+                        Cancel
+                    </button>
+                </div>
+            </form>
+            {result && <TransformResultPreview columns={result.columns} rows={result.rows} />}
+        </div>
+    );
+};
+
+SampleRowsForm.propTypes = {
+    projectId: PropTypes.string.isRequired,
+    onClose: PropTypes.func.isRequired,
+};
+
+export default SampleRowsForm;
--- a/dataloom-frontend/src/Components/forms/TransformResultPreview.jsx
+++ b/dataloom-frontend/src/Components/forms/TransformResultPreview.jsx
@@ -9,44 +9,65 @@ const TransformResultPreview = ({ columns, rows }) => {
     return <p className="text-gray-500 mt-4">No columns available</p>;
   }
 
+  const sNoColumnKey = "__row_number";
+  const hasRowNumber = columns[0] === sNoColumnKey;
+  const displayColumns = hasRowNumber ? columns.slice(1) : columns;
+  const displayRows = hasRowNumber ? rows.map((row) => row.slice(1)) : rows;
+  const sNoValues = hasRowNumber ? rows.map((row) => row[0]) : rows.map((_, index) => index + 1);
+
   return (
     <div className="p-4 mt-4 border border-gray-200 rounded-lg bg-gray-50">
       <h4 className="font-medium text-sm text-gray-700 mb-2">API Response:</h4>
-      <table className="min-w-full divide-y divide-gray-200">
-        <thead className="bg-gray-100">
-          <tr>
-            {columns.map((col, index) => (
-              <th
-                key={index}
-                className="px-6 py-3 text-left text-xs font-medium text-gray-500 uppercase tracking-wider"
-              >
-                {col}
+      <div
+        className="overflow-x-scroll overflow-y-auto border border-gray-200 rounded-lg shadow-sm"
+        style={{ maxHeight: "calc(100vh - 140px)" }}
+      >
+        <table className="min-w-full bg-white">
+          <thead className="sticky top-0 bg-gray-50">
+            <tr>
+              <th className="py-1.5 px-3 border-b border-gray-200 text-left text-xs font-medium text-gray-500">
+                <button className="w-full text-left text-gray-500 hover:text-gray-700 hover:bg-gray-100 py-0.5 px-1.5 rounded-md transition-colors duration-150">
+                  S.No.
+                </button>
               </th>
-            ))}
-          </tr>
-        </thead>
-        <tbody className="bg-white divide-y divide-gray-200">
-          {rows.map((row, rowIndex) => (
-            <tr key={rowIndex}>
-              {row.map((cell, cellIndex) => (
-                <td
-                  key={cellIndex}
-                  className="px-6 py-4 whitespace-nowrap text-sm text-gray-700"
+              {displayColumns.map((col, index) => (
+                <th
+                  key={index}
+                  className="py-1.5 px-3 border-b border-gray-200 text-left text-xs font-medium text-gray-500"
                 >
-                  {cell}
-                </td>
+                  <button className="w-full text-left text-gray-500 hover:text-gray-700 hover:bg-gray-100 py-0.5 px-1.5 rounded-md transition-colors duration-150">
+                    {col}
+                  </button>
+                </th>
               ))}
             </tr>
-          ))}
-        </tbody>
-      </table>
+          </thead>
+          <tbody>
+            {displayRows.map((row, rowIndex) => (
+              <tr
+                key={rowIndex}
+                className="border-b border-gray-100 hover:bg-gray-50 transition-colors duration-150"
+              >
+                <td className="py-1 px-3 text-xs text-gray-700">
+                  {sNoValues[rowIndex]}
+                </td>
+                {row.map((cell, cellIndex) => (
+                  <td key={cellIndex} className="py-1 px-3 text-xs text-gray-700">
+                    {cell}
+                  </td>
+                ))}
+              </tr>
+            ))}
+          </tbody>
+        </table>
+      </div>
     </div>
   );
 };
 
 TransformResultPreview.propTypes = {
   columns: PropTypes.arrayOf(PropTypes.string).isRequired,
-  rows: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.string)).isRequired,
+  rows: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.any)).isRequired,
 };
 
 export default TransformResultPreview;
--- a/dataloom-frontend/src/constants/operationTypes.js
+++ b/dataloom-frontend/src/constants/operationTypes.js
@@ -25,6 +25,8 @@ export const DROP_DUPLICATE = "dropDuplicate";
 export const ADV_QUERY_FILTER = "advQueryFilter";
 /** @type {string} Create a pivot table */
 export const PIVOT_TABLE = "pivotTables";
+/** @type {string} Sample random rows from dataset */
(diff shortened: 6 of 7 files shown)


ivantha (commented):
## ◎ Argus Review

This PR introduces a well-structured "Sample Rows" feature spanning backend service, API validation, schema definitions, and frontend UI — a solid end-to-end addition. The implementation follows established project patterns (Pydantic params, operation type enum, logged transformation replay), and the preview table improvements are a genuine UX win. However, there are two concrete bugs that need fixing before merge: a column-name collision in `sample_rows` that will raise a `ValueError` if the column `__row_number` already exists, and unhandled `TransformationError` exceptions that could surface as 500s. Several medium-priority issues also need attention, including a missing `onTransform` callback that silently prevents the main dataset from refreshing, unvalidated `randomSeed` input that can send `NaN` to the backend, and tight frontend–backend coupling via a magic string.

dataloom-frontend/src/Components/forms/SampleRowsForm.jsx
```diff
@@ -0,0 +1,106 @@
...
+                params.random_seed = parseInt(randomSeed, 10);
+            }
+
+            const response = await transformProject(projectId, {
+                operation_type: "sample",
+                sample_rows_params: params,
+            });
+            setResult(response);
+            console.log("Sample API response:", response);
+        } catch (err) {
```
ivantha, line 43: **[HIGH]** console.log statements left in production code

Two `console.log` calls are present in the submitted code:

```js
console.log("Submitting sample with parameters:", { ... });
// ...
console.log("Sample API response:", response);
```

These expose internal API payloads and response shapes to the browser console in production. For a data-transformation tool, responses may contain sensitive dataset contents. Debug logging should always be removed before merging, or replaced with a proper structured logger that is stripped or silenced in production builds.

**Suggestion:**
```
Remove both `console.log` calls entirely. If observability is needed, use a dedicated logger utility (e.g., `logger.debug(...)`) that is conditionally enabled only in development via environment variables (`import.meta.env.DEV`).
```

dataloom-backend/app/api/endpoints/transformations.py
```diff
@@ -91,6 +91,12 @@ def _handle_basic_transform(df, transformation_input, project, db, project_id):
         p = transformation_input.cast_data_type_params
         return ts.cast_data_type(df, p.column, p.target_type), True
 
+    elif op == 'sample':
+        if not transformation_input.sample_rows_params:
+            raise HTTPException(status_code=400, detail="Sample parameters required")
+        p = transformation_input.sample_rows_params
```
ivantha, line 97: **[HIGH]** Inconsistency between endpoint return value (False) and log-replay handler

The endpoint handler returns `False` as the second tuple value for the `sample` operation:

```python
return ts.sample_rows(df, p.sample_size, p.random_seed), False
```

Based on the pattern of the surrounding code (e.g., `cast_data_type` returns `True`), this second boolean controls whether the transformation is persisted/logged. Returning `False` means the sample result is never committed to the project state.

However, `apply_logged_transformation` in `transformation_service.py` includes a handler for `action_type == 'sample'`. If the operation is never saved, this replay code is dead and misleading — it implies that `sample` is a persistent transformation, contradicting the `False` return here.

Conversely, if `False` was set by mistake and `sample` is supposed to be a persistent operation, then it will silently never be saved.

**Suggestion:**
```
Decide on the intended semantics:
- **Preview-only**: Remove the `apply_logged_transformation` case for `'sample'` to avoid dead code and confusion.
- **Persistent transform**: Change the return to `True` and verify the full save/log pathway works end-to-end.
```

dataloom-backend/app/services/transformation_service.py
```diff
@@ -383,6 +383,36 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
...
+    if sample_size <= 0:
+        raise TransformationError("Sample size must be greater than 0")
+
+    # If sample_size is greater than dataset length, return all rows
+    n_rows = len(df)
+    actual_sample_size = min(sample_size, n_rows)
+
+    # Use random_seed if provided
+    kwargs = {'n': actual_sample_size, 'random_state': random_seed} if random_seed is not None else {'n': actual_sample_size}
+
```
ivantha, line 409: **[MEDIUM]** Row number computation breaks on non-integer or non-zero-based index

The row number is computed as:

```python
sampled.insert(0, "__row_number", sampled.index + 1)
```

This has two issues:
1. **Non-integer index**: If the DataFrame carries a `DatetimeIndex`, `StringIndex`, or any non-numeric index (e.g., after a `set_index()` operation upstream), `index + 1` will raise a `TypeError` at runtime, producing an unhandled 500 error rather than a clear validation message.
2. **Column name collision**: `df.insert()` raises a `ValueError` if a column named `"__row_number"` already exists in the DataFrame. If a user's dataset already has a column by that name, the operation will crash with an opaque error instead of a user-friendly one.

**Suggestion:**
```
Guard both cases explicitly:

```python
# Check for column name collision
row_col = "__row_number"
if row_col in df.columns:
    row_col = "__row_number_sampled"  # or raise TransformationError

# Guard against non-integer index
try:
    row_numbers = sampled.index + 1
except TypeError:
    row_numbers = range(1, len(sampled) + 1)  # fallback to sequential

sampled = sampled.copy()
sampled.insert(0, row_col, row_numbers)
```
```

File: dataloom-frontend/src/Components/forms/SampleRowsForm.jsx [hunk 4]
```diff
@@ -0,0 +1,106 @@
...
+                    </div>
+                    <div className="w-full sm:w-1/2 mb-2 pl-2">
+                        <label className="block mb-1 text-sm font-medium text-gray-700">Random Seed (Optional):</label>
+                        <input
+                            type="number"
+                            value={randomSeed}
+                            onChange={(e) => setRandomSeed(e.target.value)}
+                            placeholder="e.g., 42"
+                            className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+                        />
```
ivantha, line 76: **[MEDIUM]** No onTransform callback — main dataset view never refreshes

Every other form rendered by `MenuNavbar.jsx` (e.g., `PivotTableForm`, `CastDataTypeForm`, `TrimWhitespaceForm`) receives and calls an `onTransform` prop to notify the parent component that the dataset has changed, triggering a table refresh. `SampleRowsForm` is rendered without this prop:

```jsx
<SampleRowsForm
  projectId={projectId}
  onClose={() => setShowSampleRowsForm(false)}
/>
```

This means the main dataset display will never update after a successful sample operation. The component's `PropTypes` definition also omits `onTransform`, making this a deliberate omission rather than an oversight — but if it is intentional (preview-only), it should be clearly documented.

**Suggestion:**
```
If the operation should be persistent and refresh the main view:
```jsx
// In MenuNavbar.jsx
<SampleRowsForm
  projectId={projectId}
  onClose={() => setShowSampleRowsForm(false)}
  onTransform={onTransform}
/>
// In SampleRowsForm.jsx, after setResult(response):
if (onTransform) onTransform();
```
If it is intentionally preview-only, add a JSDoc comment explaining this design decision and update the PR description accordingly.
```

File: dataloom-frontend/src/Components/forms/SampleRowsForm.jsx [hunk 5]
```diff
@@ -0,0 +1,106 @@
...
+    const handleSubmit = async (e) => {
+        e.preventDefault();
+        setError("");
+
+        // Validate sample size
+        const size = parseInt(sampleSize, 10);
+        if (!sampleSize || isNaN(size) || size <= 0) {
+            setError("Sample size must be a positive integer");
+            return;
+        }
```
ivantha, line 22: **[MEDIUM]** Random seed accepts negative integers without frontend validation

The random seed input is `type="number"` with no `min` attribute, so users can freely enter negative values. NumPy's `random_state` with a negative seed raises a `ValueError` (`Seed must be between 0 and 2**32 - 1`), which will bubble up as an unhandled 500-level backend error rather than a clean validation message.

The backend's `SampleRowsParams` schema also places no constraint on `random_seed`:
```python
random_seed: Optional[int] = None
```

So neither layer catches this.

**Suggestion:**
```
Add frontend and backend validation:

**Frontend** (`SampleRowsForm.jsx`):
```jsx
<input
  type="number"
  min="0"
  max="4294967295"  // 2^32 - 1
  ...
/>
```

**Backend** (`schemas.py`):
```python
from pydantic import Field

class SampleRowsParams(BaseModel):
    sample_size: int = Field(gt=0)
    random_seed: Optional[int] = Field(default=None, ge=0, le=4294967295)
```
```

File: dataloom-backend/app/schemas.py [hunk 6]
```diff
@@ -132,6 +134,12 @@ class CastDataTypeParams(BaseModel):
     target_type: DataType
 
 
+class SampleRowsParams(BaseModel):
+    """Parameters for sampling rows from a dataset."""
+    sample_size: int
```
ivantha, line 139: **[MEDIUM]** Missing Pydantic field-level validation for sample_size

The validation that `sample_size > 0` is enforced only inside the service function (`transformation_service.py`), not at the schema level. Pydantic is the natural layer for this constraint:

```python
class SampleRowsParams(BaseModel):
    sample_size: int          # accepts 0, -1, -100 without error
    random_seed: Optional[int] = None
```

This means invalid payloads pass through request parsing before being rejected, making it harder to test and compose the schema independently.

**Suggestion:**
```
```python
from pydantic import Field

class SampleRowsParams(BaseModel):
    """Parameters for sampling rows from a dataset."""
    sample_size: int = Field(gt=0, description="Number of rows to sample (must be positive)")
    random_seed: Optional[int] = Field(default=None, ge=0)
```
This produces a standard 422 Unprocessable Entity with a structured error body instead of a custom 500-level `TransformationError`.
```

File: dataloom-frontend/src/Components/forms/TransformResultPreview.jsx [hunk 7]
```diff
@@ -9,44 +9,65 @@ const TransformResultPreview = ({ columns, rows }) => {
...
-            <tr key={rowIndex}>
-              {row.map((cell, cellIndex) => (
-                <td
-                  key={cellIndex}
-                  className="px-6 py-4 whitespace-nowrap text-sm text-gray-700"
+              {displayColumns.map((col, index) => (
+                <th
+                  key={index}
+                  className="py-1.5 px-3 border-b border-gray-200 text-left text-xs font-medium text-gray-500"
                 >
```
ivantha, line 37: **[MEDIUM]** Button elements inside table headers have no click handler — dead interactive UI

Every column header (including S.No.) is wrapped in a `<button>` element:

```jsx
<th ...>
  <button className="... hover:text-gray-700 hover:bg-gray-100 ...">S.No.</button>
</th>
```

Buttons have interactive styling (`hover:bg-gray-100`, cursor change), strongly implying sort functionality to users. However, no `onClick` handler is attached. Clicking these buttons does nothing. This violates the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and is an accessibility issue (`<button>` in `<th>` without a meaningful action misleads screen readers).

**Suggestion:**
```
Either:
1. **Remove the buttons** and render plain text if sorting is not implemented:
   ```jsx
   <th className="py-1.5 px-3 border-b border-gray-200 text-left text-xs font-medium text-gray-500">
     {col}
   </th>
   ```
2. **Implement sorting** by adding a click handler that sorts the table by that column (and updates styling to indicate sort direction).
```

dataloom-frontend/src/Components/forms/TransformResultPreview.jsx
```diff
@@ -9,44 +9,65 @@ const TransformResultPreview = ({ columns, rows }) => {
     return <p className="text-gray-500 mt-4">No columns available</p>;
   }
 
+  const sNoColumnKey = "__row_number";
+  const hasRowNumber = columns[0] === sNoColumnKey;
+  const displayColumns = hasRowNumber ? columns.slice(1) : columns;
+  const displayRows = hasRowNumber ? rows.map((row) => row.slice(1)) : rows;
+  const sNoValues = hasRowNumber ? rows.map((row) => row[0]) : rows.map((_, index) => index + 1);
+
   return (
```
ivantha, line 18: **[MEDIUM]** Tight coupling via magic column name __row_number between backend and frontend

The `TransformResultPreview` component silently inspects the first column for the internal column name `"__row_number"`:

```js
const sNoColumnKey = "__row_number";
const hasRowNumber = columns[0] === sNoColumnKey;
```

This creates invisible coupling: the frontend now depends on a backend implementation detail. If the backend renames this column (e.g., `_original_row_idx`), the frontend silently falls back to sequential numbering without any error — a subtle, hard-to-diagnose regression. Additionally, the check `columns[0] === sNoColumnKey` assumes the row-number column is always first, which is fragile.

**Suggestion:**
```
Prefer an explicit contract. For example, the API response could include a dedicated metadata field:

```json
{ "columns": [...], "rows": [...], "row_number_column": "__row_number" }
```

Or, accept a separate `rowNumberColumn` prop in `TransformResultPreview`:
```jsx
<TransformResultPreview
  columns={result.columns}
  rows={result.rows}
  rowNumberColumn="__row_number"
/>
```
This makes the dependency explicit and traceable.
```

File: dataloom-frontend/src/constants/operationTypes.js [hunk 9]
```diff
@@ -25,6 +25,8 @@ export const DROP_DUPLICATE = "dropDuplicate";
 export const ADV_QUERY_FILTER = "advQueryFilter";
 /** @type {string} Create a pivot table */
 export const PIVOT_TABLE = "pivotTables";
+/** @type {string} Sample random rows from dataset */
```
ivantha, line 28: **[LOW]** SAMPLE_ROWS constant defined but not used in SampleRowsForm

A constant `SAMPLE_ROWS = "sample"` is added to `operationTypes.js`, but `SampleRowsForm.jsx` hardcodes the string `"sample"` directly in the API call:

```js
operation_type: "sample",  // magic string, not using SAMPLE_ROWS constant
```

The entire point of the constants file is to avoid magic strings and enable refactoring. The pattern is already used by other forms, making this an inconsistency.

**Suggestion:**
```
Import and use the constant:
```js
import { SAMPLE_ROWS } from "../../constants/operationTypes";
// ...
operation_type: SAMPLE_ROWS,
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 10]
```diff
@@ -383,6 +383,36 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
...
+    Raises:
+        TransformationError: If sample_size is not positive.
+    """
+    if sample_size <= 0:
+        raise TransformationError("Sample size must be greater than 0")
+
+    # If sample_size is greater than dataset length, return all rows
+    n_rows = len(df)
+    actual_sample_size = min(sample_size, n_rows)
+
```
ivantha, line 406: **[LOW]** Unnecessary kwargs dictionary construction — pandas sample() accepts None for random_state

`pd.DataFrame.sample()` natively accepts `None` for `random_state`, which means passing `None` is equivalent to no seed (non-deterministic). The conditional kwargs dict is unnecessary complexity:

```python
kwargs = {'n': actual_sample_size, 'random_state': random_seed} if random_seed is not None else {'n': actual_sample_size}
sampled = df.sample(**kwargs)
```

This pattern also makes the code harder to read at a glance.

**Suggestion:**
```
```python
# random_seed=None is already handled by pandas — simplify to one call:
sampled = df.sample(n=actual_sample_size, random_state=random_seed)
```
```

File: dataloom-frontend/src/Components/forms/TransformResultPreview.jsx [hunk 11]
```diff
@@ -9,44 +9,65 @@ const TransformResultPreview = ({ columns, rows }) => {
...
     <div className="p-4 mt-4 border border-gray-200 rounded-lg bg-gray-50">
```
ivantha, line 19: **[LOW]** overflow-x-scroll forces persistent scrollbar; should be overflow-x-auto

The container uses `overflow-x-scroll` (Tailwind: always show horizontal scrollbar) rather than `overflow-x-auto` (show only when content overflows). On macOS the difference is subtle due to overlay scrollbars, but on Windows and Linux this will render a permanent horizontal scrollbar even on narrow tables, degrading the UX unnecessarily.

**Suggestion:**
```
```jsx
<div
  className="overflow-x-auto overflow-y-auto border border-gray-200 rounded-lg shadow-sm"
  style={{ maxHeight: "calc(100vh - 140px)" }}
>
```
```

Comment (ivantha) [inline dataloom-frontend/src/Components/forms/TransformResultPreview.jsx:19]: **[LOW]** Hardcoded maxHeight calc is brittle and context-dependent

`style={{ maxHeight: "calc(100vh - 140px)" }}` hardcodes an offset of 140px, which is presumably based on a specific parent container height at the time of writing. If the form is ever embedded in a differently-sized modal, panel, or page context, the table could overflow or be clipped unexpectedly. This magic number has no associated comment explaining where 140px comes from.

**Suggestion:**
```
Add a comment explaining the value, or prefer a CSS-based approach:
```jsx
{/* 140px = navbar(48) + form header/controls(92) */}
style={{ maxHeight: "calc(100vh - 140px)" }}
```
Or consider using `max-h-96` / `max-h-[32rem]` Tailwind utilities for a more maintainable constraint, or pass `maxHeight` as a prop so the parent controls it.
```

File: dataloom-frontend/src/Components/forms/TransformResultPreview.jsx [hunk 12]
```diff
@@ -9,44 +9,65 @@ const TransformResultPreview = ({ columns, rows }) => {
     return <p className="text-gray-500 mt-4">No columns available</p>;
   }
 
+  const sNoColumnKey = "__row_number";
+  const hasRowNumber = columns[0] === sNoColumnKey;
+  const displayColumns = hasRowNumber ? columns.slice(1) : columns;
+  const displayRows = hasRowNumber ? rows.map((row) => row.slice(1)) : rows;
+  const sNoValues = hasRowNumber ? rows.map((row) => row[0]) : rows.map((_, index) => index + 1);
```
ivantha, line 16: **[LOW]** h4 label reads 'API Response:' — implementation detail exposed to user

The preview panel's heading is:
```jsx
<h4 className="font-medium text-sm text-gray-700 mb-2">API Response:</h4>
```
"API Response" is a developer-centric term that leaks an implementation detail to end users, which may be confusing in a product UI.

**Suggestion:**
```
Use a user-facing label:
```jsx
<h4 className="font-medium text-sm text-gray-700 mb-2">Preview</h4>
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR introduces a well-scoped "Sample Rows" feature spanning backend service logic, API schema, and frontend UI — and the implementation is largely clean and follows established patterns. However, there are two high-severity issues that need addressing before merge: two `console.log` statements left in production code that could leak dataset contents, and a critical semantic inconsistency where the endpoint signals the operation as non-persistent (`False`) while a log-replay handler exists for it, creating dead or contradictory code. Several medium-severity issues also require attention: the frontend never calls `onTransform` to refresh the main dataset view, negative random seeds are not validated on either layer, Pydantic schema validation is missing for `sample_size`, and table header buttons have hover styling but no click handlers, implying unimplemented sort functionality. The PR description also directly contradicts the code by claiming the operation is logged like other transforms.

### PR Description Issues

- **[INFO]** Test checklist items unchecked — automated test coverage not provided: The PR checklist leaves two key items explicitly unchecked:

```
- [ ] Existing tests pass
- [ ] New tests added
```

While manual testing is documented, there are no automated tests added for the backend service function, API endpoint, or frontend component. Given that `sample_rows` has multiple important edge cases (size clamping, seed reproducibility, invalid sizes, index edge cases), the absence of tests is a quality gap that will compound over time.
  > **Suggestion:** Update the description to be honest about test coverage, and ideally add backend pytest tests and/or frontend Jest/Vitest tests before merge. If tests are intentionally deferred, open a follow-up issue and reference it in the description.
- **[INFO]** PR description doesn't clarify whether sample is a persistent or preview-only operation: The description states: *"The transformation is logged like other operations"* — yet the backend endpoint returns `False` for the sample operation, which (based on the surrounding code pattern) means it is **not** committed/persisted. This is a direct contradiction between the description and the actual diff behavior.
  > **Suggestion:** Clarify the intended behavior:

> **Revised description**: "Sample Rows is a **preview-only** operation: results are shown in the UI but are not committed to the project state (i.e., the underlying dataset is not mutated). Accordingly, the operation does not appear in the project's transformation log after submission." 

Or, if it should be persistent, fix the code and update accordingly.

File: dataloom-backend/app/services/transformation_service.py [hunk 13]
```diff
@@ -383,6 +383,36 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
...
+    kwargs = {'n': actual_sample_size, 'random_state': random_seed} if random_seed is not None else {'n': actual_sample_size}
+
+    sampled = df.sample(**kwargs)
+    sampled = sampled.copy()
+    sampled.insert(0, "__row_number", sampled.index + 1)
+    return sampled.reset_index(drop=True)
+
+
 def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_details: dict) -> pd.DataFrame:
     """Replay a logged transformation from its serialized form.
```
ivantha, line 418: **[LOW]** No unit tests added for sample_rows service function

The `sample_rows` function introduces several edge cases that are not covered by automated tests:
- `sample_size > len(df)` (silent clamp)
- `sample_size = 0` or negative (should raise `TransformationError`)
- `random_seed` reproducibility
- `__row_number` column naming
- Non-integer DataFrame index behavior

The PR checklist explicitly marks both "Existing tests pass" and "New tests added" as unchecked.

**Suggestion:**
```
Add pytest tests:
```python
def test_sample_rows_basic():
    df = pd.DataFrame({'a': range(10), 'b': range(10)})
    result = sample_rows(df, 5, random_seed=42)
    assert len(result) == 5
    assert '__row_number' in result.columns

def test_sample_rows_exceeds_length():
    df = pd.DataFrame({'a': range(3)})
    result = sample_rows(df, 100)
    assert len(result) == 3

def test_sample_rows_invalid_size():
    df = pd.DataFrame({'a': range(10)})
        sample_rows(df, 0)

def test_sample_rows_reproducible():
    df = pd.DataFrame({'a': range(100)})
    r1 = sample_rows(df, 10, random_seed=42)
    r2 = sample_rows(df, 10, random_seed=42)
    pd.testing.assert_frame_equal(r1, r2)
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 14]
```diff
@@ -383,6 +383,36 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
...
+    Args:
+        df: Source DataFrame.
+        sample_size: Number of rows to sample.
+        random_seed: Optional seed for reproducibility.
+
+    Returns:
+        DataFrame with sampled rows and reset indices.
+
+    Raises:
+        TransformationError: If sample_size is not positive.
```
ivantha, line 398: **[INFO]** Silent clamp when sample_size exceeds dataset length — no user feedback

When `sample_size > len(df)`, the function silently returns all rows:

```python
actual_sample_size = min(sample_size, n_rows)
```

The user asked for, say, 500 rows but gets back 100 with no indication that their request was truncated. In the frontend the result will just appear, with no warning that fewer rows were returned than requested.

**Suggestion:**
```
Consider surfacing this as a warning in the API response. If the response shape supports metadata:
```python
# In the service, return a tuple (df, warning)
if sample_size > n_rows:
    warning = f"Requested {sample_size} rows but dataset only has {n_rows}. Returning all rows."
```
Alternatively, at minimum log a warning server-side:
```python
import logging
logger = logging.getLogger(__name__)
if sample_size > n_rows:
    logger.warning("sample_size %d exceeds dataset length %d; clamping.", sample_size, n_rows)
```
```

File: dataloom-frontend/src/Components/forms/SampleRowsForm.jsx [hunk 15]
```diff
@@ -0,0 +1,106 @@
...
+
+        console.log("Submitting sample with parameters:", {
+            sample_size: size,
+            random_seed: randomSeed ? parseInt(randomSeed, 10) : null,
+        });
+        setLoading(true);
+        try {
+            const params = {
+                sample_size: size,
+            };
```
ivantha, line 32: **[INFO]** parseInt(randomSeed, 10) called twice on same value

The random seed string is parsed to an integer twice: once in the `console.log` call and once when building `params`:

```js
console.log("Submitting sample with parameters:", {
  random_seed: randomSeed ? parseInt(randomSeed, 10) : null, // first parse
});
// ...
if (randomSeed) {
  params.random_seed = parseInt(randomSeed, 10); // second parse
}
```

Minor redundancy, but the `console.log` line should be removed anyway, which resolves this.

**Suggestion:**
```
Parse once, store in a variable:
```js
const seedValue = randomSeed ? parseInt(randomSeed, 10) : null;
const params = { sample_size: size, ...(seedValue !== null && { random_seed: seedValue }) };
```
```

ivantha:

ivantha:
Thank you for implementing the sample rows transform! This work has been superseded by #216 which provides the same feature. Feel free to follow up on that PR.

PR #126, opened 2026-02-27
Add string replace transformation

### LINKED ISSUE #82: Add String Replace Transformation
**Difficulty:** Medium
**Area:** full stack

## Description

DataLoom currently lacks a string find-and-replace operation, which is one of the most commonly needed data cleaning tasks. Users often need to standardize values within a column, such as replacing abbreviations ("NY" to "New York"), fixing inconsistent formatting, or removing unwanted characters.

This issue adds a new "String Replace" transformation that allows users to select a column, specify a find string, and specify a replacement string. The operation replaces all occurrences of the find string within each cell of the selected column. This follows the same pattern as existing transformations (backend service function + operation type enum + frontend form).

## Acceptance Criteria

- [ ] A "string replace" operation type is available
- [ ] String replacement works correctly on string columns
- [ ] Non-string columns are cast to string before replacement
- [ ] Find string is treated as a literal (not regex) for safety
- [ ] Empty replacement value is supported (effectively deleting the find string)
- [ ] The string replace form renders with a column dropdown, find input, and replace input
- [ ] Operation is accessible from the menu/navbar
- [ ] Transformation is logged in the project change log
- [ ] Data table refreshes to show updated values after transformation

## Resources

- `dataloom-backend/app/services/transformation_service.py` — existing transform functions for reference
- `dataloom-backend/app/schemas.py` — `OperationType` enum
- `dataloom-backend/app/api/endpoints/transformations.py` — `_handle_basic_transform()`
- `dataloom-frontend/src/Components/forms/FilterForm.jsx` — reference form pattern
- `dataloom-frontend/src/Components/MenuNavbar.jsx` — menu where transforms are listed
- `dataloom-frontend/src/constants/operationTypes.js` — operation type constants
- [pandas `str.replace` documentation](https://pandas.pydata.org/docs/reference/api/pandas.Series.str.replace.html)

hanzalahwaheed:
## Description

Adds string replace transformation form to the project.

Fixes https://github.com/c2siorg/dataloom/issues/82

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Existing tests pass
- [ ] New tests added
- [X] Manual testing

## Screenshots
image

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -39,6 +39,7 @@ class OperationType(StrEnum):
     renameCol = "renameCol"
     castDataType = "castDataType"
     trimWhitespace = "trimWhitespace"
+    dropNa = "dropNa"
     melt = "melt"
     stringReplace = "stringReplace"
 
@@ -193,6 +194,13 @@ class StringReplaceParams(BaseModel):
     column: str
     find_value: str
     replace_value: str
+
+    @field_validator("find_value")
+    @classmethod
+    def find_value_must_not_be_empty(cls, v: str) -> str:
+        if not v:
+            raise ValueError("find_value must not be empty")
+        return v
 
 
 # --- Complex transformation parameter schemas ---
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -39,8 +39,8 @@ class OperationType(StrEnum):
     renameCol = "renameCol"
     castDataType = "castDataType"
     trimWhitespace = "trimWhitespace"
-    dropNa = "dropNa"
     melt = "melt"
+    stringReplace = "stringReplace"
 
 
 class DropDup(StrEnum):
@@ -80,6 +80,7 @@ class ActionTypes(StrEnum):
     trimWhitespace = "trimWhitespace"
     dropNa = "dropNa"
     melt = "melt"
+    stringReplace = "stringReplace"
 
 
 # --- Basic transformation parameter schemas ---
@@ -186,6 +187,14 @@ class DropNaParams(BaseModel):
         return v
 
 
+class StringReplaceParams(BaseModel):
+    """Parameters for find-and-replace on a string column."""
+
+    column: str
+    find_value: str
+    replace_value: str
+
+
 # --- Complex transformation parameter schemas ---
 
 
@@ -267,6 +276,7 @@ class TransformationInput(BaseModel):
     trim_whitespace_params: TrimWhitespaceParams | None = None
     drop_na_params: DropNaParams | None = None
     melt_params: MeltParams | None = None
+    string_replace_params: StringReplaceParams | None = None
 
 
 class BasicQueryResponse(BaseModel):
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -40,6 +40,7 @@ class OperationType(StrEnum):
     renameCol = "renameCol"
     castDataType = "castDataType"
     trimWhitespace = "trimWhitespace"
+    dropNa = "dropNa"
     melt = "melt"
     sample = "sample"
     stringReplace = "stringReplace"
@@ -195,6 +196,13 @@ class StringReplaceParams(BaseModel):
     column: str
     find_value: str
     replace_value: str
+
+    @field_validator("find_value")
+    @classmethod
+    def find_value_must_not_be_empty(cls, v: str) -> str:
+        if not v:
+            raise ValueError("find_value must not be empty")
+        return v
 
 
 # --- Complex transformation parameter schemas ---
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -40,9 +40,9 @@ class OperationType(StrEnum):
     renameCol = "renameCol"
     castDataType = "castDataType"
     trimWhitespace = "trimWhitespace"
-    dropNa = "dropNa"
     melt = "melt"
     sample = "sample"
+    stringReplace = "stringReplace"
 
 
 class DropDup(StrEnum):
@@ -82,6 +82,7 @@ class ActionTypes(StrEnum):
     trimWhitespace = "trimWhitespace"
     dropNa = "dropNa"
     melt = "melt"
+    stringReplace = "stringReplace"
 
 
 # --- Basic transformation parameter schemas ---
@@ -188,6 +189,14 @@ class DropNaParams(BaseModel):
         return v
 
 
+class StringReplaceParams(BaseModel):
+    """Parameters for find-and-replace on a string column."""
+
+    column: str
+    find_value: str
+    replace_value: str
+
+
 # --- Complex transformation parameter schemas ---
 
 
@@ -284,6 +293,7 @@ class TransformationInput(BaseModel):
     drop_na_params: DropNaParams | None = None
     melt_params: MeltParams | None = None
     sample_params: SampleParams | None = None
+    string_replace_params: StringReplaceParams | None = None
 
 
 class BasicQueryResponse(BaseModel):
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -343,6 +343,12 @@ def string_replace(df: pd.DataFrame, column: str, find_value: str, replace_value
     if column not in df.columns:
         raise TransformationError(f"Column '{column}' not found")
 
+    if not (pd.api.types.is_string_dtype(df[column]) or pd.api.types.is_object_dtype(df[column])):
+        raise TransformationError(
+            f"Column '{column}' is not a string column (dtype: {df[column].dtype}). "
+            "Cast it to string first before using string replace."
+        )
+
     df = df.copy()
     df[column] = df[column].astype(str).str.replace(find_value, replace_value, regex=False)
     return df
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -591,9 +591,13 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
         return melt_dataframe(df, params)
 
     elif action_type == "stringReplace":
-        column = action_details["string_replace_params"]["column"]
-        find_value = action_details["string_replace_params"]["find_value"]
-        replace_value = action_details["string_replace_params"]["replace_value"]
+        params = action_details.get("string_replace_params", {})
+        column = params.get("column")
+        find_value = params.get("find_value")
+        replace_value = params.get("replace_value")
+        if not column or find_value is None or replace_value is None:
+            logger.warning("Missing params for stringReplace replay: %s", action_details)
+            return df
         return string_replace(df, column, find_value, replace_value)
 
     else:
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -328,6 +328,26 @@ def trim_whitespace(df: pd.DataFrame, column: str) -> pd.DataFrame:
     return df
 
 
+def string_replace(df: pd.DataFrame, column: str, find_value: str, replace_value: str) -> pd.DataFrame:
+    """Replace occurrences of a substring in a column.
+
+    Args:
+        df: Source DataFrame.
+        column: Column name to perform replacement on.
+        find_value: The substring to find.
+        replace_value: The string to replace it with.
+
+    Returns:
+        DataFrame with replacements applied.
+    """
+    if column not in df.columns:
+        raise TransformationError(f"Column '{column}' not found")
+
+    df = df.copy()
+    df[column] = df[column].astype(str).str.replace(find_value, replace_value, regex=False)
+    return df
+
+
 def drop_duplicates(df: pd.DataFrame, columns: str, keep) -> pd.DataFrame:
     """Remove duplicate rows based on specified columns.
 
@@ -563,6 +583,12 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
     elif action_type == "melt":
         params = action_details["melt_params"]
         return melt_dataframe(df, params)
+
+    elif action_type == "stringReplace":
+        column = action_details["string_replace_params"]["column"]
+        find_value = action_details["string_replace_params"]["find_value"]
+        replace_value = action_details["string_replace_params"]["replace_value"]
+        return string_replace(df, column, find_value, replace_value)
 
     else:
         logger.warning("Unknown action type in log replay: %s", action_type)
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -338,6 +338,12 @@ def string_replace(df: pd.DataFrame, column: str, find_value: str, replace_value
     if column not in df.columns:
         raise TransformationError(f"Column '{column}' not found")
 
+    if not (pd.api.types.is_string_dtype(df[column]) or pd.api.types.is_object_dtype(df[column])):
+        raise TransformationError(
+            f"Column '{column}' is not a string column (dtype: {df[column].dtype}). "
+            "Cast it to string first before using string replace."
+        )
+
     df = df.copy()
     df[column] = df[column].astype(str).str.replace(find_value, replace_value, regex=False)
     return df
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -323,6 +323,26 @@ def trim_whitespace(df: pd.DataFrame, column: str) -> pd.DataFrame:
     return df
 
 
+def string_replace(df: pd.DataFrame, column: str, find_value: str, replace_value: str) -> pd.DataFrame:
+    """Replace occurrences of a substring in a column.
+
+    Args:
+        df: Source DataFrame.
+        column: Column name to perform replacement on.
+        find_value: The substring to find.
+        replace_value: The string to replace it with.
+
+    Returns:
+        DataFrame with replacements applied.
+    """
+    if column not in df.columns:
+        raise TransformationError(f"Column '{column}' not found")
+
+    df = df.copy()
+    df[column] = df[column].astype(str).str.replace(find_value, replace_value, regex=False)
+    return df
+
+
 def drop_duplicates(df: pd.DataFrame, columns: str, keep) -> pd.DataFrame:
     """Remove duplicate rows based on specified columns.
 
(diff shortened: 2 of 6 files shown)


dataloom-backend/app/schemas.py
```diff
@@ -39,7 +39,6 @@ class OperationType(StrEnum):
     renameCol = "renameCol"
     castDataType = "castDataType"
     trimWhitespace = "trimWhitespace"
-    dropNa = "dropNa"
```
ivantha, line 47: **[CRITICAL]** `dropNa` Removed from `OperationType` Enum — Existing Feature Broken

The diff removes `dropNa` from the `OperationType` enum:

```diff
-    dropNa = "dropNa"
```

If `TransformationInput.operation_type` is typed against `OperationType`, all existing drop-NA API calls will now fail Pydantic validation with a `422 Unprocessable Entity` error. This is an unintentional regression introduced by this PR. `dropNa` remains present in `ActionTypes`, making the two enums inconsistent and suggesting the removal from `OperationType` was a mistake.

**Previous behavior**: `operation_type: "dropNa"` was a valid, accepted value.
**New behavior**: Requests with `operation_type: "dropNa"` will be rejected at the schema level.

**Suggestion:**
```
Re-add `dropNa` to `OperationType`:

```python
class OperationType(StrEnum):
    ...
    trimWhitespace = "trimWhitespace"
    dropNa = "dropNa"
    stringReplace = "stringReplace"  # also add this
```
```

File: dataloom-backend/app/schemas.py [hunk 2]
```diff
@@ -78,6 +77,7 @@ class ActionTypes(StrEnum):
     castDataType = "castDataType"
     trimWhitespace = "trimWhitespace"
     dropNa = "dropNa"
+    stringReplace = "stringReplace"
```
ivantha, line 86: **[CRITICAL]** `stringReplace` Not Added to `OperationType` — New Feature May Fail Validation

The new `stringReplace` operation is added only to `ActionTypes` but **not** to `OperationType`. If `TransformationInput.operation_type` is validated against the `OperationType` enum (the natural expectation based on its name and the pattern established by all other operations), then API requests with `operation_type: "stringReplace"` will be rejected with a `422 Unprocessable Entity` before even reaching `_handle_basic_transform`.

This means the primary feature being introduced by this PR might be entirely non-functional in production.

**Suggestion:**
```
Add `stringReplace` to `OperationType`:

```python
class OperationType(StrEnum):
    ...
    trimWhitespace = "trimWhitespace"
    dropNa = "dropNa"       # restored
    stringReplace = "stringReplace"  # added
```
```

File: dataloom-frontend/src/Components/forms/StringReplaceForm.jsx [hunk 3]
```diff
@@ -0,0 +1,108 @@
...
+    } catch (error) {
+      console.error("Error replacing string:", error);
+
+      showToast(error.response?.data?.detail || "Failed to replace string.", "error");
+    }
+
+    onClose();
+  };
+
+  return (
```
ivantha, line 37: **[HIGH]** `onClose()` Called Unconditionally — Form Closes Even on Error

In `handleSubmit`, `onClose()` is placed **outside** the `try/catch` block and will always execute — both on success and on failure:

```js
try {
  const response = await transformProject(...);
  onTransform(response);
} catch (error) {
  showToast(...);
}
onClose(); // ← always runs, even after error
```

When an error occurs, the user sees a toast notification but the form immediately closes, preventing them from correcting the input and retrying. Compare this to similar forms in the project — the error path should keep the form open.

**Suggestion:**
```
Move `onClose()` into the success path only:

```js
const handleSubmit = async (e) => {
  e.preventDefault();
  try {
    const response = await transformProject(projectId, { ... });
    onTransform(response);
    onClose(); // only close on success
  } catch (error) {
    console.error("Error replacing string:", error);
    showToast(error.response?.data?.detail || "Failed to replace string.", "error");
    // do NOT call onClose() here — let the user retry
  }
};
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 4]
```diff
@@ -488,6 +508,12 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
         columns = action_details.get("drop_na_params", {}).get("columns")
         return drop_na(df, columns)
 
+    elif action_type == "stringReplace":
+        column = action_details["string_replace_params"]["column"]
+        find_value = action_details["string_replace_params"]["find_value"]
+        replace_value = action_details["string_replace_params"]["replace_value"]
```
ivantha: **[HIGH]** Unguarded `KeyError` in Log Replay for `stringReplace`

The `apply_logged_transformation` handler for `stringReplace` uses bare dict indexing:

```python
column = action_details["string_replace_params"]["column"]
find_value = action_details["string_replace_params"]["find_value"]
replace_value = action_details["string_replace_params"]["replace_value"]
```

If any key is absent (e.g., due to a log schema migration, data corruption, or version skew), this will raise an unhandled `KeyError` and crash the replay pipeline. By contrast, the adjacent `dropNa` handler uses `.get()` with a fallback:

```python
```

This inconsistency makes the `stringReplace` replay path fragile.

**Suggestion:**
```
Use `.get()` consistently with a clear error message or fallback:

```python
elif action_type == "stringReplace":
    params = action_details.get("string_replace_params", {})
    column = params.get("column")
    find_value = params.get("find_value")
    replace_value = params.get("replace_value")
    if not column or find_value is None or replace_value is None:
        logger.warning("Missing params for stringReplace replay: %s", action_details)
        return df
    return string_replace(df, column, find_value, replace_value)
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 5]
```diff
@@ -314,6 +314,26 @@ def trim_whitespace(df: pd.DataFrame, column: str) -> pd.DataFrame:
...
+    """Replace occurrences of a substring in a column.
+
+    Args:
+        df: Source DataFrame.
+        column: Column name to perform replacement on.
+        find_value: The substring to find.
+        replace_value: The string to replace it with.
+
+    Returns:
+        DataFrame with replacements applied.
```
ivantha, line 336: **[MEDIUM]** Silent Type Coercion — Non-String Columns Converted Without Warning

The `string_replace` function unconditionally casts the column to `str` before replacement:

```python
df[column] = df[column].astype(str).str.replace(find_value, replace_value, regex=False)
```

For numeric, datetime, boolean, or other non-string columns, this silently changes the column's dtype to `object`/`str`. A user selecting a numeric column like `price` and replacing `"1"` with `"2"` would end up with a column of string values instead of numbers — a silent data corruption that may not be immediately obvious.

**Suggestion:**
```
Consider warning or raising an error for non-string columns:

```python
if not pd.api.types.is_string_dtype(df[column]) and not pd.api.types.is_object_dtype(df[column]):
    raise TransformationError(
        f"Column '{column}' is of type '{df[column].dtype}', which is not a string type. "
        "Cast it to string first using the Cast Data Type transformation."
    )
```
Alternatively, add a frontend warning when a non-string column is selected.
```

File: dataloom-backend/app/schemas.py [hunk 6]
```diff
@@ -184,6 +184,14 @@ def columns_must_not_be_empty(cls, v):
...
 
 
+class StringReplaceParams(BaseModel):
+    """Parameters for find-and-replace on a string column."""
+
+    column: str
+    find_value: str
+    replace_value: str
+
+
```
ivantha, line 207: **[MEDIUM]** Empty `find_value` Allowed — Produces Unexpected Behavior

`StringReplaceParams` does not validate that `find_value` is non-empty:

```python
class StringReplaceParams(BaseModel):
    column: str
    find_value: str
    replace_value: str
```

When `find_value` is `""` (empty string) and `regex=False`, pandas `str.replace` inserts `replace_value` between every character and at the start/end of each string. For example:
- `"hello".replace("", "X")` → `"XhXeXlXlXoX"`

This is almost certainly not the intended behavior and could corrupt data silently.

**Suggestion:**
```
Add a Pydantic validator to reject empty `find_value`:

```python

class StringReplaceParams(BaseModel):
    column: str
    find_value: str
    replace_value: str

    @field_validator("find_value")
    @classmethod
    def find_value_must_not_be_empty(cls, v: str) -> str:
        if not v:
            raise ValueError("find_value must not be empty")
        return v
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR adds a string find-and-replace transformation feature with a clean implementation across backend and frontend. However, it contains two critical bugs: `dropNa` is accidentally removed from the `OperationType` enum (breaking existing Drop NA functionality), and `stringReplace` is never added to that same enum (potentially making the new feature itself non-functional if request validation is tied to `OperationType`). Additionally, the frontend form closes unconditionally after errors, preventing users from retrying failed operations. The core `string_replace` service function is well-written with correct use of `regex=False` and defensive copying, but it needs input validation for empty `find_value` and better handling of non-string columns. The enums `OperationType` and `ActionTypes` are growing increasingly inconsistent and need documentation or consolidation.

### ⚠️ Breaking Changes

- **[HIGH]** `dropNa` Removed from `OperationType` Enum: **Previous behavior**: `OperationType` included `dropNa = "dropNa"`, allowing API requests with `operation_type: "dropNa"` to pass schema validation.

**New behavior**: `dropNa` is no longer a member of `OperationType`. If `TransformationInput.operation_type` is validated against this enum (as its name implies), all Drop NA transformation requests will be rejected with a `422 Unprocessable Entity` at the schema layer.

**Who is affected**: Any client (frontend or API consumer) that submits a Drop NA transformation request.

**Migration path**: This removal appears unintentional. The fix is to restore `dropNa` to `OperationType`. If intentional, all callers must migrate to a different operation type value, and the `_handle_basic_transform` handler for `dropNa` must also be removed.
  > **Suggestion:** Restore the removed value:
```python
class OperationType(StrEnum):
    ...
    trimWhitespace = "trimWhitespace"
    dropNa = "dropNa"          # restore this
    stringReplace = "stringReplace"  # and add this
```


### PR Description Issues

- **[LOW]** PR Description Contains Template Boilerplate and a Typo: The PR description has two issues:

1. **Typo**: "poroject" in the first line instead of "project".
2. **Unfilled template section**: The testing section retains the template placeholder text `"Describe the tests you ran to verify your changes."` with no actual description of what was manually tested (which columns, what data, what edge cases were tried).
  > **Suggestion:** Suggested description update:

```markdown
## Description
Adds a "Find & Replace" string transformation to the project. Users can select a column, specify a substring to find, and provide a replacement string. The operation is applied to all matching substrings in the column.

Fixes #82

## How Has This Been Tested?
- Manually tested on a dataset with a string column containing repeated substrings
- Verified that the replacement appears correctly in the data grid after applying
- Verified that selecting a non-existent column returns a 400 error
- Verified that canceling the form does not trigger the transformation
```

dataloom-frontend/src/Components/forms/StringReplaceForm.jsx
```diff
@@ -0,0 +1,108 @@
...
+  const handleSubmit = async (e) => {
+    e.preventDefault();
+
+    try {
+      const response = await transformProject(projectId, {
+        operation_type: "stringReplace",
+        string_replace_params: {
+          column,
+          find_value: findValue,
+          replace_value: replaceValue,
```
ivantha, line 25: **[MEDIUM]** Hard-Coded Operation Type String Instead of Exported Constant

The form uses the raw string literal `"stringReplace"` directly:

```jsx
operation_type: "stringReplace",
```

The project explicitly defines a `STRING_REPLACE` constant in `operationTypes.js` for exactly this purpose, but `StringReplaceForm.jsx` doesn't import or use it. This creates a maintenance risk — if the operation type string ever changes, the constant will be updated but the form will silently use the stale value.

**Suggestion:**
```
Import and use the constant:

```jsx
import { STRING_REPLACE } from "../../constants/operationTypes";

// ...
operation_type: STRING_REPLACE,
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 8]
```diff
@@ -314,6 +314,26 @@ def trim_whitespace(df: pd.DataFrame, column: str) -> pd.DataFrame:
...
+    if column not in df.columns:
+        raise TransformationError(f"Column '{column}' not found")
+
+    df = df.copy()
+    df[column] = df[column].astype(str).str.replace(find_value, replace_value, regex=False)
+    return df
+
+
 def drop_duplicates(df: pd.DataFrame, columns: str, keep) -> pd.DataFrame:
     """Remove duplicate rows based on specified columns.
```
ivantha, line 354: **[MEDIUM]** No Unit Tests for `string_replace` Function

The new `string_replace` service function has no automated tests. Given the edge cases present (empty `find_value`, non-string column coercion, missing column, special characters in strings), the absence of tests makes it easy for regressions to slip in undetected. The PR description also leaves the "New tests added" checkbox unchecked.

Key test cases that should be covered:
- Basic replacement in a string column
- Column not found → `TransformationError`
- Non-string column behavior (silent coercion or explicit error)
- Empty `find_value` (should be rejected)
- `find_value` not present in any row (no-op)
- Replacement in only some rows (partial match)
- Special characters in `find_value` (ensure `regex=False` prevents misinterpretation)

**Suggestion:**
```
```python
def test_string_replace_basic():
    df = pd.DataFrame({"name": ["Alice", "Bob", "Charlie"]})
    result = string_replace(df, "name", "Alice", "Alicia")
    assert result["name"].tolist() == ["Alicia", "Bob", "Charlie"]

def test_string_replace_column_not_found():
    df = pd.DataFrame({"name": ["Alice"]})
        string_replace(df, "nonexistent", "x", "y")

def test_string_replace_no_mutation():
    df = pd.DataFrame({"name": ["Alice"]})
    result = string_replace(df, "name", "Alice", "Bob")
    assert df["name"].tolist() == ["Alice"]  # original unchanged
```
```

File: dataloom-backend/app/schemas.py (hunk 2 above)
Comment (ivantha) [inline dataloom-backend/app/schemas.py:86]: **[LOW]** `OperationType` and `ActionTypes` Enums Are Diverging Without Documentation

The two enums `OperationType` and `ActionTypes` serve similar but apparently distinct purposes, yet there's no inline documentation explaining their difference. With this PR, the enums are further diverged:
- `OperationType` is missing both `dropNa` (accidentally removed) and `stringReplace` (not added)
- `ActionTypes` has both `dropNa` and `stringReplace`

Future contributors will not know which enum to update when adding new operations, leading to more inconsistencies like this.

**Suggestion:**
```
Add class-level docstrings clarifying the purpose of each enum:

```python
class OperationType(StrEnum):
    """Valid values for `TransformationInput.operation_type` in API requests."""
    ...

class ActionTypes(StrEnum):
    """Valid action types recorded in transformation logs for replay."""
    ...
```
Consider whether these can be unified into a single enum, or add a test that asserts both enums contain the same members.
```

File: dataloom-frontend/src/Components/forms/StringReplaceForm.jsx [hunk 9]
```diff
@@ -0,0 +1,108 @@
...
+            className="border border-gray-300 rounded-md w-full px-3 py-2 bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+            required
+          />
+        </div>
+
+        <div className="mb-4">
+          <label className="block text-sm font-medium text-gray-700">Replace with:</label>
+          <input
+            type="text"
+            value={replaceValue}
```
ivantha, line 82: **[LOW]** Missing Placeholder Text and Accessibility Labels on Inputs

The `Find` and `Replace with` inputs lack `placeholder` attributes and do not have associated `htmlFor`/`id` pairs linking `<label>` to `<input>`. While the visual layout is clear, this affects:
1. **Accessibility**: Screen readers rely on `for`/`id` associations to announce which label belongs to which input.
2. **UX**: Placeholder text (e.g., `"e.g. old text"`) gives users a hint about expected input format before they type.

**Suggestion:**
```
```jsx
<label htmlFor="find-value" className="block text-sm font-medium text-gray-700">
  Find:
</label>
<input
  id="find-value"
  type="text"
  value={findValue}
  onChange={(e) => setFindValue(e.target.value)}
  placeholder="Text to find..."
  className="..."
  required
/>
```
Apply the same pattern for `replace_value` and the `column` select.
```

File: dataloom-backend/app/services/transformation_service.py (hunk 8 above)
Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:354]: **[INFO]** Positive: Good Docstring, Defensive Copying, and Regex Safety

The `string_replace` function demonstrates several good practices worth highlighting:

1. **`regex=False`**: Prevents regex injection — if `find_value` were treated as a regex pattern, a user could accidentally (or maliciously) supply a broken pattern like `[unclosed` and cause a `re.error`. Using `regex=False` makes this a pure literal string replacement.
2. **`df.copy()`**: Avoids pandas `SettingWithCopyWarning` and prevents mutation of the input DataFrame, which is consistent with the functional style used elsewhere in the service.
3. **Column existence check**: Explicitly raises `TransformationError` before attempting the operation.
4. **Docstring**: Follows the established Args/Returns docstring format in the codebase.

**Suggestion:**
```
No change needed. Consider documenting the `regex=False` decision in a comment for future maintainers:

```python
# regex=False ensures find_value is treated as a literal string, not a regex pattern
df[column] = df[column].astype(str).str.replace(find_value, replace_value, regex=False)
```
```

dataloom-frontend/src/Components/forms/StringReplaceForm.jsx
hanzalahwaheed: adding this. 

plz do note that we don't have `id` and `placeholder` for some other forms as well. We can create a separate input component where this is a required parameter.

dataloom-backend/app/schemas.py
hanzalahwaheed: I think we should combine both `ActionType` and `OperationType` into a `TranformationType` as both are identical. 

Currently, `action_type` is being used only to store in logs, and `operation_type` is used in transformation backend, but the same data gets stored in logs anyways. `operation_type` values are what get persisted as action_type in logs.

i've verified the flow of log persistence and we do store the `OperationType` as an `ActionType`

do tell if I should combine them as this will create a ripple of changes where we replace `action_type` and `operation_type` with transformation_type, or was this separation intentional.

dataloom-backend/app/services/transformation_service.py
hanzalahwaheed: done

dataloom-frontend/src/Components/forms/StringReplaceForm.jsx
hanzalahwaheed: done

dataloom-backend/app/schemas.py
hanzalahwaheed: done

dataloom-backend/app/schemas.py
hanzalahwaheed: done

dataloom-frontend/src/Components/forms/StringReplaceForm.jsx
hanzalahwaheed: done

dataloom-backend/app/services/transformation_service.py
hanzalahwaheed: done

dataloom-backend/app/services/transformation_service.py
hanzalahwaheed: done

dataloom-backend/app/schemas.py
hanzalahwaheed: done

hanzalahwaheed:
@ivantha addressed all comments. need input reg this: https://github.com/c2siorg/dataloom/pull/126#discussion_r2949947932

ivantha:
Thanks for the PR! It currently has merge conflicts with main (or is behind). Could you rebase onto the latest main and push the updated branch? Happy to merge once it's auto-rebasable.

hanzalahwaheed:
@ivantha 

rebased. ready to merge.

PR #127, opened 2026-02-27
Added GroupBy Aggregation Transformation

### LINKED ISSUE #89: Add GroupBy Aggregation Transformation
**Difficulty:** Hard
**Area:** full stack

## Description

GroupBy aggregation is one of the most powerful and frequently used data analysis operations. It allows users to split their data into groups based on one or more columns, apply an aggregation function (such as sum, mean, or count) to another column within each group, and combine the results into a new summarized dataset.

DataLoom currently supports pivot tables, which are related but serve a different purpose. A dedicated GroupBy transformation provides a more straightforward interface for common aggregation tasks. For example, a user might want to group sales data by region and calculate the total revenue per region, or group employee data by department and find the average salary.

This issue adds a complete GroupBy aggregation feature spanning the backend service, API endpoint, and a new frontend form component.

## Acceptance Criteria

- [ ] A "groupby" operation type is available
- [ ] Grouping and aggregation produces correct results for the selected columns and function
- [ ] Grouping by multiple columns simultaneously is supported
- [ ] All aggregation functions work: sum, mean, count, min, max, median
- [ ] The result has a flat structure (no multi-level index)
- [ ] Invalid column names return a descriptive error
- [ ] The groupby form renders with a group column multi-select, aggregation column dropdown, and function dropdown
- [ ] GroupBy option is accessible from the side menu
- [ ] Transformation is logged in the project change log with all parameters
- [ ] Data table updates to show the aggregated result

## Resources

- `dataloom-backend/app/services/transformation_service.py` — existing transform functions
- `dataloom-backend/app/schemas.py` — `OperationType` and `AggFunc` enums
- `dataloom-backend/app/api/endpoints/transformations.py` — `_handle_complex_transform()`
- `dataloom-frontend/src/Components/forms/PivotTableForm.jsx` — reference for complex transform form
- `dataloom-frontend/src/api/transforms.js` — `complexTransformProject()`
- `dataloom-frontend/src/Components/MenuNavbar.jsx` — side menu
- [pandas `groupby` documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html)

Sharma-Ji-21:
## Description

Implemented a new GroupBy aggregation feature across backend and frontend, plus UI polishing for the preview table. Backend changes add a new `groupby` operation type, request schema, and transformation function that supports grouping by multiple columns and aggregating with `sum`, `mean`, `count`, `min`, `max`, and `median`, returning a flat (non-multi-index) result and descriptive errors for invalid columns. Frontend changes add a GroupBy form (group columns input, aggregation column input, aggregation function select) and menu entry, and update the transform preview table to match the main dataset table’s padding/scroll behavior with a slightly larger font size for readability.  

Fixes #89 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Python syntax validation with `python3 -m py_compile` on updated backend files  
- No automated test suite run  
- No manual UI run recorded

- [ ] Existing tests pass
- [ ] New tests added
- [ ] Manual testing

## Screenshots

Screenshot 2026-02-27 at 4 16 55 PM
Screenshot 2026-02-27 at 4 18 06 PM

https://github.com/user-attachments/assets/1199825e-216c-4758-a9ec-2a5c023ed6e6

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -32,6 +32,7 @@ class OperationType(str, Enum):
     dropDuplicate = 'dropDuplicate'
     advQueryFilter = 'advQueryFilter'
     pivotTables = 'pivotTables'
+    groupby = 'groupby'
     changeCellValue = 'changeCellValue'
     renameCol = 'renameCol'
     castDataType = 'castDataType'
@@ -66,6 +67,7 @@ class ActionTypes(str, Enum):
     dropDuplicate = 'dropDuplicate'
     advQueryFilter = 'advQueryFilter'
     pivotTables = 'pivotTables'
+    groupby = 'groupby'
     changeCellValue = 'changeCellValue'
     renameCol = 'renameCol'
     castDataType = 'castDataType'
@@ -158,6 +160,13 @@ class Pivot(BaseModel):
     aggfun: AggFunc
 
 
+class GroupBy(BaseModel):
+    """Parameters for creating a group by aggregation."""
+    group_columns: str
+    agg_column: str
+    agg_func: AggFunc
+
+
 class RevertRequest(BaseModel):
     """Request body for reverting to a checkpoint."""
     checkpoint_id: uuid.UUID
@@ -189,6 +198,7 @@ class TransformationInput(BaseModel):
     drop_duplicate: Optional[DropDuplicates] = None
     adv_query: Optional[AdvQuery] = None
     pivot_query: Optional[Pivot] = None
+    groupby_query: Optional[GroupBy] = None
     change_cell_value: Optional[ChangeCellValue] = None
     rename_col_params: Optional[RenameColumnParams] = None
     cast_data_type_params: Optional[CastDataTypeParams] = None
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -383,6 +383,29 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
     return result.reset_index()
 
 
+def group_by(df: pd.DataFrame, group_columns: str, agg_column: str, agg_func: str = "sum") -> pd.DataFrame:
+    """Perform a group by aggregation on the DataFrame.
+
+    Args:
+        df: Source DataFrame.
+        group_columns: Comma-separated column names to group by.
+        agg_column: Column to aggregate.
+        agg_func: Aggregation function name.
+
+    Returns:
+        Grouped and aggregated DataFrame with flat structure.
+    """
+    group_cols = [c.strip() for c in group_columns.split(',')]
+    agg_col = agg_column.strip()
+
+    missing = [c for c in group_cols + [agg_col] if c not in df.columns]
+    if missing:
+        raise TransformationError(f"Columns {missing} not found in dataset")
+
+    result = df.groupby(group_cols, as_index=False)[agg_col].agg(agg_func)
+    return result
+
+
 def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_details: dict) -> pd.DataFrame:
     """Replay a logged transformation from its serialized form.
 
--- a/dataloom-frontend/src/Components/MenuNavbar.jsx
+++ b/dataloom-frontend/src/Components/MenuNavbar.jsx
@@ -4,6 +4,7 @@ import SortForm from "./forms/SortForm";
 import DropDuplicateForm from "./forms/DropDuplicateForm";
 import AdvQueryFilterForm from "./forms/AdvQueryFilterForm";
 import PivotTableForm from "./forms/PivotTableForm";
+import GroupByForm from "./forms/GroupByForm";
 import CastDataTypeForm from "./forms/CastDataTypeForm";
 import TrimWhitespaceForm from "./forms/TrimWhitespaceForm";
 import LogsPanel from "./history/LogsPanel";
@@ -33,6 +34,7 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
   const [showDropDuplicateForm, setShowDropDuplicateForm] = useState(false);
   const [showAdvQueryFilterForm, setShowAdvQueryFilterForm] = useState(false);
   const [showPivotTableForm, setShowPivotTableForm] = useState(false);
+  const [showGroupByForm, setShowGroupByForm] = useState(false);
   const [showLogs, setShowLogs] = useState(false);
   const [showCheckpoints, setShowCheckpoints] = useState(false);
   const [showCastDataTypeForm, setShowCastDataTypeForm] = useState(false);
@@ -120,7 +122,7 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
     setShowSortForm(false);
     setShowDropDuplicateForm(false);
     setShowAdvQueryFilterForm(false);
-    setShowPivotTableForm(false);
+    setShowGroupByForm(false);
     setShowCastDataTypeForm(false);
     setShowTrimWhitespaceForm(false);
     setShowLogs(false);
@@ -141,6 +143,9 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
         break;
       case "PivotTableForm":
         setShowPivotTableForm(true);
+        break;
+      case "GroupByForm":
+        setShowGroupByForm(true);
         break;
       case "CastDataTypeForm":
         setShowCastDataTypeForm(true);
@@ -214,6 +219,11 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
             icon: LuTable2,
             onClick: () => handleMenuClick("PivotTableForm"),
           },
+          {
+            label: "GroupBy",
+            icon: LuTable2,
+            onClick: () => handleMenuClick("GroupByForm"),
+          },
         ],
       },
     ],
@@ -227,8 +237,8 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
             key={tabName}
             onClick={() => setActiveTab(tabName)}
             className={`px-4 py-1.5 text-sm font-medium ${activeTab === tabName
-                ? "text-blue-600 border-b-2 border-blue-500"
-                : "text-gray-500 hover:text-gray-700"
+              ? "text-blue-600 border-b-2 border-blue-500"
+              : "text-gray-500 hover:text-gray-700"
               }`}
           >
             {tabName}
@@ -280,6 +290,9 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
       )}
       {showPivotTableForm && (
         <PivotTableForm onClose={() => setShowPivotTableForm(false)} projectId={projectId} />
+      )}
+      {showGroupByForm && (
+        <GroupByForm onClose={() => setShowGroupByForm(false)} projectId={projectId} />
       )}
       {showCastDataTypeForm && (
         <CastDataTypeForm
--- a/dataloom-frontend/src/Components/forms/GroupByForm.jsx
+++ b/dataloom-frontend/src/Components/forms/GroupByForm.jsx
@@ -1,0 +1,108 @@
+import { useState } from "react";
+import PropTypes from "prop-types";
+import TransformResultPreview from "./TransformResultPreview";
+import { transformProject } from "../../api";
+import useError from "../../hooks/useError";
+import FormErrorAlert from "../common/FormErrorAlert";
+
+const GroupByForm = ({ projectId, onClose }) => {
+    const [groupColumns, setGroupColumns] = useState("");
+    const [aggColumn, setAggColumn] = useState("");
+    const [aggFunc, setAggFunc] = useState("sum");
+    const [result, setResult] = useState(null);
+    const [loading, setLoading] = useState(false);
+    const { error, clearError, handleError } = useError();
+
+    const handleSubmit = async (e) => {
+        e.preventDefault();
+        setLoading(true);
+        clearError();
+        try {
+            const response = await transformProject(projectId, {
+                operation_type: "groupby",
+                groupby_query: { group_columns: groupColumns, agg_column: aggColumn, agg_func: aggFunc },
+            });
+            setResult(response);
+        } catch (err) {
+            handleError(err);
+        } finally {
+            setLoading(false);
+        }
+    };
+
+    return (
+        <div className="p-4 border border-gray-200 rounded-lg bg-white">
+            <form onSubmit={handleSubmit}>
+                <h3 className="font-semibold text-gray-900 mb-2">GroupBy Aggregation</h3>
+                <div className="flex space-x-2 mb-4">
+                    <div className="flex-1">
+                        <label className="block text-sm font-medium text-gray-700">Group Columns:</label>
+                        <input
+                            type="text"
+                            value={groupColumns}
+                            onChange={(e) => setGroupColumns(e.target.value)}
+                            className="border border-gray-300 rounded-md w-full px-3 py-2 bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+                            placeholder="e.g., col1,col2"
+                            required
+                        />
+                    </div>
+                    <div className="flex-1">
+                        <label className="block text-sm font-medium text-gray-700">Aggregation Column:</label>
+                        <input
+                            type="text"
+                            value={aggColumn}
+                            onChange={(e) => setAggColumn(e.target.value)}
+                            className="border border-gray-300 rounded-md w-full px-3 py-2 bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+                            placeholder="e.g., Sales"
+                            required
+                        />
+                    </div>
+                </div>
+                <div className="flex space-x-2 mb-4">
+                    <div className="flex-1">
+                        <label className="block text-sm font-medium text-gray-700">
+                            Aggregation Function:
+                        </label>
+                        <select
+                            value={aggFunc}
+                            onChange={(e) => setAggFunc(e.target.value)}
+                            className="border border-gray-300 rounded-md w-full px-3 py-2 bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+                        >
+                            <option value="sum">Sum</option>
+                            <option value="mean">Mean</option>
+                            <option value="count">Count</option>
+                            <option value="min">Min</option>
+                            <option value="max">Max</option>
+                            <option value="median">Median</option>
+                        </select>
+                    </div>
+                </div>
+                <div className="flex justify-between">
+                    <button
+                        type="submit"
+                        className="bg-blue-500 hover:bg-blue-600 text-white px-4 py-2 rounded-md font-medium transition-colors duration-150"
+                        disabled={loading}
+                    >
+                        {loading ? "Loading..." : "Submit"}
+                    </button>
+                    <button
+                        type="button"
+                        onClick={onClose}
+                        className="bg-white border border-gray-300 text-gray-700 hover:bg-gray-50 px-4 py-2 rounded-md font-medium transition-colors duration-150"
+                    >
+                        Cancel
+                    </button>
+                </div>
+            </form>
+            <FormErrorAlert message={error} />
+            {result && <TransformResultPreview columns={result.columns} rows={result.rows} />}
+        </div>
+    );
+};
+
+GroupByForm.propTypes = {
+    projectId: PropTypes.string.isRequired,
+    onClose: PropTypes.func.isRequired,
+};
+
+export default GroupByForm;
--- a/dataloom-frontend/src/Components/forms/TransformResultPreview.jsx
+++ b/dataloom-frontend/src/Components/forms/TransformResultPreview.jsx
@@ -12,34 +12,42 @@ const TransformResultPreview = ({ columns, rows }) => {
   return (
     <div className="p-4 mt-4 border border-gray-200 rounded-lg bg-gray-50">
       <h4 className="font-medium text-sm text-gray-700 mb-2">API Response:</h4>
-      <table className="min-w-full divide-y divide-gray-200">
-        <thead className="bg-gray-100">
-          <tr>
-            {columns.map((col, index) => (
-              <th
-                key={index}
-                className="px-6 py-3 text-left text-xs font-medium text-gray-500 uppercase tracking-wider"
-              >
-                {col}
-              </th>
-            ))}
-          </tr>
-        </thead>
-        <tbody className="bg-white divide-y divide-gray-200">
-          {rows.map((row, rowIndex) => (
-            <tr key={rowIndex}>
-              {row.map((cell, cellIndex) => (
-                <td
-                  key={cellIndex}
-                  className="px-6 py-4 whitespace-nowrap text-sm text-gray-700"
+      <div
+        className="overflow-x-scroll overflow-y-auto border border-gray-200 rounded-lg shadow-sm"
+        style={{ maxHeight: "calc(100vh - 140px)" }}
+      >
+        <table className="min-w-full bg-white">
+          <thead className="bg-gray-50">
+            <tr>
+              {columns.map((col, index) => (
+                <th
+                  key={index}
+                  className="py-1.5 px-3 border-b border-gray-200 text-left text-sm font-medium text-gray-500 uppercase tracking-wider"
                 >
-                  {cell}
-                </td>
+                  {col}
+                </th>
               ))}
             </tr>
-          ))}
-        </tbody>
-      </table>
+          </thead>
+          <tbody>
+            {rows.map((row, rowIndex) => (
+              <tr
+                key={rowIndex}
+                className="border-b border-gray-100 hover:bg-gray-50 transition-colors duration-150"
+              >
+                {row.map((cell, cellIndex) => (
+                  <td
+                    key={cellIndex}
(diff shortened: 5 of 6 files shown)


dataloom-backend/app/schemas.py
```diff
@@ -158,6 +160,13 @@ class Pivot(BaseModel):
     aggfun: AggFunc
 
 
+class GroupBy(BaseModel):
+    """Parameters for creating a group by aggregation."""
+    group_columns: str
+    agg_column: str
+    agg_func: AggFunc
+
```
ivantha: **[MEDIUM]** `group_columns` typed as `str` instead of `List[str]`, requiring fragile comma-splitting

The `GroupBy` Pydantic model accepts `group_columns` as a bare `str` (e.g. `"col1,col2"`). This means:
1. There is no type-level validation — empty strings, extra spaces, trailing commas, and other malformed input all pass schema validation silently.
2. The parsing contract (comma-separated) is an implicit convention documented only in the service layer, not enforceable at the API boundary.
3. Clients must know the internal serialization format.

This contrasts with REST API best practices where structured data should be modeled as arrays.

```python
# Current — fragile
class GroupBy(BaseModel):
    group_columns: str   # "col1, col2"
```

**Suggestion:**
```
Change `group_columns` to `List[str]` and update the service function accordingly. Add a validator to reject empty lists or blank column names:
```python
from typing import List

class GroupBy(BaseModel):
    group_columns: List[str]
    agg_column: str
    agg_func: AggFunc

    @field_validator('group_columns')
    @classmethod
    def columns_must_be_non_empty(cls, v):
        if not v or any(not c.strip() for c in v):
            raise ValueError('group_columns must be a non-empty list of non-blank column names')
        return [c.strip() for c in v]
```
Update the service signature to `group_columns: List[str]` and remove the `.split(',')` logic.
```

File: dataloom-backend/app/services/transformation_service.py [hunk 2]
```diff
@@ -383,6 +383,29 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
...
+    group_cols = [c.strip() for c in group_columns.split(',')]
+    agg_col = agg_column.strip()
+
+    missing = [c for c in group_cols + [agg_col] if c not in df.columns]
+    if missing:
+        raise TransformationError(f"Columns {missing} not found in dataset")
+
+    result = df.groupby(group_cols, as_index=False)[agg_col].agg(agg_func)
+    return result
+
```
ivantha: **[MEDIUM]** Invalid `agg_func` string crashes with unhandled pandas `ValueError` (500 response)

The service function accepts `agg_func: str` without validating it against the allowed set before passing it to `pandas`. If an invalid function name reaches pandas (e.g. due to a direct service call or an enum mismatch), pandas raises a raw `ValueError` that is not caught, resulting in an unhandled 500 Internal Server Error instead of a descriptive 400 Bad Request.

Similarly, applying a numeric aggregation function (e.g. `sum`, `mean`) to a non-numeric `agg_column` (e.g. a string column) will raise a `TypeError` from pandas that is also not caught.

```python
# pandas will raise ValueError: No axis named ... or similar
result = df.groupby(group_cols, as_index=False)[agg_col].agg(agg_func)
```

**Suggestion:**
```
Wrap the aggregation call in a try/except and translate pandas exceptions to `TransformationError`:
```python
ALLOWED_FUNCS = {'sum', 'mean', 'count', 'min', 'max', 'median'}

def group_by(df, group_columns, agg_column, agg_func='sum'):
    ...
    if agg_func not in ALLOWED_FUNCS:
        raise TransformationError(f"Unsupported aggregation function: '{agg_func}'. Choose from {sorted(ALLOWED_FUNCS)}.")
    try:
        result = df.groupby(group_cols, as_index=False)[agg_col].agg(agg_func)
    except (TypeError, ValueError) as e:
        raise TransformationError(f"Aggregation failed: {e}. Ensure '{agg_col}' is compatible with '{agg_func}'.")
    return result
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR adds a GroupBy aggregation feature across the backend (new schema, service function, and endpoint handler) and frontend (new `GroupByForm` component and menu entry), along with a UI refresh of `TransformResultPreview`. The core feature logic is sound and follows the established patterns of the codebase, but there is a **high-severity regression bug** in `MenuNavbar.jsx` where the addition of `setShowGroupByForm(false)` *replaces* `setShowPivotTableForm(false)` rather than supplementing it, breaking the "close all forms" behavior for the Pivot Table. There are also several medium-severity issues including a fragile comma-delimited string schema for `group_columns`, missing pandas error handling for non-numeric columns, and a UI layout magic number, all of which need to be addressed before merging.

File: dataloom-frontend/src/Components/MenuNavbar.jsx [hunk 3]
```diff
@@ -120,7 +122,7 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
     setShowSortForm(false);
```
ivantha, line 122: **[CRITICAL]** PivotTableForm is never closed in closeAllForms — silent regression

The `closeAllForms` function had `setShowPivotTableForm(false)` **replaced** by `setShowGroupByForm(false)` instead of **adding** the new line alongside the existing one. As a result, whenever `closeAllForms` is triggered (e.g., switching menu tabs, opening another form), `showPivotTableForm` will never be reset to `false`. This means the PivotTable form can remain visible at the same time as another form, causing overlapping UI panels and broken UX for an existing, previously-working feature.

```diff
- setShowPivotTableForm(false);
+ setShowGroupByForm(false);
```
This is almost certainly an accidental line replacement rather than intentional deletion.

**Suggestion:**
```
Both lines must be present:
```js
setShowPivotTableForm(false);
setShowGroupByForm(false);
```
```

File: dataloom-frontend/src/Components/forms/GroupByForm.jsx [hunk 4]
```diff
@@ -0,0 +1,108 @@
...
+                operation_type: "groupby",
+                groupby_query: { group_columns: groupColumns, agg_column: aggColumn, agg_func: aggFunc },
+            });
+            setResult(response);
+        } catch (err) {
+            handleError(err);
+        } finally {
+            setLoading(false);
+        }
+    };
```
ivantha, line 31: **[HIGH]** No "Apply" step — transformation is previewed but never committed to the dataset

The form calls `transformProject` on submit and stores the result in `result` for preview rendering. However, there is no secondary "Apply" / "Confirm" button that would actually persist the transformed result back to the dataset. Comparing with the pattern used by other forms (e.g., PivotTableForm), most forms in this codebase require a separate apply/save action after preview. If `transformProject` here is a **preview-only** endpoint, users will be confused — they'll see the grouped result but their dataset won't change. If it's a combined preview+apply endpoint, the UX still doesn't make this clear (the button is labelled "Submit" with no confirmation).

**Suggestion:**
```
Follow the established form pattern: show the preview first, then surface an "Apply Transformation" button that calls the appropriate save/apply API or passes the result through `onTransform`. If the endpoint already applies the transform, rename the button to "Apply" and add a success notification / call `onTransform` to refresh the dataset table.
```

dataloom-frontend/src/Components/forms/GroupByForm.jsx
```diff
@@ -0,0 +1,108 @@
...
+                        <label className="block text-sm font-medium text-gray-700">
+                            Aggregation Function:
+                        </label>
+                        <select
+                            value={aggFunc}
+                            onChange={(e) => setAggFunc(e.target.value)}
+                            className="border border-gray-300 rounded-md w-full px-3 py-2 bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+                        >
+                            <option value="sum">Sum</option>
+                            <option value="mean">Mean</option>
```
ivantha, line 72: **[HIGH]** "median" offered in UI but likely absent from the backend AggFunc enum

The `GroupByForm` dropdown includes `median` as a selectable aggregation function. The backend `AggFunc` enum (used in `GroupBy.agg_func`) is **not modified in this diff** — only the reuse of an existing `AggFunc` type is shown. If the existing enum does not include `"median"`, Pydantic will reject any request with `agg_func: "median"` with a validation error, breaking median aggregation silently from the user's perspective. Even if Pydantic accepts free strings through that enum, pandas `groupby().agg('median')` raises a `FutureWarning`/error in newer versions because the correct spelling is `'median'` (it is supported, but the enum must explicitly include it).

**Suggestion:**
```
1. Verify the existing `AggFunc` enum definition and add `median = 'median'` if missing.
2. Add an integration test or at minimum a Python unit test that calls `group_by(df, ..., agg_func='median')` and asserts a correct result.
3. In `group_by()`, catch pandas exceptions around the `.agg()` call and re-raise as `TransformationError` with a human-readable message.
```

dataloom-backend/app/schemas.py
ivantha, line 168: **[HIGH]** group_columns typed as str instead of List[str] — fragile parsing in the service layer

`group_columns: str` accepts an arbitrary string such as `"col1, col2"` or `" col1 ,col2 "`. The parsing (splitting on `,` and stripping whitespace) is delegated to `transformation_service.group_by()`. This means:
- Schema validation cannot catch "zero columns" or column-name format issues at the Pydantic layer
- There is no maximum number of group columns enforced
- Leading/trailing commas (e.g., `",col1"`) produce empty-string entries that pass the split but fail with a confusing error: `Columns [''] not found in dataset`

Using `List[str]` would let Pydantic handle the list boundary conditions and remove all the string-splitting logic from the service.

**Suggestion:**
```
```python
class GroupBy(BaseModel):
    group_columns: List[str]  # e.g. ["col1", "col2"]
    agg_column: str
    agg_func: AggFunc
```
Update `group_by()` signature accordingly:
```python
def group_by(df, group_columns: List[str], agg_column: str, agg_func: str) -> pd.DataFrame:
    agg_col = agg_column.strip()
    missing = [c for c in group_columns + [agg_col] if c not in df.columns]
    ...
```
Update the frontend to send an array:
```js
groupby_query: { group_columns: groupColumns.split(',').map(s => s.trim()), ... }
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 6]
```diff
@@ -383,6 +383,29 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
...
+    Args:
+        df: Source DataFrame.
+        group_columns: Comma-separated column names to group by.
+        agg_column: Column to aggregate.
+        agg_func: Aggregation function name.
+
+    Returns:
+        Grouped and aggregated DataFrame with flat structure.
+    """
```
ivantha, line 397: **[MEDIUM]** Empty or whitespace-only column entries not filtered after comma-split

If a user submits `group_columns = "col1,,col2"` or `"col1, ,col2"`, the split produces `['col1', '', 'col2']` or `['col1', ' ', 'col2']`. After stripping, the empty string `''` will not match any column name and the error message becomes `Columns [''] not found in dataset` — confusing and non-actionable for end users. Similarly, `agg_column = " "` (all whitespace) strips to `""` and produces the same unhelpful error.

**Suggestion:**
```
```python
group_cols = [c.strip() for c in group_columns.split(',') if c.strip()]
if not group_cols:
    raise TransformationError("At least one group column must be specified")
agg_col = agg_column.strip()
if not agg_col:
    raise TransformationError("Aggregation column must not be empty")
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 7]
```diff
@@ -383,6 +383,29 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
...
+    group_cols = [c.strip() for c in group_columns.split(',')]
+    agg_col = agg_column.strip()
+
+    missing = [c for c in group_cols + [agg_col] if c not in df.columns]
+    if missing:
```
ivantha, line 402: **[MEDIUM]** Unsupported agg_func passed to pandas raises an opaque internal error

The `agg_func` string is passed directly to `pandas.DataFrameGroupBy.agg()`. If the string is not a recognised pandas aggregation function (e.g., `"variance"`, `"stdev"`, or a user typo), pandas raises a `TypeError` or `AttributeError` with an internal traceback rather than a clean user-facing error. The Pydantic schema uses `AggFunc` (an enum) so this _should_ be caught at the schema level — but if the enum is not exhaustive (see median issue above), or if `group_by()` is ever called directly from another code path, the unhandled exception will propagate as a 500.

**Suggestion:**
```
```python
try:
    result = df.groupby(group_cols, as_index=False)[agg_col].agg(agg_func)
except (TypeError, AttributeError) as e:
    raise TransformationError(f"Unsupported aggregation function '{agg_func}': {e}")
```
```

File: dataloom-backend/app/services/transformation_service.py (hunk 2 above)
Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:407]: **[MEDIUM]** group_by only aggregates a single column — major functional limitation undocumented

The function signature accepts a single `agg_column: str`, meaning users can only aggregate one column per GroupBy operation. A common real-world use case is `groupby(['Region', 'Category']).agg({'Sales': 'sum', 'Qty': 'mean'})`. The PR description says the feature supports "grouping by multiple columns" (addressed via comma-separated `group_columns`) but is silent about the one-column aggregation limitation. Users expecting multi-column aggregation will be surprised.

**Suggestion:**
```
Either document this limitation explicitly in the UI (placeholder text "Single column only") or extend the schema and service to accept multiple `agg_columns` as a list:
```python
result = df.groupby(group_cols, as_index=False)[agg_cols].agg(agg_func)
```
This naturally returns a DataFrame with one row per group and one result column per aggregation column.
```

File: dataloom-frontend/src/Components/forms/TransformResultPreview.jsx [hunk 8]
```diff
@@ -12,34 +12,42 @@ const TransformResultPreview = ({ columns, rows }) => {
...
-              {row.map((cell, cellIndex) => (
-                <td
-                  key={cellIndex}
-                  className="px-6 py-4 whitespace-nowrap text-sm text-gray-700"
+      <div
+        className="overflow-x-scroll overflow-y-auto border border-gray-200 rounded-lg shadow-sm"
+        style={{ maxHeight: "calc(100vh - 140px)" }}
+      >
+        <table className="min-w-full bg-white">
+          <thead className="bg-gray-50">
```
ivantha, line 20: **[MEDIUM]** Hardcoded maxHeight with magic number creates fragile layout across different contexts

`style={{ maxHeight: "calc(100vh - 140px)" }}` is a magic number that assumes the preview table is rendered at a fixed offset from the top of the viewport. This assumption will break if the component is used inside a modal with different chrome height, or in a sidebar context, or if the page layout changes. The value `140px` has no explanation in comments.

**Suggestion:**
```
Pass `maxHeight` as a prop with a sensible default, or use a CSS variable/design token:
```jsx
const TransformResultPreview = ({ columns, rows, maxHeight = '60vh' }) => (
  ...
  <div className="overflow-x-auto overflow-y-auto ..." style={{ maxHeight }}>
```
Also change `overflow-x-scroll` to `overflow-x-auto` (see next finding).
```

Comment (ivantha) [inline dataloom-frontend/src/Components/forms/TransformResultPreview.jsx:20]: **[LOW]** overflow-x-scroll always shows a horizontal scrollbar even when unnecessary

`overflow-x-scroll` forces the browser to render a horizontal scrollbar at all times, even when the table content fits within the container. This wastes vertical space and looks unpolished. `overflow-x-auto` only shows the scrollbar when content overflows, which is the standard pattern.

**Suggestion:**
```
```jsx
className="overflow-x-auto overflow-y-auto border border-gray-200 rounded-lg shadow-sm"
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR implements a GroupBy aggregation feature with a solid overall structure that follows existing codebase patterns, but contains a **critical regression bug**: `setShowPivotTableForm(false)` was accidentally deleted from `closeAllForms` and replaced with `setShowGroupByForm(false)`, meaning the PivotTable form can no longer be dismissed — a silent breakage of existing functionality. There are also high-severity concerns around the `median` aggregation function potentially not being present in the existing `AggFunc` enum (causing silent API failures), and the absence of an "Apply" step that actually persists the transformation to the dataset. Additionally, the `group_columns` schema type (`str` instead of `List[str]`) pushes parsing logic into the service layer without proper validation, and no automated or manual tests were run. The UI polish changes to `TransformResultPreview` are reasonable but introduce a hardcoded magic-number `maxHeight` and use `overflow-x-scroll` where `auto` is more appropriate.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes — all additions are additive and backward-compatible: **Previous behavior**: The API accepted a defined set of operation types and transformation inputs without a `groupby` path.

**New behavior**: `groupby` is added as a new enum value in `OperationType`/`ActionTypes`, a new optional field `groupby_query` is added to `TransformationInput`, and a new route branch handles the operation.

**Who is affected**: No existing consumers are affected — new enum values do not break exhaustive enum switches in Python (Pydantic handles unknowns), the new field is `Optional` with a `None` default, and the new COMPLEX_OPERATIONS entry only adds a case.

**Migration path**: None required. This is a purely additive change.

SuggestedSemver: `minor` — new capability added with no removals or incompatible changes.


### PR Description Issues

- **[MEDIUM]** Testing section is honest but reveals missing test coverage — risk not mitigated: The PR description explicitly states:
- `python3 -m py_compile` (syntax-only, not runtime)
- No automated test suite run
- No manual UI run recorded

None of the testing checklist items are checked. This is appreciated honesty, but it means the PR is merging a new feature with zero functional test coverage. Given the bugs found in this review (closeAllForms regression, potential enum mismatch for `median`), at least some manual testing should be required before merge.
  > **Suggestion:** Update the testing section to require at minimum:
- Manual walkthrough of the GroupBy UI end-to-end
- Verification that PivotTable form still closes correctly (regression check)
- One passing test for `group_by()` service function

Example improved checklist:
```markdown
## How Has This Been Tested?
- [x] `python3 -m py_compile` on modified backend files
- [ ] Unit test added for `group_by()` service function
- [ ] Manual UI test: open GroupBy form, submit, verify result
- [ ] Regression test: verify PivotTable form still closes correctly
- [ ] Tested with median aggregation to confirm enum support
```
- **[LOW]** PR description omits apply-vs-preview distinction and dataset persistence semantics: The description says the form "supports grouping by multiple columns and aggregating" but does not clarify whether submitting the GroupBy form permanently modifies the dataset or only shows a preview. This is a critical UX question that reviewers and future maintainers need to understand. The absence of an "Apply" button in the form (see separate finding) makes this ambiguity more significant.
  > **Suggestion:** Add a section to the description:
```markdown
## Behavior
The GroupBy form **previews** the aggregated result without modifying the dataset. 
To persist the result, users must click "Apply" (to be implemented in follow-up #XX).
```
or, if submit does persist:
```markdown
Submitting the form applies the transformation and updates the active dataset in place. 
The preview table shows the new dataset state.
```

dataloom-frontend/src/Components/MenuNavbar.jsx
```diff
@@ -214,6 +219,11 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
             icon: LuTable2,
             onClick: () => handleMenuClick("PivotTableForm"),
           },
+          {
+            label: "GroupBy",
+            icon: LuTable2,
```
ivantha, line 224: **[LOW]** GroupBy menu item reuses PivotTable icon — visually indistinguishable

Both "PivotTable" and "GroupBy" menu entries use `LuTable2` as their icon. While they are conceptually related, users benefit from distinct icons to quickly distinguish operations. This is a minor UX concern but can cause confusion in an icon-heavy menu.

**Suggestion:**
```
Consider using a more semantically appropriate icon such as `LuGroup`, `LuLayers`, or `LuFunctionSquare` from lucide-react for the GroupBy operation:
```jsx
import { LuTable2, LuGroup } from "react-icons/lu";
// ...
{ label: "GroupBy", icon: LuGroup, onClick: () => handleMenuClick("GroupByForm") }
```
```

File: dataloom-backend/app/services/transformation_service.py (hunk 2 above)
Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:407]: **[LOW]** No automated tests for group_by service function or API endpoint

The PR author explicitly acknowledges that no automated tests were run and no new tests were added. The `group_by()` function has several edge cases (missing columns, empty group_cols, unsupported agg_func, single-row groups, NaN values in group or agg columns) that should be covered by unit tests. Without tests, regressions in the groupby logic will be undetected until a user reports a bug.

**Suggestion:**
```
Add pytest unit tests covering at minimum:
- Happy path: valid columns, each supported agg_func
- `TransformationError` raised for missing group column
- `TransformationError` raised for missing agg column
- Empty group_columns string
- NaN values in group columns (pandas drops NaN groups by default — is this the desired behaviour?)
- Result is flat (no MultiIndex)

```python
def test_group_by_sum():
    df = pd.DataFrame({'region': ['A', 'A', 'B'], 'sales': [10, 20, 30]})
    result = group_by(df, 'region', 'sales', 'sum')
    assert list(result.columns) == ['region', 'sales']
    assert result[result.region == 'A']['sales'].iloc[0] == 30
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 10]
```diff
@@ -383,6 +383,29 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
     return result.reset_index()
 
 
+def group_by(df: pd.DataFrame, group_columns: str, agg_column: str, agg_func: str = "sum") -> pd.DataFrame:
+    """Perform a group by aggregation on the DataFrame.
+
```
ivantha, line 388: **[LOW]** Docstring does not mention comma-separated format for group_columns

The docstring says `group_columns: Comma-separated column names to group by` but the type annotation says `str`. Given that this parsing contract is invisible to callers (and will change if the type is fixed to `List[str]`), the docstring should at minimum document the expected format, constraints, and what happens on malformed input.

**Suggestion:**
```
```python
Args:
    group_columns: Comma-separated column names to group by (e.g. "col1,col2").
                   Whitespace around commas is stripped. At least one column required.
    agg_column: Name of the column to aggregate. Must exist in the DataFrame.
    agg_func: Aggregation function — one of: sum, mean, count, min, max, median.

Raises:
    TransformationError: If any specified column is not found, or if group_columns is empty.
```
```

File: dataloom-backend/app/schemas.py [hunk 11]
```diff
@@ -32,6 +32,7 @@ class OperationType(str, Enum):
     dropDuplicate = 'dropDuplicate'
     advQueryFilter = 'advQueryFilter'
     pivotTables = 'pivotTables'
+    groupby = 'groupby'
```
ivantha, line 35: **[INFO]** OperationType and ActionTypes enums are duplicated — potential drift risk

Both `OperationType` and `ActionTypes` enums contain identical members (including the new `groupby`). This is pre-existing duplication, not introduced by this PR, but the PR adds to it. Any future operation addition requires updating both enums and it's easy to forget one. This PR correctly updated both, but the pattern is fragile.

**Suggestion:**
```
Consider consolidating to a single enum, or having one inherit/alias the other. At minimum, add a comment pointing maintainers to the other enum that must be kept in sync:
```python
# NOTE: Keep ActionTypes in sync with OperationType below
class OperationType(str, Enum):
    ...
```
```

File: dataloom-backend/app/services/transformation_service.py (hunk 7 above)
Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:402]: **[INFO]** NaN group handling is implicit — pandas silently drops NaN groups by default

`df.groupby(group_cols, as_index=False)` uses pandas' default `dropna=True`, which silently drops any rows where a group column contains `NaN`. This may be surprising to users who expect NaN to appear as its own group or to raise an error. No validation or documentation covers this behaviour.

**Suggestion:**
```
Decide on the intended behaviour and make it explicit:
```python
# Option A: preserve NaN as a group
result = df.groupby(group_cols, as_index=False, dropna=False)[agg_col].agg(agg_func)
# Option B: current implicit drop — document it
# NaN values in group columns are excluded from results (pandas default)
```
Consider surfacing this choice as an optional parameter or at least documenting it in the UI tooltip.
```

File: dataloom-frontend/src/Components/forms/GroupByForm.jsx [hunk 12]
```diff
@@ -0,0 +1,108 @@
...
+                            required
+                        />
+                    </div>
+                    <div className="flex-1">
+                        <label className="block text-sm font-medium text-gray-700">Aggregation Column:</label>
+                        <input
+                            type="text"
+                            value={aggColumn}
+                            onChange={(e) => setAggColumn(e.target.value)}
+                            className="border border-gray-300 rounded-md w-full px-3 py-2 bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
```
ivantha, line 55: **[INFO]** Positive: form follows established patterns well — good code reuse

The `GroupByForm` component correctly reuses `useError`, `FormErrorAlert`, `TransformResultPreview`, and `transformProject` from the existing codebase rather than reimplementing them. The PropTypes definition is complete and the loading state is handled correctly. The inline Tailwind classes are consistent with the rest of the project. This is a solid structural foundation for the new feature.

**Suggestion:**
```
No change needed — this is a positive observation recognising good adherence to established patterns.
```

ivantha:
Thanks for the PR! CI is currently failing on Backend Lint and Frontend Lint. Could you run `uv run ruff check . && uv run ruff format --check .` in `dataloom-backend/` and `npm run lint` in `dataloom-frontend/` locally, fix the reported issues, and push? Happy to re-review once CI is green.

ivantha:
Thank you for adding GroupBy aggregation! This work has been superseded by #212 which implements the same feature. Feel free to follow up on that PR.

PR #134, opened 2026-02-27
fix(backend): update last_modified on transformation and checkpoint

### LINKED ISSUE #66: [Bug] last_modified timestamp never updates: "Recent Projects" ordering is broken
### Describe the bug
The `last_modified` field on the Project model is only set at creation time via `server_default=func.now()`. No code ever updates it afterward. This causes the "Recent Projects" list on the homepage (sorted via /projects/recent) to always show projects in upload order instead of actual last-activity order.

### To Reproduce
1. Upload Project A, wait a moment, then upload Project B.
2. Go to homepage: Project B appears first.
3. Open Project A, apply transformations, save a checkpoint.
4. Go back to homepage.
5. Observe: Project B is still shown first, even though A was modified more recently.

### Expected behavior
After modifying Project A, it should appear above Project B in the "Recent Projects" list.

### Screenshots
Homepage after modifying Project Alpha, ordering is unchanged (Project Beta still first):

Image

### Environment:
- OS: Windows 11
- Python version: 3.12
- Browser: Chrome

### Additional context
Root cause in app/models.py:  
```
last_modified: datetime | None = Field(
    default=None,
    sa_column=Column(DateTime, server_default=func.now()),  # only set on INSERT
) 
```
A search across the entire backend confirms last_modified is read for ordering but never updated after creation.  

### Suggested Fix  
- Update last_modified = datetime.now() in log_transformation() and create_checkpoint() in project_service.py
- Optionally add onupdate=func.now() to the column definition

nodesagar:
# fix(backend): update `last_modified` on transformation and checkpoint

**Closes #66**

## Summary
`last_modified` was only set at insert time, so Recent Projects ordering reflected upload time instead of actual activity. This PR updates `last_modified` on transformation/checkpoint writes and adds deterministic regression coverage.

## Root Cause
`get_recent_projects()` orders by `Project.last_modified`, but service paths that mutate project data (`log_transformation`, `create_checkpoint`) did not reliably advance it.

### `dataloom-backend/app/services/project_service.py`
- Added `_touch_project(db, project_id)` as the single explicit path to update `last_modified`.
- Uses direct bulk `UPDATE` for efficiency and raises `ValueError` when `project_id` is missing.
- Added rollback-safe transaction handling in `log_transformation` and `create_checkpoint` to avoid dirty-session leakage on error.
- Documented clock-source tradeoff (`datetime.now(timezone.utc)` vs DB clock) in helper docstring.

### `dataloom-backend/app/models.py`
- `last_modified` remains `DateTime(timezone=True)`.
- Removed misleading `onupdate=func.now()` because this PR uses bulk updates (`Query.update`) where `onupdate` does not fire.
- Reverted `upload_date` to timezone-naive `DateTime` (intentional: avoid unrelated schema side effect).

### `dataloom-backend/alembic/versions/c4d5e6f7a8b9_make_project_datetimes_timezone_aware.py`
- Added migration for existing data.
- Scope is now **only** `last_modified` (not `upload_date`).
- SQLite path normalizes legacy naive `last_modified` values by appending `+00:00`.

### `dataloom-backend/tests/test_save_revert.py`
- Replaced timing-sensitive patterns with explicit timestamp setup.
- Added/expanded regression tests for:
  - transformation updates `last_modified`
  - checkpoint updates `last_modified`
  - recent project ordering after modifying older project
  - missing-project error paths
  - no orphan log/checkpoint persistence after error
  - clean session state after missing-project exceptions
  - `upload_date` remains timezone-naive
  - dedicated `server_default` population check for `last_modified`
- Clarified misleading assertion messages around explicit timestamp assignment.

### Other
- `.gitignore`: scoped DB ignore patterns to backend paths.
- Removed committed `dataloom-backend/test.db`.
- `CONTRIBUTING.md`: added a short backend regression-test pattern section.

## Behavioral Notes / Compatibility
- `log_transformation` and `create_checkpoint` now explicitly raise `ValueError` for missing projects.
- Existing deployments with data should run the migration so legacy `last_modified` values are normalized before mixed naive/aware comparisons.

## Validation
- Backend lint/format:
  - `cd dataloom-backend && uv run ruff check .`
  - `cd dataloom-backend && uv run ruff format --check .`
- Backend tests:
  - `cd dataloom-backend && uv run pytest -q` -> `103 passed`
- Frontend checks:
  - `cd dataloom-frontend && npm run lint`
  - `cd dataloom-frontend && npx prettier --check "src/**/*.{js,jsx,css,md}"`
  - `cd dataloom-frontend && npm test -- --run` -> `48 passed`


ivantha (commented):
## ◎ Argus Review

The PR correctly identifies and fixes the root cause of the `last_modified` ordering bug: the timestamp was only ever set by `server_default` on INSERT and never updated. The fix is functionally sound and the regression tests validate the before/after behaviour. However, the implementation introduces an extra SELECT per operation where a direct UPDATE or SQLAlchemy's `onupdate` hook would be more idiomatic, a silent no-op when the project doesn't exist (masking potential integrity issues), a subtle timezone inconsistency between creation and update paths, and two of the three new tests rely on `time.sleep(0.01)` which is a fragile pattern for CI environments.

dataloom-backend/app/models.py
```diff
@@ -25,11 +25,15 @@ class Project(SQLModel, table=True):
...
     last_modified: datetime | None = Field(
         default=None,
-        sa_column=Column(DateTime, server_default=func.now()),
+        sa_column=Column(
+            DateTime(timezone=True),
+            server_default=func.now(),
+            onupdate=func.now(),
+        ),
     )
     file_path: str
```
ivantha, line 39: **[HIGH]** Schema change from DateTime to DateTime(timezone=True) requires database migration

Both `upload_date` and `last_modified` columns were changed from `Column(DateTime, ...)` to `Column(DateTime(timezone=True), ...)`. This is a schema-level change.

**Previous behavior:** Datetimes stored as timezone-naive values in the database.

**New behavior:** SQLAlchemy now expects/returns timezone-aware datetimes. In SQLite, `DateTime(timezone=True)` stores values as ISO strings with offset info. Existing rows with naive datetime strings may be read back as naive datetimes, causing `TypeError` or incorrect comparisons when mixed with the new timezone-aware values set by `datetime.now(UTC)`.

**Who is affected:** Any existing deployment with data already stored. Any code that compares or processes `upload_date` or `last_modified` values.

**Migration path:** A migration script (e.g., Alembic revision) must be provided to backfill existing rows, normalising stored datetimes to UTC ISO format. Without it, `project.last_modified > original_ts` comparisons may throw `TypeError: can't compare offset-naive and offset-aware datetimes` on real data.

No migration script is included in this PR.

**Suggestion:**
```
Add an Alembic migration that normalises all existing datetime strings to UTC. Additionally, pin a minimum supported database version or document the manual upgrade path:
```python
# alembic/versions/xxxx_make_datetimes_timezone_aware.py
def upgrade():
    # SQLite: re-store all datetime strings with +00:00 suffix
    op.execute("""
        UPDATE project
        SET upload_date = upload_date || '+00:00'
        WHERE upload_date IS NOT NULL
          AND upload_date NOT LIKE '%+%'
    """)
    # repeat for last_modified
```
```

File: dataloom-backend/app/services/project_service.py [hunk 2]
```diff
@@ -1,6 +1,7 @@
 """Database operations for projects, logs, and checkpoints."""
 
 import uuid
+from datetime import UTC, datetime
```
ivantha: **[MEDIUM]** from datetime import UTC requires Python 3.11+

The `UTC` singleton (`datetime.UTC`) was introduced in **Python 3.11** (PEP 689). Using `from datetime import UTC` will raise an `ImportError` on Python 3.9 or 3.10, which are still within their support windows and commonly used in production environments.

Interestingly, the PR description shows `from datetime import datetime, timezone` and `datetime.now(timezone.utc)` — the cross-version-compatible form — but the actual diff uses the 3.11-only shorthand. This discrepancy suggests the description was drafted before the final implementation.

**Suggestion:**
```
Use the version-compatible import unless the project already enforces Python ≥ 3.11:
```python
# Compatible with Python 3.9+
from datetime import datetime, timezone

# then use:
datetime.now(timezone.utc)  # instead of datetime.now(UTC)
```
If Python 3.11+ is already a documented requirement, add it to `pyproject.toml` / `setup.cfg` and close this concern.
```

File: dataloom-backend/app/services/project_service.py [hunk 3]
```diff
@@ -85,6 +102,9 @@ def log_transformation(db: Session, project_id: uuid.UUID, operation_type: str,
         action_details=details,
     )
     db.add(log)
+
+    _touch_project(db, project_id)
+
     db.commit()
     logger.debug("Logged transformation: project_id=%s, type=%s", project_id, operation_type)
```
ivantha, line 124: **[MEDIUM]** Session left in dirty state if _touch_project raises after db.add(log)

In `log_transformation`, a new `TransformationLog` object is added to the session with `db.add(log)` **before** `_touch_project` is called. If `_touch_project` raises `ValueError` (project not found), the exception propagates up without a `db.rollback()`. The log object remains in the session's identity map in a pending state.

If the caller reuses the same `db` session after catching the error — which is common in FastAPI dependency-injected sessions — the orphaned pending log may be flushed in a subsequent, unrelated commit, silently persisting a log entry for a non-existent project.

A similar issue exists in `create_checkpoint` where the checkpoint object is added before `_touch_project`.

**Suggestion:**
```
Validate the project exists **before** adding any objects to the session, or wrap the block in explicit rollback logic:
```python
def log_transformation(
    db: Session, project_id: uuid.UUID, operation_type: str, details: dict
) -> None:
    # Validate first — no session objects created yet
    _touch_project(db, project_id)   # raises ValueError early if missing

    log = models.TransformationLog(
        project_id=project_id,
        operation_type=operation_type,
        action_details=details,
    )
    db.add(log)
    db.commit()
```
Alternatively, explicitly call `db.rollback()` in an except block, but early validation is cleaner.
```

File: dataloom-backend/app/models.py [hunk 4]
```diff
@@ -25,11 +25,15 @@ class Project(SQLModel, table=True):
...
```
ivantha, line 38: **[MEDIUM]** onupdate=func.now() does not fire for Query.update() bulk operations

The model adds `onupdate=func.now()` to the `last_modified` column:
```python
Column(DateTime(timezone=True), server_default=func.now(), onupdate=func.now())
```
SQLAlchemy's `Column.onupdate` is an **ORM unit-of-work** hook — it fires when SQLAlchemy constructs UPDATE SQL for a tracked ORM object (e.g., `project.name = "x"; db.commit()`). It does **not** fire for bulk-style updates issued via `Query.update({...})`.

`_touch_project` uses exactly this bulk-update path:
```python
db.query(models.Project).filter(...).update({"last_modified": datetime.now(UTC)})
```
So `onupdate` is effectively a dead code path for the primary write mechanism in this PR. It will only trigger on direct ORM attribute mutations — which may or may not be a desired side effect (see the finding below).

This creates a misleading model definition that future developers may rely on incorrectly.

**Suggestion:**
```
Either document clearly that `onupdate` is intentional for ORM-path updates and does NOT apply to bulk updates, or remove it and rely solely on the explicit `_touch_project` call:
```python
# Option A: remove onupdate, make the intent explicit
last_modified: datetime | None = Field(
    default=None,
    sa_column=Column(
        DateTime(timezone=True),
        server_default=func.now(),
        # No onupdate — last_modified is managed explicitly via _touch_project
    ),
)

# Option B: keep it but add a comment
    onupdate=func.now(),  # fires on ORM-unit-of-work updates only; bulk updates use _touch_project
```
```

Comment (ivantha) [inline dataloom-backend/app/models.py:38]: **[LOW]** onupdate=func.now() causes last_modified to change on any project field mutation

Because `onupdate=func.now()` fires on all ORM-level UPDATE statements for the `Project` model, any field change — renaming a project, updating its description, or any future field — will automatically bump `last_modified`. This may be **desirable** (treating any project mutation as activity), but it is undocumented and untested.

The PR description only describes updating `last_modified` on transformations and checkpoints. A user renaming a project would silently cause it to appear at the top of the Recent Projects list, which may or may not be the intended UX.

**Suggestion:**
```
Make this a deliberate, documented choice. Add a comment to the model and a test:
```python
# last_modified auto-updates on any ORM-level project change (name, description, etc.)
# This is intentional: any project modification counts as recent activity.
onupdate=func.now(),
```
And add a test:
```python
def test_renaming_project_updates_last_modified(db):
    project = models.Project(name="old", file_path="/tmp/a.csv")
    db.add(project); db.commit(); db.refresh(project)
    original_ts = project.last_modified
    project.name = "new"
    db.commit(); db.refresh(project)
    assert project.last_modified > original_ts
```
```

File: dataloom-backend/app/models.py [hunk 5]
```diff
@@ -25,11 +25,15 @@ class Project(SQLModel, table=True):
     description: str | None = None
     upload_date: datetime | None = Field(
         default=None,
-        sa_column=Column(DateTime, server_default=func.now()),
+        sa_column=Column(DateTime(timezone=True), server_default=func.now()),
     )
     last_modified: datetime | None = Field(
         default=None,
```
ivantha, line 32: **[LOW]** upload_date column silently changed to timezone-aware — undocumented side effect

`upload_date` was also changed from `Column(DateTime, ...)` to `Column(DateTime(timezone=True), ...)`. This change is not mentioned anywhere in the PR description, the summary, or the checklist. It is a side effect of the fix rather than an intentional change.

This affects:
- Existing stored `upload_date` values (same migration concern as `last_modified`)
- Any code that compares `upload_date` with Python datetimes
- API responses that serialize `upload_date`

Since this was not intentional, it should either be reverted (fix only `last_modified`) or explicitly acknowledged and tested.

**Suggestion:**
```
If `upload_date` timezone-awareness is not intentional, revert it:
```python
upload_date: datetime | None = Field(
    default=None,
    sa_column=Column(DateTime, server_default=func.now()),  # unchanged
)
```
If it is intentional, add it to the PR description and include migration + tests.
```

File: dataloom-backend/tests/test_save_revert.py [hunk 6]
```diff
@@ -41,3 +47,111 @@ def test_checkpoint_message(self, db):
...
+        missing_id = uuid.uuid4()
+        with pytest.raises(ValueError, match=str(missing_id)):
+            create_checkpoint(db, missing_id, "save point")
+
+    def test_recent_projects_ordering_after_transformation(self, db):
+        """Recent projects should reorder after modifying an older project (issue #66).
+
+        Reproduces the exact scenario from the bug report:
+        - Upload Project A, then Project B (B is newer)
+        - Project B appears first (most recently uploaded)
```
ivantha, line 201: **[LOW]** test_log_transformation_missing_project_raises may leave session dirty

The test calls `log_transformation(db, missing_id, "addRow", {})` inside `pytest.raises(ValueError)`. Inside `log_transformation`, a `TransformationLog` object is created and `db.add(log)` is called before `_touch_project` raises. The test captures the exception, but the test's `db` session still has the pending (uncommitted) log object.

If the test fixture rolls back the session between tests (common for isolation), this is harmless. But if the session is committed or reused in a way that flushes pending objects, the orphaned log could cause a FK violation or pollute subsequent test state. It's worth verifying the `db` fixture behaviour explicitly.

**Suggestion:**
```
After the `pytest.raises` block, assert the session has no pending objects, or add an explicit expunge:
```python
def test_log_transformation_missing_project_raises(self, db):
    missing_id = uuid.uuid4()
    with pytest.raises(ValueError, match=str(missing_id)):
        log_transformation(db, missing_id, "addRow", {})
    db.rollback()  # Explicit cleanup for clarity
    # Or assert no log was persisted:
    assert db.query(models.TransformationLog).count() == 0
```
```

File: dataloom-backend/tests/test_save_revert.py [hunk 7]
```diff
@@ -41,3 +47,111 @@ def test_checkpoint_message(self, db):
...
+        )
+        db.add(project)
+        db.commit()
+        db.refresh(project)
+
+        original_ts = project.last_modified
+        assert original_ts is not None, "server_default did not populate last_modified on creation"
+
+        log_transformation(db, project.project_id, "addRow", {"row_params": {"index": 0}})
+
```
ivantha, line 72: **[LOW]** Regression tests assert server_default populated last_modified on creation — unreliable with explicit value

In `test_log_transformation_updates_last_modified` and `test_create_checkpoint_updates_last_modified`, the project is created with an explicit `last_modified=before`. Then the test asserts:
```python
assert original_ts is not None, "server_default did not populate last_modified on creation"
```
This assertion message is misleading — `last_modified` is not None because it was **explicitly set**, not because `server_default` fired. The `server_default` would only matter if no value was provided. The comment creates a false impression that the test validates `server_default` behaviour.

**Suggestion:**
```
Either fix the assertion message or add a separate, dedicated test for `server_default`:
```python
# Fix misleading message:
assert original_ts is not None, "last_modified was unexpectedly None after explicit assignment"

# Or add a dedicated server_default test:
def test_server_default_populates_last_modified(self, db):
    project = models.Project(name="test", file_path="/tmp/x.csv")  # no explicit last_modified
    db.add(project); db.commit(); db.refresh(project)
    assert project.last_modified is not None, "server_default did not populate last_modified"
```
```

File: dataloom-backend/app/services/project_service.py [hunk 8]
```diff
@@ -70,6 +71,22 @@ def delete_project(db: Session, project: models.Project) -> None:
...
+        ValueError: If the project_id does not match any existing project.
+    """
+    # Use UTC explicitly; server_default also targets UTC via func.now()
+    rows_updated = (
+        db.query(models.Project)
+        .filter(models.Project.project_id == project_id)
+        .update({"last_modified": datetime.now(UTC)})
+    )
+    if rows_updated == 0:
+        raise ValueError(f"Project {project_id} not found")
```
ivantha, line 93: **[LOW]** _touch_project uses datetime.now(UTC) while model server_default uses func.now() — potential clock skew

`_touch_project` sets `last_modified` using the **application server's clock** (`datetime.now(UTC)`), while `server_default=func.now()` and `onupdate=func.now()` use the **database engine's clock** (SQLite's `CURRENT_TIMESTAMP`). In this project (SQLite embedded), both clocks are the same process, so there's no issue today.

However, if the project ever migrates to a networked database (PostgreSQL, MySQL), the two clocks could diverge by milliseconds or more, causing inconsistent timestamps. It's a latent portability concern worth documenting.

**Suggestion:**
```
Add a comment and consider standardising on one clock source:
```python
def _touch_project(db: Session, project_id: uuid.UUID) -> None:
    """...
    Note: uses application-side UTC clock. If migrating to a networked DB,
    consider using func.now() via a server_onupdate to avoid app/DB clock skew.
    """
    rows_updated = (
        db.query(models.Project)
        .filter(models.Project.project_id == project_id)
        .update({"last_modified": datetime.now(UTC)})
    )
```
```

File: dataloom-backend/tests/test_save_revert.py [hunk 9]
```diff
@@ -41,3 +47,111 @@ def test_checkpoint_message(self, db):
...
+        assert recent[0].name == "Project B"
+        assert recent[1].name == "Project A"
+
+        # Modify Project A via a transformation — last_modified should jump to now
+        log_transformation(db, project_a.project_id, "addRow", {"row_params": {"index": 0}})
+
+        # Now Project A should be first because its last_modified was just updated
+        recent = get_recent_projects(db, limit=2)
+        assert recent[0].name == "Project A"
+        assert recent[1].name == "Project B"
```
ivantha, line 242: **[INFO]** Positive: Excellent regression test coverage with clear reproduction of the original bug

The `TestLastModified` class is well-structured and covers the exact scenario from the bug report in `test_recent_projects_ordering_after_transformation`. The use of **explicit timestamps** with `timedelta` offsets to work around SQLite's second-level precision is a thoughtful, pragmatic solution. The docstring on that test clearly reproduces the bug report table, making it easy to understand why the test exists. Tests for both missing-project error paths are also included.

**Suggestion:**
```
No change needed. Consider adding this pattern as a template for future regression tests in the project's CONTRIBUTING guide.
```

dataloom-backend/app/services/project_service.py
ivantha, line 93: **[INFO]** Positive: _touch_project helper is clean, efficient, and follows single-responsibility

Extracting the `last_modified` update into a private `_touch_project` helper is a good design decision. It:
- Uses a single bulk `Query.update()` call (one SQL statement, no ORM overhead)
- Validates existence via `rows_updated == 0` check, providing early feedback
- Is reusable across `log_transformation` and `create_checkpoint` without duplication
- Has a clear docstring with exception contract

The naming convention (underscore prefix) correctly signals it is a module-private function not intended for direct use by service consumers.

**Suggestion:**
```
No change needed. As a future enhancement, if more service methods need to track project activity, `_touch_project` is already in the right place to be reused.
```

.gitignore
```diff
@@ -25,3 +25,7 @@ Desktop.ini
 .env
 .env.local
 .env.*.local
+
+# SQLite databases (auto-generated during test runs)
+dataloom-backend/*.db
+dataloom-backend/tests/*.db
```
ivantha, line 31: **[INFO]** Positive: Removing committed test.db and updating .gitignore prevents recurrence

The `test.db` file was previously committed to the repository, which is a common mistake with SQLite-backed test suites. This PR correctly:
1. Deletes the committed binary file
2. Adds both `dataloom-backend/*.db` and `dataloom-backend/tests/*.db` to `.gitignore`

This prevents the file from appearing as a dirty working tree entry on future test runs.

**Suggestion:**
```
Consider also adding a pre-commit hook or CI check that fails if `*.db` files are staged, as a belt-and-suspenders measure.
```

ivantha (commented):
## ◎ Argus Review

This is a well-motivated bug fix for a real issue — `last_modified` never advancing after project creation — with clear root cause analysis and solid regression tests. However, the implementation introduces several concerns beyond the stated scope: the schema change from `DateTime` to `DateTime(timezone=True)` (applied silently to `upload_date` as well as `last_modified`) has no accompanying migration and will cause `TypeError` on mixed naive/aware comparisons against existing data. The new `ValueError` exception contract in `log_transformation` and `create_checkpoint` is a silent breaking change for callers. Additionally, the `from datetime import UTC` import requires Python 3.11+, and the `onupdate=func.now()` model hook does not fire for the bulk-update path used by `_touch_project`, creating a misleading model definition. The PR description unfortunately describes an earlier draft of the implementation rather than what was actually merged, omitting the model changes entirely.

### ⚠️ Breaking Changes

- **[HIGH]** Schema change from DateTime to DateTime(timezone=True) requires database migration: Both `upload_date` and `last_modified` columns were changed from `Column(DateTime, ...)` to `Column(DateTime(timezone=True), ...)`. This is a schema-level change.

**Previous behavior:** Datetimes stored as timezone-naive values in the database.

**New behavior:** SQLAlchemy now expects/returns timezone-aware datetimes. In SQLite, `DateTime(timezone=True)` stores values as ISO strings with offset info. Existing rows with naive datetime strings may be read back as naive datetimes, causing `TypeError` or incorrect comparisons when mixed with the new timezone-aware values set by `datetime.now(UTC)`.

**Who is affected:** Any existing deployment with data already stored. Any code that compares or processes `upload_date` or `last_modified` values.

**Migration path:** A migration script (e.g., Alembic revision) must be provided to backfill existing rows, normalising stored datetimes to UTC ISO format. Without it, `project.last_modified > original_ts` comparisons may throw `TypeError: can't compare offset-naive and offset-aware datetimes` on real data.

No migration script is included in this PR.
  > **Suggestion:** Add an Alembic migration that normalises all existing datetime strings to UTC. Additionally, pin a minimum supported database version or document the manual upgrade path:
```python
# alembic/versions/xxxx_make_datetimes_timezone_aware.py
def upgrade():
    # SQLite: re-store all datetime strings with +00:00 suffix
    op.execute("""
        UPDATE project
        SET upload_date = upload_date || '+00:00'
        WHERE upload_date IS NOT NULL
          AND upload_date NOT LIKE '%+%'
    """)
    # repeat for last_modified
```
- **[HIGH]** log_transformation and create_checkpoint now raise ValueError for missing project — exception contract changed: **Previous behavior:** Calling `log_transformation` or `create_checkpoint` with a non-existent `project_id` would either succeed silently (no FK constraint enforced in SQLite by default) or raise a database-level `IntegrityError` at commit time.

**New behavior:** `_touch_project` now raises `ValueError` *before* `db.commit()` is reached. The exception type has changed from `IntegrityError` (or no error) to `ValueError`.

**Who is affected:** All API endpoint handlers, background tasks, or integration tests that call these two service functions. Any caller that catches `sqlalchemy.exc.IntegrityError` to handle the missing-project case will silently swallow the new `ValueError`, potentially hiding data integrity issues.

**Migration path:** Callers must add a `except ValueError` handler, or the router layer must map `ValueError` to an appropriate HTTP 404/422 response. Document the new contract in the function docstrings.
  > **Suggestion:** Update each call site's exception handling, and update the docstrings to declare the new exception:
```python
def log_transformation(
    db: Session, project_id: uuid.UUID, operation_type: str, details: dict
) -> None:
    """
    ...
    Raises:
        ValueError: If project_id does not correspond to an existing project.
    """
```
In the router:
```python
try:
    log_transformation(db, project_id, op_type, details)
except ValueError as exc:
    raise HTTPException(status_code=404, detail=str(exc)) from exc
```


### PR Description Issues

- **[MEDIUM]** PR description shows a different implementation than what was actually merged: The PR description's **Changes** section shows this implementation:
```python
project = db.query(models.Project).filter(models.Project.project_id == project_id).first()
if project:
    project.last_modified = datetime.now(timezone.utc)
```
But the actual merged code uses a bulk update via `_touch_project`:
```python
db.query(models.Project).filter(...).update({"last_modified": datetime.now(UTC)})
```
These are meaningfully different: the description version silently ignores a missing project (`if project:`), while the actual code raises `ValueError`. Also `timezone.utc` vs `UTC` differ. The description appears to be from an earlier draft.
  > **Suggestion:** Update the Changes section to reflect the actual implementation:
```python
def _touch_project(db: Session, project_id: uuid.UUID) -> None:
    """Update last_modified to now; raises ValueError if project not found."""
    rows_updated = (
        db.query(models.Project)
        .filter(models.Project.project_id == project_id)
        .update({"last_modified": datetime.now(UTC)})
    )
    if rows_updated == 0:
        raise ValueError(f"Project {project_id} not found")
```
Also note that `log_transformation` and `create_checkpoint` now raise `ValueError` for missing projects.
- **[MEDIUM]** PR description omits significant model changes (timezone-aware columns, onupdate hook): The description focuses entirely on the `project_service.py` changes but does not mention two significant changes in `models.py`:

1. Both `upload_date` and `last_modified` were changed from `DateTime` to `DateTime(timezone=True)` — a schema change that affects existing data.
2. `onupdate=func.now()` was added to `last_modified` — this means *any* ORM-level project update (not just transformations/checkpoints) will now touch `last_modified`.

The Changes table in the description only lists `project_service.py` and `tests/test_save_revert.py`, missing `models.py` entirely.
  > **Suggestion:** Add a models.py row to the Changes table:
```markdown
### `dataloom-backend/app/models.py`

| Change | Detail |
|--------|--------|
| `DateTime` → `DateTime(timezone=True)` | Both `upload_date` and `last_modified` are now timezone-aware. **Requires migration for existing data.** |
| `onupdate=func.now()` on `last_modified` | `last_modified` auto-updates on any ORM-level mutation to a Project row (name, description changes, etc.). |

> ⚠️ A database migration script is required for existing deployments.
```
- **[LOW]** PR description does not document the new ValueError exception contract: The two service functions now raise `ValueError` for missing projects — a new exception contract that callers must handle. This is not mentioned in the PR description, the checklist, or the summary. A developer reading the description and updating their error-handling code would miss this behavioral change.
  > **Suggestion:** Add to the summary or checklist:
```markdown
- [x] `log_transformation` and `create_checkpoint` now raise `ValueError` if the project does not exist (previously: silent or `IntegrityError` at commit time). Callers should add `except ValueError` handling or rely on the API layer to map this to HTTP 404.
```

nodesagar:
@ivantha 

All requested fixes are now included in a single squashed commit (c7bb9a6) on fix/last-modified-timestamp, and PR #134 description has been fully updated to match the final implementation (migration scope, service behavior, tests, and validation).

ivantha:
This PR is behind the base branch. Please rebase onto latest `main` so CI can re-run on the merged state before we can merge.

nodesagar:
@ivantha 

Thanks for the review and follow-ups. I’ve addressed all requested changes:

- Added SQLite downgrade handling for trailing `Z` timestamps and covered it with a regression test (`dataloom-backend/tests/test_migration_timezone_sqlite.py`).
- Updated the `_touch_project` comment in `project_service.py` to avoid implying `func.now()` is guaranteed UTC across DB configurations.
- Ran backend formatting/lint (`ruff format` + `ruff check`) and force-pushed a single-commit branch.

All PR checks are now passing. Please re-review when you have a moment.

ivantha:
Thank you for the fix! Unfortunately this PR includes 248,857 additions that appear to include GSoC PDFs and Python bytecode files which should not be committed. Please clean up the branch (remove the binary/PDF files, add them to .gitignore), and resubmit the fix as a focused PR.

PR #140, opened 2026-02-27
fix(frontend): consistent TopBar height and centered Profile button

### LINKED ISSUE #118: UI: TopBar stylings
**Describe the bug**
- Top Bar is'nt the same width on `/projects` and `/workspace/:id`
- Profile btn not centered.

**Expected behavior**
Same styling across all pages.

**Screenshots**

Image

Image

nodesagar:
# fix(frontend): consistent TopBar height and centered Profile button

**Fixes #118**

---

## Summary

The top navigation bar had two visual bugs. The navbar rendered at different heights depending on the current route — `h-12` (48px) on `/workspace/:id` and `h-16` (64px) on `/projects` — making it appear inconsistent across pages. Additionally, the Profile button was not vertically centered because its container used `items-end` on a plain `div`, which has no effect without `display: flex`.

---

## Root Cause

**File:** `dataloom-frontend/src/Components/Navbar.jsx`

- **Inconsistent height** — `isSmall` toggled between `h-12` and `h-16` based on route, causing a different-sized navbar on each page.
- **No vertical alignment** — `p-4` applied padding on all sides with no `items-center` on the nav, pushing children toward the top instead of centering them.
- **Profile button not centered** — the container div had `items-end` with no `flex`, so the property was silently ignored.

---

### Before

**`/projects` — taller bar (`h-16`), profile button not vertically centered**

<p align="center">
  
  <br/>
  <em>/projects page before alignment fix</em>
</p>

---

**`/workspace/:id` — shorter bar (`h-12`), profile button not vertically centered**

<p align="center">
  
  <br/>
  <em>/workspace page before alignment fix</em>
</p>

---

### After

**`/projects` — consistent `h-14`, profile button centered**

<p align="center">
  
  <br/>
  <em>/projects page after alignment fix</em>
</p>

---

**`/workspace/:id` — consistent `h-14`, profile button centered**

<p align="center">
  
  <br/>
  <em>/workspace page after alignment fix</em>
</p>

---

## What Changed

**File:** `dataloom-frontend/src/Components/Navbar.jsx`

| Problem | Before | After |
|---------|--------|-------|
| Navbar height | `h-12` on workspace, `h-16` on home | Fixed `h-14` on all pages |
| Vertical alignment | `p-4` with no `items-center` — children align to top | `items-center` on nav — all children vertically centered |
| Profile button centering | `items-end` on a non-flex `div` — ignored | `flex items-center` on container — properly centered |
| Padding | `p-4` all sides, conflicts with fixed height | `px-4 md:px-10` horizontal only |
| Profile button sizing | `py-1 px-3 text-sm` on workspace / `py-2 px-4` on home | Unified `py-1.5 px-4 text-sm` everywhere |

---

## Checklist

- [x] Navbar height is consistent (`h-14`) on both `/projects` and `/workspace/:id`
- [x] Profile button is vertically centered on all pages
- [x] Horizontal padding is consistent across routes
- [x] Profile button has uniform sizing across routes
- [x] Project name display on workspace route is preserved
- [x] No breaking changes to layout or other components
- [x] Rebased cleanly onto latest `main`

### DIFF
--- a/dataloom-frontend/src/Components/Navbar.jsx
+++ b/dataloom-frontend/src/Components/Navbar.jsx
@@ -5,40 +5,33 @@ import DataLoomLogo from "./common/DataLoomLogo";
 const Navbar = () => {
   const location = useLocation();
   const { projectName } = useProjectContext();
-  const isSmall = location.pathname.startsWith("/workspace/");
+  const isWorkspacePage = location.pathname.startsWith("/workspace/");
+  const displayProjectName = projectName || "Untitled Project";
 
   return (
     <header role="banner">
       <nav
         aria-label="Main navigation"
-        className={`bg-white border-b border-gray-200 flex p-4 transition-all duration-300 ${
-          isSmall ? "h-12" : "h-16"
-        }`}
+        className="bg-white border-b border-gray-200 flex items-center h-14 px-4 md:px-10"
       >
-        <div
-          className={`text-gray-900 font-semibold ${
-            isSmall ? "text-base" : "text-lg"
-          } flex items-center ml-4 md:ml-10`}
-        >
+        <div className="text-gray-900 font-semibold flex items-center">
           <Link to="/projects" className="flex items-center gap-2">
-            <DataLoomLogo className={isSmall ? "w-5 h-5" : "w-6 h-6"} />
-            DataLoom
+            <DataLoomLogo className="w-5 h-5" />
+            <span className="text-base">DataLoom</span>
           </Link>
         </div>
-        {isSmall && (
-          <div
-            className={`text-gray-700 font-medium ${
-              isSmall ? "text-base" : "text-lg"
-            } flex items-center ml-auto mr-4`}
-          >
-            {projectName || "Untitled Project"}
-          </div>
-        )}
-        <div className="ml-auto text-gray-500 items-end">
+        <div className="ml-auto flex items-center gap-2 min-w-0">
+          {isWorkspacePage && (
+            <div
+              className="text-gray-700 font-medium text-base min-w-0 max-w-[50vw] md:max-w-md mr-2 truncate"
+              title={displayProjectName}
+            >
+              {displayProjectName}
+            </div>
+          )}
           <button
-            className={`bg-white border border-gray-300 rounded-md text-gray-700 hover:bg-gray-50 transition-colors duration-150 ${
-              isSmall ? "py-1 px-3 text-sm" : "py-2 px-4"
-            }`}
+            type="button"
+            className="bg-white border border-gray-300 rounded-md text-gray-700 text-sm py-1.5 px-4 hover:bg-gray-50 transition-colors duration-150"
           >
             Profile
           </button>


dataloom-frontend/src/Components/Navbar.jsx
```diff
@@ -11,35 +11,21 @@ const Navbar = () => {
...
-          <button
-            className={`bg-white border border-gray-300 rounded-md text-gray-700 hover:bg-gray-50 transition-colors duration-150 ${
-              isSmall ? "py-1 px-3 text-sm" : "py-2 px-4"
-            }`}
-          >
+        <div className="ml-auto flex items-center text-gray-500">
+          <button className="bg-white border border-gray-300 rounded-md text-gray-700 text-sm py-1.5 px-4 hover:bg-gray-50 transition-colors duration-150">
             Profile
           </button>
         </div>
```
ivantha: **[MEDIUM]** Double `ml-auto` creates unintended centering of project name when `isSmall` is true

When `isSmall` is `true`, the rendered flex row contains three children:

```
[Logo div]  [project-name div: ml-auto mr-4]  [profile div: ml-auto]
```

In CSS Flexbox, when **multiple children** carry `ml-auto`, the browser distributes all remaining free space **equally** between all auto-margins. This means the project-name div will be roughly **centered** between the logo and the profile button, rather than sitting flush to the right adjacent to the profile.

**Before this PR** (on the workspace route), the project-name already had `ml-auto` and the profile had `ml-auto`, so this same behavior existed — but the screenshots suggest the project name appeared center-adjacent to the profile, implying the old `ml-auto` on the profile container had no visible effect (zero free space remained after the project-name consumed it all). The new code introduces the same layout but the difference is that the Profile container *now correctly renders as flex*, so its `ml-auto` will **now actively consume space**. The project name will visually shift leftward compared to what both the before screenshot and the design intent suggest.

To preserve the original right-aligned project name behavior, `ml-auto` should be removed from the profile container and instead rely on the project-name div's `ml-auto` alone to push everything right:

```jsx
{isSmall && (
  <div className="text-gray-700 font-medium text-base flex items-center ml-auto mr-4">
    {projectName || "Untitled Project"}
  </div>
)}
<div className="flex items-center text-gray-500">   {/* ← no ml-auto here */}
  <button ...>Profile</button>
</div>
```

Alternatively, use `justify-between` on the nav and restructure children into semantic left/center/right groups.

**Suggestion:**
```
Remove `ml-auto` from the profile `div` when the project name is visible. The project-name `div` already has `ml-auto` which pushes it (and everything after it) to the right edge. Only the profile `div` should carry `ml-auto` in the non-`isSmall` case. Consider a structural approach:

```jsx
<nav className="bg-white border-b border-gray-200 flex items-center h-14 px-4 md:px-10">
  {/* Left */}
  <div className="text-gray-900 font-semibold text-base flex items-center">
    <Link to="/projects" className="flex items-center gap-2">
      <DataLoomLogo className="w-5 h-5" />
      DataLoom
    </Link>
  </div>

  {/* Center (workspace only) */}
  {isSmall && (
    <div className="absolute left-1/2 -translate-x-1/2 text-gray-700 font-medium text-base">
      {projectName || "Untitled Project"}
    </div>
  )}

  {/* Right */}
  <div className="ml-auto flex items-center">
    <button type="button" className="...">
      Profile
    </button>
  </div>
</nav>
```

Using absolute centering for the project name ensures it is always visually centered regardless of sibling widths.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR correctly identifies and resolves two genuine visual bugs in the Navbar: the non-functional `items-end` on a non-flex container and the route-dependent height inconsistency. The diff is minimal and well-scoped. However, there is a subtle but consequential layout behavioral change introduced by having two `ml-auto` siblings in the flex row when `isSmall` is true, which may not match the intended design. Several pre-existing issues are also worth surfacing while reviewing the touched file, even though they are out of scope for this specific fix.

**Comment by nodesagar** [comment]
@ivantha this branch is clean, updated with main, and ready for review — closes https://github.com/c2siorg/dataloom/issues/118 on merge.

File: dataloom-frontend/src/Components/Navbar.jsx [hunk 2]
```diff
@@ -5,40 +5,31 @@ import DataLoomLogo from "./common/DataLoomLogo";
...
-              isSmall ? "text-base" : "text-lg"
-            } flex items-center ml-auto mr-4`}
-          >
+        {isWorkspacePage && (
+          <div className="text-gray-700 font-medium text-base flex items-center ml-auto mr-4">
             {projectName || "Untitled Project"}
           </div>
         )}
-        <div className="ml-auto text-gray-500 items-end">
+        <div className="flex items-center gap-2">
```
ivantha: **[HIGH]** Profile button container missing `ml-auto` — layout regression on `/projects`

The profile button wrapper `<div>` no longer has `ml-auto`, which means on the `/projects` route (where `isWorkspacePage` is `false` and the project-name div is not rendered), the profile button sits immediately to the right of the logo with no space between them — both elements are flush to the left edge.

Before, the outer div had `ml-auto`, which pushed it to the far right regardless of route. Now, the only element that carries `ml-auto` is the project-name div, which is conditionally rendered only when `isWorkspacePage` is `true`. On the workspace route this incidentally still looks correct (the project-name div acts as the spacer), but on every other route the profile button is anchored to the left and the right side of the navbar is empty.

This is a visual regression that contradicts the stated goals of the PR.

**Suggestion:**
```
Add `ml-auto` back to the profile button container so it is always right-aligned:
```jsx
<div className="ml-auto flex items-center gap-2">
  {/* TODO: wire up to /profile route or dropdown menu */}
  <button
    type="button"
    disabled
    className="bg-white border border-gray-300 rounded-md text-gray-700 text-sm py-1.5 px-4 hover:bg-gray-50 transition-colors duration-150 opacity-50 cursor-not-allowed"
  >
    Profile
  </button>
</div>
```
With this, on workspace pages the project-name and profile button both naturally stack to the right, and on all other pages the profile button stays pinned right.
```

File: dataloom-frontend/src/Components/Navbar.jsx [hunk 3]
```diff
@@ -5,40 +5,31 @@ import DataLoomLogo from "./common/DataLoomLogo";
...
           <button
-            className={`bg-white border border-gray-300 rounded-md text-gray-700 hover:bg-gray-50 transition-colors duration-150 ${
-              isSmall ? "py-1 px-3 text-sm" : "py-2 px-4"
-            }`}
+            type="button"
+            disabled
+            className="bg-white border border-gray-300 rounded-md text-gray-700 text-sm py-1.5 px-4 hover:bg-gray-50 transition-colors duration-150 opacity-50 cursor-not-allowed"
           >
             Profile
           </button>
```
ivantha, line 37: **[HIGH]** Profile button permanently `disabled` — undisclosed functional regression

The `disabled` attribute has been added to the Profile button along with `opacity-50 cursor-not-allowed`. The button was previously always enabled (even though it had no `onClick` handler). Shipping a visually greyed-out, non-clickable button in a navigation bar with no tooltip, popover, or accessible explanation is a significant UX regression.

The PR description and checklist make no mention of this change; the only trace is a `TODO` comment in the JSX. Users will see a broken/disabled control with no way to understand why or when it will work. Additionally, the existing `hover:bg-gray-50` and `transition-colors` classes are dead code when the button is disabled — hover styles do not apply to disabled buttons in most browsers.

**Suggestion:**
```
**Option A – Remove the button entirely until the profile feature is ready:**
```jsx
{/* Profile button — placeholder, to be implemented in #xxx */}
```

**Option B – Keep the button but add a tooltip explaining it is coming soon:**
```jsx
<button
  type="button"
  disabled
  title="Profile settings coming soon"
  aria-label="Profile settings (coming soon)"
  className="... opacity-50 cursor-not-allowed"
>
  Profile
</button>
```

**Option C – Remove `disabled` and add a no-op `onClick` with a console warning in development, so the visual appearance is preserved.**

Also remove the dead `hover:bg-gray-50 transition-colors` classes when `disabled` is present, since they have no effect.
```

File: dataloom-frontend/src/Components/Navbar.jsx [hunk 4]
```diff
@@ -5,40 +5,31 @@ import DataLoomLogo from "./common/DataLoomLogo";
 const Navbar = () => {
   const location = useLocation();
   const { projectName } = useProjectContext();
-  const isSmall = location.pathname.startsWith("/workspace/");
+  const isWorkspacePage = location.pathname.startsWith("/workspace/");
 
   return (
-    <header role="banner">
+    <header>
```
ivantha: **[MEDIUM]** Removed `role="banner"` from `<header>` — accessibility regression

The `role="banner"` ARIA landmark was explicitly present on the `<header>` element before and has been silently removed. While the HTML5 `<header>` element *implicitly* carries `role="banner"` when it is a **direct child of `<body>`**, this is not guaranteed if the component is ever wrapped in another element (e.g., a layout container, a `<main>`, or a `<div>`). Many React apps render into a single root `<div>`, which means `<header>` is a child of that `<div>`, not of `<body>` — in that case the implicit landmark mapping does **not** apply per the ARIA spec, and screen-reader users lose the ability to navigate directly to the banner landmark.

The explicit `role="banner"` is the safe, spec-compliant choice regardless of DOM depth.

**Suggestion:**
```
Restore the explicit ARIA role:
```jsx
<header role="banner">
```
This is zero-cost and guarantees correct screen-reader landmark navigation in all rendering contexts.
```

File: dataloom-frontend/src/Components/Navbar.jsx [hunk 5]
```diff
@@ -5,40 +5,31 @@ import DataLoomLogo from "./common/DataLoomLogo";
...
-            } flex items-center ml-auto mr-4`}
-          >
+        {isWorkspacePage && (
+          <div className="text-gray-700 font-medium text-base flex items-center ml-auto mr-4">
             {projectName || "Untitled Project"}
           </div>
         )}
-        <div className="ml-auto text-gray-500 items-end">
+        <div className="flex items-center gap-2">
+          {/* TODO: wire up to /profile route or dropdown menu */}
```
ivantha: **[MEDIUM]** TODO comment left in production code without an issue reference

```jsx
{/* TODO: wire up to /profile route or dropdown menu */}
```
TODO comments without a linked issue or ticket are low-signal: they don't encode a deadline, an owner, or acceptance criteria. In a codebase with multiple contributors this kind of comment tends to accumulate and never gets resolved. The fact that the related button is `disabled` makes this particularly important — there is no other signal to users or developers about what the expected behaviour is.

**Suggestion:**
```
Either link the TODO to a tracked issue (`TODO(#xxx): wire up to /profile`) or remove the comment and capture the work in an issue separately. If an issue already exists (e.g., the parent feature ticket), reference it:
```jsx
{/* TODO(#???): wire up to /profile route or dropdown menu */}
```
```

File: dataloom-frontend/src/Components/Navbar.jsx [hunk 6]
```diff
@@ -5,40 +5,31 @@ import DataLoomLogo from "./common/DataLoomLogo";
...
-          className={`text-gray-900 font-semibold ${
-            isSmall ? "text-base" : "text-lg"
-          } flex items-center ml-4 md:ml-10`}
-        >
+        <div className="text-gray-900 font-semibold flex items-center">
           <Link to="/projects" className="flex items-center gap-2">
-            <DataLoomLogo className={isSmall ? "w-5 h-5" : "w-6 h-6"} />
-            DataLoom
+            <DataLoomLogo className={isWorkspacePage ? "w-5 h-5" : "w-6 h-6"} />
+            <span className={isWorkspacePage ? "text-base" : "text-lg"}>DataLoom</span>
```
ivantha: **[LOW]** Logo icon and text size still vary by route — partial consistency fix

The PR's stated goal is to make the navbar consistent across routes. While the height and Profile button sizing are now unified, the logo icon (`w-5 h-5` vs `w-6 h-6`) and the brand name text (`text-base` vs `text-lg`) still change depending on whether the user is on a workspace page. The PR description's change table shows "Profile button sizing" as unified, but does not acknowledge the continued route-dependent logo/text sizing. This creates a half-finished consistency story and leaves `isWorkspacePage` doing double duty — driving both layout (project name) and cosmetic (logo size) concerns.

**Suggestion:**
```
If the intent is full visual consistency, unify the logo and text sizes to a single value:
```jsx
<DataLoomLogo className="w-5 h-5" />
<span className="text-base">DataLoom</span>
```
If the size difference is intentional (e.g., to give workspace pages more room for the project name), document this in the PR description and consider extracting it into a named token:
```js
const logoSize = isWorkspacePage ? "w-5 h-5" : "w-6 h-6";
```
```

File: dataloom-frontend/src/Components/Navbar.jsx [hunk 7]
```diff
@@ -5,40 +5,31 @@ import DataLoomLogo from "./common/DataLoomLogo";
...
-        <div className="ml-auto text-gray-500 items-end">
+        <div className="flex items-center gap-2">
+          {/* TODO: wire up to /profile route or dropdown menu */}
           <button
-            className={`bg-white border border-gray-300 rounded-md text-gray-700 hover:bg-gray-50 transition-colors duration-150 ${
-              isSmall ? "py-1 px-3 text-sm" : "py-2 px-4"
-            }`}
+            type="button"
+            disabled
+            className="bg-white border border-gray-300 rounded-md text-gray-700 text-sm py-1.5 px-4 hover:bg-gray-50 transition-colors duration-150 opacity-50 cursor-not-allowed"
```
ivantha: **[LOW]** Dead Tailwind utility classes on a `disabled` button

`hover:bg-gray-50` and `transition-colors duration-150` are applied to the Profile button, but the button also has the `disabled` HTML attribute. Browsers do not apply `:hover` pseudo-class styles to disabled form elements, so these two utilities have zero visual effect. This is not a runtime error, but it adds noise to the class string and could mislead future developers reviewing the component.

**Suggestion:**
```
Remove the unreachable hover/transition utilities from the disabled button:
```jsx
className="bg-white border border-gray-300 rounded-md text-gray-700 text-sm py-1.5 px-4 opacity-50 cursor-not-allowed"
```
If the button is ever re-enabled, the hover/transition classes can be restored at that time.
```

File: dataloom-frontend/src/Components/Navbar.jsx [hunk 8]
```diff
@@ -5,40 +5,31 @@ import DataLoomLogo from "./common/DataLoomLogo";
...
 
   return (
-    <header role="banner">
+    <header>
       <nav
         aria-label="Main navigation"
-        className={`bg-white border-b border-gray-200 flex p-4 transition-all duration-300 ${
-          isSmall ? "h-12" : "h-16"
-        }`}
+        className="bg-white border-b border-gray-200 flex items-center h-14 px-4 md:px-10"
```
ivantha, line 15: **[INFO]** `transition-all duration-300` removed — minor motion regression

The original `<nav>` had `transition-all duration-300`, which animated any property changes on the navbar. Since the height was previously dynamic, this transition smoothed height changes between routes. The new, fixed height makes the height transition unnecessary, so removing these classes is correct. This is a positive cleanup — just worth noting explicitly in the PR description for completeness.

**Suggestion:**
```
No action needed. This removal is correct and intentional. Optionally document it in the PR change table for reviewers.
```

dataloom-frontend/src/Components/Navbar.jsx
```diff
@@ -5,40 +5,31 @@ import DataLoomLogo from "./common/DataLoomLogo";
 const Navbar = () => {
   const location = useLocation();
   const { projectName } = useProjectContext();
-  const isSmall = location.pathname.startsWith("/workspace/");
+  const isWorkspacePage = location.pathname.startsWith("/workspace/");
```
ivantha, line 8: **[INFO]** Positive: `isSmall` renamed to `isWorkspacePage` — improved readability

`isSmall` was a poorly named boolean — "small" is relative and describes a visual side effect rather than the semantic condition. `isWorkspacePage` accurately describes *what the condition represents*, making every downstream use of the variable self-documenting. This is a good, low-risk improvement.

**Suggestion:**
```
No action needed. This is a positive change worth noting.
```

dataloom-frontend/src/Components/Navbar.jsx
ivantha, line 37: **[INFO]** Positive: `type="button"` added to Profile button

The original button lacked a `type` attribute. In HTML, a `<button>` inside a `<form>` defaults to `type="submit"` and can accidentally submit the form. Explicitly adding `type="button"` is a defensive best practice that prevents unexpected form submissions if this component is ever rendered inside a form context. This is a correct and good addition.

**Suggestion:**
```
No action needed. This is a positive change.
```

ivantha (commented):
## ◎ Argus Review

This PR correctly addresses the stated goals of fixing inconsistent navbar height and vertically centering the Profile button, and includes several good practices (explicit `type="button"`, clearer variable naming, horizontal-only padding). However, it introduces a **layout regression**: the profile button container no longer has `ml-auto`, so on the `/projects` route the button renders flush against the logo on the left instead of the right. A second undisclosed change — permanently disabling the Profile button — is a functional regression not described in the PR. There is also a minor accessibility regression from removing the explicit `role="banner"` ARIA landmark. The overall code quality is good and the PR description is exemplary, but the layout regression and undisclosed disabled-button change should be resolved before merging.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes detected — patch-level change: **Previous behavior:** `Navbar` exported a single default React component with no props interface. No public API, no exported types or constants, no API endpoints, no database schema, and no configuration keys are modified.

**New behavior:** The same component is exported with identical props signature (`none`) and mount interface. Internal rendering logic and class names changed, but consumers import and render `<Navbar />` identically.

**Who is affected:** No downstream consumers are affected by the internal refactor.

**Migration path:** None required.
  > **Suggestion:** This PR is a `patch` semver bump. The changes are bug fixes and internal style refactors with no exported API changes.


### PR Description Issues

- **[MEDIUM]** PR description omits the Profile button `disabled` change: The description's change table and checklist make no mention of the Profile button being changed from enabled to permanently `disabled`. A reviewer or QA engineer reading the description would expect the Profile button to look the same as before (just better centered), not greyed-out and non-interactive. This undisclosed behavioral change could cause confusion during review and in production.
  > **Suggestion:** Add a row to the change table:

| Problem | Before | After |
|---------|--------|-------|
| Profile button state | Enabled (no handler) | Disabled with `opacity-50 cursor-not-allowed` pending #xxx |

Also add a checklist item:
- [x] Profile button is visually marked as unavailable pending feature implementation (#???)
- **[LOW]** Description does not mention removal of `role="banner"` from `<header>`: The accessibility attribute `role="banner"` was silently removed from the `<header>` element. This is not surfaced anywhere in the PR description, screenshots, or checklist. Accessibility changes should always be explicitly called out so reviewers with a11y expertise can evaluate them.
  > **Suggestion:** Add a note in the description, e.g.:
> **Note:** Removed explicit `role="banner"` from `<header>` — the HTML5 element carries this role implicitly. (Alternatively, restored below if this is a concern.)

Or, preferably, restore the attribute and remove the need to mention it.
- **[LOW]** Description does not mention logo/text size still varies by route: The description's stated goal is a *consistent* navbar. The change table entry for "Profile button sizing" correctly notes unification, but there is no corresponding entry acknowledging that logo icon size (`w-5/w-6`) and brand text size (`text-base/text-lg`) still differ by route. A reader would expect full visual consistency and may be surprised to find these remnants.
  > **Suggestion:** Add a row to the change table or a prose note:
> **Note:** Logo icon and brand text sizes (`w-5/w-6`, `text-base/text-lg`) remain route-dependent to preserve available horizontal space on workspace pages. Full unification is tracked in #xxx.
- **[INFO]** Positive: PR description is thorough, well-structured, and visually supported: The description includes a clear root cause analysis, a before/after table, before/after screenshots, and a comprehensive checklist. This is an exemplary level of documentation for a UI bug fix PR and makes the review process significantly easier.
  > **Suggestion:** No action needed. Continue this standard for future PRs.

nodesagar:
@ivantha 

Addressed all review threads on this branch (#140):

- Restored right alignment of the Profile control across routes (`ml-auto`).
- Re-enabled the Profile button and kept `type="button"` as a defensive default.
- Restored explicit `role="banner"` on the header.
- Removed the production TODO comment.
- Unified DataLoom logo/text sizing across routes for full navbar consistency.
- Confirmed fixed-height navbar cleanup (removal of `transition-all duration-300`) is intentional.

Validation is clean:

- `npm run lint` passes
- `npm test -- --run` passes (`48/48`)

ivantha:
This PR is behind the base branch. Please rebase onto latest `main` so CI can re-run on the merged state before we can merge.

nodesagar:
@ivantha 

 I’ve addressed the Navbar overflow issue for long `projectName` values.

Changes made:
- Kept the right side in a single flex container so controls don’t compete for `ml-auto`.
- Constrained the workspace project-name block with `min-w-0` + `max-w` + `truncate`.
- Added a `title` attribute so the full project name is still visible on hover.

This prevents long filenames from pushing the Profile button off-screen on smaller widths.

PR #142, opened 2026-02-27
feat: create initial tables with UUID primary keys for datasets, checkpoints, and user logs

### LINKED ISSUE #141: `alembic upgrade head` fails on fresh database setup, first migration assumes pre-existing tables
**Describe the bug**

Running `uv run alembic upgrade head` on a fresh database (as instructed in CONTRIBUTING.md) fails immediately with a `ProgrammingError`. The first migration (`34e620d988fc`) was written as an ALTER migration — it assumes `datasets`, `checkpoints`, and `user_logs` tables already exist from a prior integer-based schema. On a clean database these tables don't exist, making the documented setup flow broken for any new contributor.

**To Reproduce**

1. Spin up a fresh PostgreSQL instance (e.g. a new Docker container)
2. Clone the repository and `cd dataloom-backend`
3. Copy `.env.example` to `.env` and configure `DATABASE_URL`
4. Run `uv sync`
5. Run `uv run alembic upgrade head`
6. See error:

```
INFO  [alembic.runtime.migration] Running upgrade  -> 34e620d988fc, use_uuids_for_dataset_and_checkpoint_ids
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) relation "checkpoints" does not exist
[SQL: ALTER TABLE checkpoints DROP CONSTRAINT checkpoints_dataset_id_fkey]
```

**Expected behavior**

`alembic upgrade head` completes all three migrations successfully, creating and seeding the database so the backend server can start normally.

**Screenshots**

N/A

**Environment:**
- OS: Windows 11
- Browser: N/A
- Node.js version: N/A
- Python version: 3.12
- PostgreSQL: 16 (Docker)
- `uv` version: latest

**Additional context**

The first migration (`34e620d988fc`) was written to ALTER tables from an older integer-ID schema to UUIDs, but that original "create tables" migration is absent from the repository. As a result there is no valid migration path for a fresh database.

The fix is to rewrite `34e620d988fc`'s `upgrade()` to use `op.create_table()` for all three tables (`datasets`, `checkpoints`, `user_logs`) with UUID keys directly, instead of attempting to ALTER non-existent tables. The subsequent migrations `a1b2c3d4e5f6` and `b2c3d4e5f6a7` require no changes.

Denyme24:
## Description

Fixes fresh database setup failure during backend onboarding.

This PR updates the initial Alembic migration (`34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids`) so it creates the base tables (`datasets`, `checkpoints`, `user_logs`) directly with UUID keys, instead of trying to alter non-existent tables on a clean database.

Without this change, running `uv run alembic upgrade head` on a new PostgreSQL instance fails with `relation "checkpoints" does not exist`, blocking new contributors from completing setup.

Fixes #141

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [ ] New tests added
- [ ] Manual testing

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py
+++ b/dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py
@@ -19,111 +19,50 @@ depends_on: Union[str, Sequence[str], None] = None
 
 
 def upgrade() -> None:
-    """Upgrade schema."""
-    # Step 1: Drop all foreign keys that reference columns being altered
-    op.drop_constraint(op.f('checkpoints_dataset_id_fkey'), 'checkpoints', type_='foreignkey')
-    op.drop_constraint(op.f('user_logs_dataset_id_fkey'), 'user_logs', type_='foreignkey')
-    op.drop_constraint(op.f('user_logs_checkpoint_id_fkey'), 'user_logs', type_='foreignkey')
+    """Create initial tables with UUID primary keys."""
+    # datasets table (will be renamed to projects in a later migration)
+    op.create_table(
+        'datasets',
+        sa.Column('dataset_id', sa.Uuid(), server_default=sa.text('gen_random_uuid()'), nullable=False),
+        sa.Column('name', sa.String(), nullable=False),
+        sa.Column('description', sa.String(), nullable=True),
+        sa.Column('upload_date', sa.DateTime(), server_default=sa.text('now()'), nullable=True),
+        sa.Column('last_modified', sa.DateTime(), server_default=sa.text('now()'), nullable=True),
+        sa.Column('file_path', sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint('dataset_id'),
+    )
+    op.create_index(op.f('ix_datasets_name'), 'datasets', ['name'], unique=False)
 
-    # Step 2: Drop indexes on columns being altered
-    op.drop_index(op.f('ix_checkpoints_id'), table_name='checkpoints')
-    op.drop_index(op.f('ix_datasets_dataset_id'), table_name='datasets')
+    # checkpoints table
+    op.create_table(
+        'checkpoints',
+        sa.Column('id', sa.Uuid(), server_default=sa.text('gen_random_uuid()'), nullable=False),
+        sa.Column('dataset_id', sa.Uuid(), nullable=False),
+        sa.Column('message', sa.String(), nullable=False),
+        sa.Column('created_at', sa.DateTime(), server_default=sa.text('now()'), nullable=True),
+        sa.ForeignKeyConstraint(['dataset_id'], ['datasets.dataset_id'], name='checkpoints_dataset_id_fkey'),
+        sa.PrimaryKeyConstraint('id'),
+    )
 
-    # Step 3: Drop serial defaults before type change (nextval can't cast to UUID)
-    op.alter_column('datasets', 'dataset_id', server_default=None)
-    op.alter_column('checkpoints', 'id', server_default=None)
-
-    # Step 4: Alter all primary key columns first (referenced by FKs)
-    op.alter_column('datasets', 'dataset_id',
-               existing_type=sa.INTEGER(),
-               type_=sa.Uuid(),
-               existing_nullable=False,
-               postgresql_using='gen_random_uuid()')
-    op.alter_column('checkpoints', 'id',
-               existing_type=sa.INTEGER(),
-               type_=sa.Uuid(),
-               existing_nullable=False,
-               postgresql_using='gen_random_uuid()')
-
-    # Step 5: Set new UUID defaults
-    op.alter_column('datasets', 'dataset_id', server_default=sa.text('gen_random_uuid()'))
-    op.alter_column('checkpoints', 'id', server_default=sa.text('gen_random_uuid()'))
-
-    # Step 6: Alter all foreign key columns to match
-    op.alter_column('checkpoints', 'dataset_id',
-               existing_type=sa.INTEGER(),
-               type_=sa.Uuid(),
-               existing_nullable=False,
-               postgresql_using='gen_random_uuid()')
-    op.alter_column('user_logs', 'dataset_id',
-               existing_type=sa.INTEGER(),
-               type_=sa.Uuid(),
-               existing_nullable=False,
-               postgresql_using='gen_random_uuid()')
-    op.alter_column('user_logs', 'checkpoint_id',
-               existing_type=sa.INTEGER(),
-               type_=sa.Uuid(),
-               existing_nullable=True,
-               postgresql_using='checkpoint_id::text::uuid')
-
-    # Step 7: Alter non-FK columns
-    op.alter_column('datasets', 'name',
-               existing_type=sa.VARCHAR(),
-               nullable=False)
-    op.alter_column('datasets', 'file_path',
-               existing_type=sa.VARCHAR(),
-               nullable=False)
-
-    # Step 8: Recreate all foreign keys
-    op.create_foreign_key(None, 'checkpoints', 'datasets', ['dataset_id'], ['dataset_id'])
-    op.create_foreign_key(None, 'user_logs', 'datasets', ['dataset_id'], ['dataset_id'])
-    op.create_foreign_key(None, 'user_logs', 'checkpoints', ['checkpoint_id'], ['id'])
+    # user_logs table
+    op.create_table(
+        'user_logs',
+        sa.Column('change_log_id', sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column('dataset_id', sa.Uuid(), nullable=False),
+        sa.Column('action_type', sa.String(length=50), nullable=False),
+        sa.Column('action_details', sa.JSON(), nullable=False),
+        sa.Column('timestamp', sa.DateTime(), server_default=sa.text('now()'), nullable=False),
+        sa.Column('checkpoint_id', sa.Uuid(), nullable=True),
+        sa.Column('applied', sa.Boolean(), server_default='false', nullable=False),
+        sa.ForeignKeyConstraint(['checkpoint_id'], ['checkpoints.id'], name='user_logs_checkpoint_id_fkey'),
+        sa.ForeignKeyConstraint(['dataset_id'], ['datasets.dataset_id'], name='user_logs_dataset_id_fkey'),
+        sa.PrimaryKeyConstraint('change_log_id'),
+    )
 
 
 def downgrade() -> None:
-    """Downgrade schema."""
-    # Step 1: Drop all foreign keys
-    op.drop_constraint(None, 'user_logs', type_='foreignkey')
-    op.drop_constraint(None, 'user_logs', type_='foreignkey')
-    op.drop_constraint(None, 'checkpoints', type_='foreignkey')
+    op.drop_table('user_logs')
+    op.drop_table('checkpoints')
+    op.drop_index(op.f('ix_datasets_name'), table_name='datasets')
+    op.drop_table('datasets')
 
-    # Step 2: Revert FK columns back to INTEGER
-    op.alter_column('user_logs', 'checkpoint_id',
-               existing_type=sa.Uuid(),
-               type_=sa.INTEGER(),
-               existing_nullable=True)
-    op.alter_column('user_logs', 'dataset_id',
-               existing_type=sa.Uuid(),
-               type_=sa.INTEGER(),
-               existing_nullable=False)
-    op.alter_column('checkpoints', 'dataset_id',
-               existing_type=sa.Uuid(),
-               type_=sa.INTEGER(),
-               existing_nullable=False)
-
-    # Step 3: Revert PK columns back to INTEGER
-    op.alter_column('datasets', 'dataset_id',
-               existing_type=sa.Uuid(),
-               type_=sa.INTEGER(),
-               existing_nullable=False)
-    op.alter_column('checkpoints', 'id',
-               existing_type=sa.Uuid(),
-               type_=sa.INTEGER(),
-               existing_nullable=False)
-
-    # Step 4: Revert non-FK column changes
-    op.alter_column('datasets', 'file_path',
-               existing_type=sa.VARCHAR(),
-               nullable=True)
-    op.alter_column('datasets', 'name',
-               existing_type=sa.VARCHAR(),
-               nullable=True)
-
-    # Step 5: Recreate indexes
-    op.create_index(op.f('ix_datasets_dataset_id'), 'datasets', ['dataset_id'], unique=False)
-    op.create_index(op.f('ix_checkpoints_id'), 'checkpoints', ['id'], unique=False)
-
-    # Step 6: Recreate foreign keys
-    op.create_foreign_key(op.f('checkpoints_dataset_id_fkey'), 'checkpoints', 'datasets', ['dataset_id'], ['dataset_id'], ondelete='CASCADE')
-    op.create_foreign_key(op.f('user_logs_checkpoint_id_fkey'), 'user_logs', 'checkpoints', ['checkpoint_id'], ['id'], ondelete='SET NULL')
-    op.create_foreign_key(op.f('user_logs_dataset_id_fkey'), 'user_logs', 'datasets', ['dataset_id'], ['dataset_id'], ondelete='CASCADE')


dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py
```diff
@@ -20,114 +20,50 @@
...
+        sa.Column('change_log_id', sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column('dataset_id', sa.Uuid(), nullable=False),
+        sa.Column('action_type', sa.String(length=50), nullable=False),
+        sa.Column('action_details', sa.JSON(), nullable=False),
+        sa.Column('timestamp', sa.DateTime(), server_default=sa.text('now()'), nullable=False),
+        sa.Column('checkpoint_id', sa.Uuid(), nullable=True),
+        sa.Column('applied', sa.Boolean(), server_default='false', nullable=False),
+        sa.ForeignKeyConstraint(['checkpoint_id'], ['checkpoints.id'], name='user_logs_checkpoint_id_fkey'),
+        sa.ForeignKeyConstraint(['dataset_id'], ['datasets.dataset_id'], name='user_logs_dataset_id_fkey'),
+        sa.PrimaryKeyConstraint('change_log_id'),
```
ivantha: **[HIGH]** Missing ondelete CASCADE / SET NULL behavior on all foreign keys

The original `downgrade()` function (before this PR) explicitly recreated the three foreign keys with referential-action rules:

- `checkpoints_dataset_id_fkey` → `ondelete="CASCADE"`
- `user_logs_dataset_id_fkey` → `ondelete="CASCADE"`
- `user_logs_checkpoint_id_fkey` → `ondelete="SET NULL"`

The new `create_table` calls define all three FKs **without** any `ondelete` clause, which defaults to `RESTRICT` in PostgreSQL. This is a silent behavioral regression:

1. Deleting a `datasets` row now **raises a FK violation error** instead of cascading to its checkpoints and user_logs.
2. Deleting a `checkpoints` row now **raises a FK violation error** instead of nullifying `user_logs.checkpoint_id`.

This may break application logic that relied on the database to handle cascades automatically, and it may cause orphaned-row bugs if the application code was never written to manually delete dependent rows first.

**Suggestion:**
```
Add `ondelete` to each `ForeignKeyConstraint` to match the original intent:

```python
# checkpoints table
sa.ForeignKeyConstraint(
    ['dataset_id'], ['datasets.dataset_id'],
    name='checkpoints_dataset_id_fkey',
    ondelete='CASCADE'
),

# user_logs table
sa.ForeignKeyConstraint(
    ['checkpoint_id'], ['checkpoints.id'],
    name='user_logs_checkpoint_id_fkey',
    ondelete='SET NULL'
),
sa.ForeignKeyConstraint(
    ['dataset_id'], ['datasets.dataset_id'],
    name='user_logs_dataset_id_fkey',
    ondelete='CASCADE'
),
```
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py [hunk 2]
```diff
@@ -20,114 +20,50 @@
...
 def downgrade() -> None:
-    """Downgrade schema."""
-    # Step 1: Drop all foreign keys
-    op.drop_constraint(None, "user_logs", type_="foreignkey")
-    op.drop_constraint(None, "user_logs", type_="foreignkey")
-    op.drop_constraint(None, "checkpoints", type_="foreignkey")
+    op.drop_table('user_logs')
+    op.drop_table('checkpoints')
+    op.drop_index(op.f('ix_datasets_name'), table_name='datasets')
+    op.drop_table('datasets')
```
ivantha: **[HIGH]** Rewriting an existing migration in-place rather than creating a new squash migration

Alembic migrations are append-only by convention. Mutating the body of `34e620d988fc` in-place is problematic for any developer or CI environment that has **already applied** the old version of this migration:

- Their `alembic_version` table records the revision as applied, so `alembic upgrade head` will **skip** this file entirely — they will never pick up the fixed schema.
- Running `alembic downgrade` on a database that was previously migrated with the old (ALTER-based) script will now call `op.drop_table(...)` on tables that were never created by this migration, which will drop real data.

The safe Alembic pattern for fixing a broken initial migration depends on the project's stage:
- **Pre-release / no deployed instances**: squash or rewrite is acceptable, but it should be documented clearly.
- **Any deployed instance exists**: create a new migration (`alembic revision`) that corrects the schema differences, and leave the old migration intact.

The PR description implies this only affects fresh setups and fixes #141, but it provides no guidance to existing users about what to do, and there is no guard (e.g., `op.get_bind().dialect.has_table(...)`) to make the script idempotent.

**Suggestion:**
```
If this truly only affects brand-new databases (i.e., no one has successfully run the old migration in any environment), add a prominent comment at the top of the file and in the PR description explaining that existing databases should treat this revision as already applied. If there is any doubt, create a new corrective migration instead:

```bash
alembic revision --autogenerate -m "fix_initial_tables_create_if_not_exist"
```

And use `op.get_bind()` checks or conditional logic inside the migration body to make it safe to run on both fresh and existing databases.
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py (hunk 1 above)
Comment (ivantha) [inline dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py]: **[MEDIUM]** No indexes created on foreign key columns

Three foreign key columns are declared across two tables, but none of them have a corresponding index:

- `checkpoints.dataset_id` — lookups like "give me all checkpoints for dataset X" will do a full sequential scan on `checkpoints`.
- `user_logs.dataset_id` — same problem; full scan for any per-dataset audit query.
- `user_logs.checkpoint_id` — same problem for checkpoint-based queries.

PostgreSQL does **not** automatically create indexes on FK columns (unlike some other databases). Without these indexes, JOIN and WHERE-clause queries on these columns will degrade as data grows. The original migration did create `ix_checkpoints_id` (on the PK) and `ix_datasets_dataset_id` (redundant with the PK), but it never addressed FK column indexes either — so this is also a pre-existing gap being perpetuated.

**Suggestion:**
```
Add index creation calls after each `create_table`:

```python
# After creating checkpoints
op.create_index('ix_checkpoints_dataset_id', 'checkpoints', ['dataset_id'], unique=False)

# After creating user_logs
op.create_index('ix_user_logs_dataset_id', 'user_logs', ['dataset_id'], unique=False)
op.create_index('ix_user_logs_checkpoint_id', 'user_logs', ['checkpoint_id'], unique=False)
```

And remember to drop them in `downgrade()` before dropping the tables.
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py [hunk 3]
```diff
@@ -20,114 +20,50 @@
...
 
 def downgrade() -> None:
-    """Downgrade schema."""
-    # Step 1: Drop all foreign keys
-    op.drop_constraint(None, "user_logs", type_="foreignkey")
-    op.drop_constraint(None, "user_logs", type_="foreignkey")
-    op.drop_constraint(None, "checkpoints", type_="foreignkey")
+    op.drop_table('user_logs')
+    op.drop_table('checkpoints')
+    op.drop_index(op.f('ix_datasets_name'), table_name='datasets')
```
ivantha: **[MEDIUM]** downgrade() drops index before dropping the table, but uses wrong drop order for cascading FKs

The `downgrade()` function drops `user_logs` first, then `checkpoints`, then the `datasets` index, then `datasets`. The order of table drops is correct given the FK dependencies. However, since `ix_datasets_name` is an index **on** the `datasets` table, dropping it before dropping the table is unnecessary (dropping the table drops all its indexes automatically) and adds risk of error if the index name has drifted. More critically, if future migrations add additional indexes to `checkpoints` or `user_logs`, this `downgrade()` will fail to drop them, leaving orphaned index entries.

Also, if the `checkpoints` → `datasets` FK has `ondelete='CASCADE'`, dropping `user_logs` and `checkpoints` before `datasets` is still needed for `RESTRICT` semantics, but with `CASCADE` it would be automatic. The current downgrade is fragile if ondelete semantics change.

**Suggestion:**
```
Drop indexes explicitly before dropping their parent tables for clarity, and document why the order matters:

```python
def downgrade() -> None:
    # Drop in reverse dependency order
    op.drop_table('user_logs')
    op.drop_table('checkpoints')
    op.drop_index(op.f('ix_datasets_name'), table_name='datasets')
    op.drop_table('datasets')
```

The current order is actually correct — just add a comment explaining the dependency chain and consider whether explicit index drops for all created indexes are needed.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR correctly addresses a real bug: the initial Alembic migration was written to `ALTER` tables that don't yet exist on a fresh database, making onboarding impossible. Switching to `CREATE TABLE` statements is the right architectural fix. However, the rewrite introduces several regressions — most notably, the removal of `ondelete` referential-integrity rules from all three foreign keys, which changes production database behavior and can cause orphaned rows. There are also missing performance indexes on FK columns and a few minor style inconsistencies.

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py [hunk 4]
```diff
@@ -20,114 +20,50 @@
...
-        existing_nullable=False,
-        postgresql_using="gen_random_uuid()",
+    # user_logs table
+    op.create_table(
+        'user_logs',
+        sa.Column('change_log_id', sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column('dataset_id', sa.Uuid(), nullable=False),
+        sa.Column('action_type', sa.String(length=50), nullable=False),
+        sa.Column('action_details', sa.JSON(), nullable=False),
+        sa.Column('timestamp', sa.DateTime(), server_default=sa.text('now()'), nullable=False),
```
ivantha, line 55: **[HIGH]** Missing ON DELETE behavior on foreign keys — silent referential integrity regression

The old migration's `downgrade()` recreated the original FKs with `ondelete='CASCADE'` for `checkpoints_dataset_id_fkey` and `user_logs_dataset_id_fkey`, and `ondelete='SET NULL'` for `user_logs_checkpoint_id_fkey`. These document the **intended referential integrity semantics**. The new `CREATE TABLE` statements omit `ondelete` entirely, so all three FKs default to `RESTRICT` in PostgreSQL, meaning:
- Deleting a `datasets` row will fail if any `checkpoints` or `user_logs` reference it (should CASCADE).
- Deleting a `checkpoints` row will fail if any `user_logs` reference it (should SET NULL on `checkpoint_id`).

This causes unexpected errors at the application level on any delete operation.

**Suggestion:**
```
Add `ondelete` parameters to match the original intent:
```python
sa.ForeignKeyConstraint(['dataset_id'], ['datasets.dataset_id'],
    name='checkpoints_dataset_id_fkey', ondelete='CASCADE'),
# in user_logs:
sa.ForeignKeyConstraint(['dataset_id'], ['datasets.dataset_id'],
    name='user_logs_dataset_id_fkey', ondelete='CASCADE'),
sa.ForeignKeyConstraint(['checkpoint_id'], ['checkpoints.id'],
    name='user_logs_checkpoint_id_fkey', ondelete='SET NULL'),
```
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py [hunk 5]
```diff
@@ -20,114 +20,50 @@
...
+    op.create_index(op.f('ix_datasets_name'), 'datasets', ['name'], unique=False)
+
+    # checkpoints table
+    op.create_table(
+        'checkpoints',
+        sa.Column('id', sa.Uuid(), server_default=sa.text('gen_random_uuid()'), nullable=False),
+        sa.Column('dataset_id', sa.Uuid(), nullable=False),
+        sa.Column('message', sa.String(), nullable=False),
+        sa.Column('created_at', sa.DateTime(), server_default=sa.text('now()'), nullable=True),
+        sa.ForeignKeyConstraint(['dataset_id'], ['datasets.dataset_id'], name='checkpoints_dataset_id_fkey'),
```
ivantha, line 44: **[MEDIUM]** Missing indexes on FK columns — unindexed foreign keys in checkpoints and user_logs

The `checkpoints` table defines a FK on `dataset_id` but no explicit index is created for it. PostgreSQL does **not** automatically index FK columns (unlike PKs). Without this, any join from `checkpoints` to `datasets` requires a full sequential scan as data grows. Similarly, `user_logs` has FK columns `dataset_id` and `checkpoint_id` with no supporting indexes, making all FK-driven lookups and joins on `user_logs` potentially expensive.

**Suggestion:**
```
Add index creation after each `create_table` call:
```python
op.create_index('ix_checkpoints_dataset_id', 'checkpoints', ['dataset_id'], unique=False)
op.create_index('ix_user_logs_dataset_id', 'user_logs', ['dataset_id'], unique=False)
op.create_index('ix_user_logs_checkpoint_id', 'user_logs', ['checkpoint_id'], unique=False)
```
And drop them in `downgrade()` before dropping the tables.
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py (hunk 4 above)
Comment (ivantha) [inline dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py:55]: **[MEDIUM]** server_default='false' is a raw Python string, not a SQL expression

The `applied` column uses `server_default='false'` (a bare Python string). While PostgreSQL accepts `'false'` for boolean columns, this is inconsistent with every other `server_default` in this file, which consistently wraps SQL in `sa.text()`. Using a raw string bypasses SQLAlchemy's type-aware quoting and can silently break on non-PostgreSQL databases (e.g., MySQL does not accept bare `'false'` for BOOL).

**Suggestion:**
```
Use `sa.text()` for consistency:
```python
sa.Column('applied', sa.Boolean(), server_default=sa.text('false'), nullable=False),
```
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py [hunk 6]
```diff
@@ -20,114 +20,50 @@
...
-    op.create_foreign_key(None, "user_logs", "checkpoints", ["checkpoint_id"], ["id"])
 
 
 def downgrade() -> None:
-    """Downgrade schema."""
-    # Step 1: Drop all foreign keys
-    op.drop_constraint(None, "user_logs", type_="foreignkey")
-    op.drop_constraint(None, "user_logs", type_="foreignkey")
-    op.drop_constraint(None, "checkpoints", type_="foreignkey")
+    op.drop_table('user_logs')
```
ivantha, line 65: **[MEDIUM]** Downgrade is now fully destructive with no warning or guard

The `downgrade()` function now uses `op.drop_table()` for all three tables. While logically correct as the inverse of `CREATE TABLE`, this means running `alembic downgrade` will **permanently destroy all data** in `datasets`, `checkpoints`, and `user_logs` with no prompt or guard. The original downgrade preserved data by reverting column types. This is especially risky if `alembic downgrade` is ever run in a production rollback scenario.

**Suggestion:**
```
Add a prominent warning docstring:
```python
def downgrade() -> None:
    """
    WARNING: DESTRUCTIVE OPERATION.
    Drops user_logs, checkpoints, and datasets tables, deleting all data.
    Ensure you have a verified backup before running on any non-fresh database.
    """
    op.drop_table('user_logs')
    op.drop_table('checkpoints')
    op.drop_index(op.f('ix_datasets_name'), table_name='datasets')
    op.drop_table('datasets')
```
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py [hunk 7]
```diff
@@ -20,114 +20,50 @@
...
+    """Create initial tables with UUID primary keys."""
+    # datasets table (will be renamed to projects in a later migration)
+    op.create_table(
+        'datasets',
+        sa.Column('dataset_id', sa.Uuid(), server_default=sa.text('gen_random_uuid()'), nullable=False),
+        sa.Column('name', sa.String(), nullable=False),
+        sa.Column('description', sa.String(), nullable=True),
+        sa.Column('upload_date', sa.DateTime(), server_default=sa.text('now()'), nullable=True),
+        sa.Column('last_modified', sa.DateTime(), server_default=sa.text('now()'), nullable=True),
+        sa.Column('file_path', sa.String(), nullable=False),
```
ivantha, line 32: **[LOW]** New ix_datasets_name index introduced silently without rationale

The original migration dropped `ix_datasets_dataset_id`. The new migration never creates it (correct — the PK constraint already provides an index on `dataset_id`) but silently introduces a new `ix_datasets_name` index without any comment explaining why name lookups are expected to be frequent. Future contributors won't understand the reasoning. It is also worth confirming whether `unique=True` should be used to prevent duplicate dataset names.

**Suggestion:**
```
Add an inline comment explaining the index purpose:
```python
# Index for efficient name-based lookups in list/search endpoints
op.create_index(op.f('ix_datasets_name'), 'datasets', ['name'], unique=False)
```
Also evaluate whether `unique=True` is appropriate for the `name` column.
```

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py (hunk 6 above)
Comment (ivantha) [inline dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py:65]: **[LOW]** downgrade() missing docstring while upgrade() has one

The `upgrade()` function has a docstring (`"""Create initial tables with UUID primary keys."""`), but `downgrade()` has none. Given that this downgrade is now destructive (drops tables with all data), a docstring is especially important for anyone who may invoke it.

**Suggestion:**
```
```python
def downgrade() -> None:
    """Drop all tables created by this migration. WARNING: all row data will be lost."""
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR correctly identifies and fixes a real onboarding bug — the original migration assumed tables already existed and would fail on a fresh PostgreSQL instance. The rewrite to `CREATE TABLE` is clean, readable, and uses explicit FK constraint names (a genuine improvement over the original). However, there are two significant issues requiring attention before merge: (1) reusing the same revision ID `34e620d988fc` will silently leave any existing database schema out of sync, as Alembic will skip the new logic for anyone who already ran the old migration — a new revision ID or a clear mandatory-downgrade notice is needed; (2) the `ondelete=CASCADE` and `ondelete=SET NULL` behaviors documented in the original migration are absent from the new FK definitions, defaulting to `RESTRICT` and causing unexpected errors on deletes. Additionally, the PR description leaves the testing section unfilled and the breaking-change checkbox unchecked, both of which should be corrected before review approval.

### ⚠️ Breaking Changes

- **[HIGH]** Same revision ID reused with incompatible schema — existing users will have silent schema mismatch: The migration retains the revision ID `34e620d988fc` but completely rewrites `upgrade()` from `ALTER TABLE` operations to `CREATE TABLE` operations. Any developer or CI environment that has **already run the old version** of this migration will have it recorded as applied in `alembic_version`. Running `alembic upgrade head` will skip it entirely, leaving those databases with the old schema (integer PKs, altered columns) while the codebase now expects UUID PKs created from scratch.

**Previous behavior:** Migration altered existing integer-PK tables to UUID.
**New behavior:** Migration creates fresh tables with UUID PKs from scratch.
**Who is affected:** Any contributor or deployment environment that ran the old migration. Their `alembic_version` table already contains `34e620d988fc`, so this new CREATE TABLE logic will never execute for them.
**Migration path:** Affected users must manually run `alembic downgrade <previous_rev>` and then re-upgrade, or drop and recreate the affected tables. This must be documented prominently.


### PR Description Issues

- **[HIGH]** Testing section left as template placeholder — no description of actual tests run: The "How Has This Been Tested?" section contains the unedited prompt text: *"Describe the tests you ran to verify your changes."* — it was never filled in. Only `[x] Existing tests pass` is checked, with no explanation of which tests, in what environment, or whether the core bug scenario (running `alembic upgrade head` on a blank PostgreSQL instance) was actually verified. For a migration fix, this is the most critical test case to document.
  > **Suggestion:** Replace the placeholder with a concrete test description, for example:
> **Verified on a fresh PostgreSQL 15 instance:**
> 1. Created a blank database: `createdb dataloom_test`
> 2. Updated `.env` with the new DB URL
> 3. Ran `uv run alembic upgrade head` — completed without errors; all 3 tables created
> 4. Confirmed schema: `\d datasets`, `\d checkpoints`, `\d user_logs` in `psql`
> 5. Ran full test suite: `uv run pytest` — all N tests passed
- **[MEDIUM]** Breaking change checkbox left unchecked despite behavioral regressions for existing users: The PR marks `[ ] Breaking change` as unchecked, but this migration **is** a breaking change for developers who have already run the old version of `34e620d988fc`. Reusing the same revision ID with an entirely different schema operation will cause their Alembic state to diverge from reality. Additionally, the missing `ondelete=CASCADE/SET NULL` on FKs changes delete behavior for all users. Both issues warrant explicit acknowledgment in the PR.
  > **Suggestion:** Check `[x] Breaking change` and add a migration notice:
> **⚠️ Breaking for existing databases:** If you ran the previous version of this migration, you must downgrade and re-apply:
> ```bash
> uv run alembic downgrade base
> uv run alembic upgrade head
> ```
> **Warning:** This drops all data in `datasets`, `checkpoints`, and `user_logs`. Back up your data first.
- **[LOW]** Entire self-review checklist is unchecked — template not completed: All items in the Checklist section (`My code follows the project's style guidelines`, `I have performed a self-review`, `I have added/updated documentation as needed`, `My changes generate no new warnings`, `Tests pass locally`) are unchecked. For a database migration that changes referential integrity behavior, a genuine self-review is particularly important before merge.
  > **Suggestion:** Complete the checklist honestly. At minimum, verify style guidelines compliance and confirm whether any onboarding documentation (e.g., a README or `CONTRIBUTING.md` that references `alembic upgrade head`) needs to be updated to mention the downgrade requirement for existing users.

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py (hunk 6 above)
Comment (ivantha) [inline dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py:65]: **[INFO]** Positive: explicit FK constraint naming is a clear improvement over the original

The new `CREATE TABLE` approach names all foreign key constraints explicitly (e.g., `checkpoints_dataset_id_fkey`, `user_logs_dataset_id_fkey`, `user_logs_checkpoint_id_fkey`). This is a meaningful improvement over the old `upgrade()`, which passed `None` as constraint names when recreating FKs — making future `drop_constraint` calls error-prone and fragile. The new approach makes migrations more predictable and easier to maintain.

File: dataloom-backend/alembic/versions/34e620d988fc_use_uuids_for_dataset_and_checkpoint_ids.py [hunk 8]
```diff
@@ -20,114 +20,50 @@
...
-
-    # Step 4: Alter all primary key columns first (referenced by FKs)
-    op.alter_column(
-        "datasets",
-        "dataset_id",
-        existing_type=sa.INTEGER(),
-        type_=sa.Uuid(),
-        existing_nullable=False,
-        postgresql_using="gen_random_uuid()",
+    """Create initial tables with UUID primary keys."""
```
ivantha, line 23: **[INFO]** Positive: forward-looking comment about future table rename aids contributor onboarding

The inline comment `# datasets table (will be renamed to projects in a later migration)` is an excellent example of developer communication. It signals the intent of future migrations without cluttering the code, helping reviewers and contributors understand the broader design trajectory and reducing surprises during onboarding.

hanzalahwaheed:
https://github.com/c2siorg/dataloom/pull/213

i've created a PR that depends on this PR as migration needs to be fixed so that docker compose runs properly later on.

ivantha:
Holding this PR — the latest CI run shows Backend Lint failing. Please fix the lint errors (`cd dataloom-backend && uv run ruff check . && uv run ruff format --check .`) and push so CI goes green before we can merge.

ivantha:
Thank you for the Alembic migration work! This has been superseded by #233 which provides a cleaner fresh-DB migration setup. Feel free to follow up on that PR.

PR #143, opened 2026-02-27
feat: implement CSV export functionality with customizable options

### LINKED ISSUE #92: CSV Export Format Options
**Difficulty:** Medium
**Area:** full stack

## Description

Currently, DataLoom does not provide users with control over how their data is exported. When downloading the working data, users get a standard comma-separated CSV file with headers and UTF-8 encoding. However, different downstream tools and systems have different format requirements. Some tools expect tab-separated values, some legacy systems need semicolons as delimiters, certain European tools expect Latin-1 encoding, and some import processes require files without header rows.

This issue adds an export dialog that allows users to customize the output format before downloading. The backend will receive format parameters and generate the file accordingly, streaming it back as a downloadable response.

## Acceptance Criteria

- [ ] An export endpoint exists and returns a downloadable CSV file
- [ ] Delimiter parameter works correctly for comma, tab, semicolon, and pipe
- [ ] Header parameter correctly includes or excludes the header row
- [ ] Encoding parameter correctly encodes the file in UTF-8, Latin-1, ASCII, or UTF-16
- [ ] Invalid delimiter or encoding values are rejected with a 400 error
- [ ] Response has correct Content-Disposition header with filename
- [ ] Frontend export dialog shows delimiter, header, and encoding options
- [ ] Clicking "Export" downloads the file with the chosen options
- [ ] Export option is accessible from the menu/toolbar
- [ ] Downloaded file opens correctly in external tools (Excel, text editors)

## Resources

- `dataloom-backend/app/utils/pandas_helpers.py` — CSV reading utilities (reference for file handling)
- `dataloom-backend/app/services/file_service.py` — file management patterns
- `dataloom-frontend/src/Components/MenuNavbar.jsx` — menu for adding export option
- `dataloom-frontend/src/Components/common/Modal.jsx` — modal component for dialog
- `dataloom-frontend/src/api/client.js` — Axios instance
- [pandas `to_csv` documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html)
- [FastAPI StreamingResponse](https://fastapi.tiangolo.com/advanced/custom-response/#streamingresponse)

Denyme24:
## Description

Implements configurable CSV export options, allowing users to choose delimiter, encoding, and header inclusion before downloading their data.

- Added `ExportDelimiter` and `ExportEncoding` Pydantic enums to schemas.py
- Updated `GET /projects/{project_id}/export` to accept `delimiter`, `include_header`, and `encoding` query params, validated via enums, returning a `StreamingResponse` with correct `Content-Disposition` and charset headers
- Invalid delimiter or encoding values are rejected with a `422` error automatically via FastAPI enum validation
- Updated `exportProject()` in projects.js to forward format options as query params (defaults preserved, no breaking change for existing callers)
- Added `ExportDialog` component with delimiter, encoding, and header options
- Wired `ExportDialog` into `MenuNavbar`, the existing Export button now opens the dialog instead of triggering an immediate download

Fixes #92

here are some screenshots : 
image
image
image

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [ ] New tests added
- [ ] Manual testing

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -3,10 +3,11 @@
 Handles upload, retrieval, save (checkpoint), and revert operations.
 """
 
+import io
 import uuid
 
-from fastapi import APIRouter, Depends, File, Form, HTTPException, UploadFile
-from fastapi.responses import FileResponse
+from fastapi import APIRouter, Depends, File, Form, HTTPException, Query, UploadFile
+from fastapi.responses import StreamingResponse
 from sqlmodel import Session
 
 from app import database, models, schemas
@@ -185,10 +186,31 @@ async def revert_to_checkpoint(
 
 
 @router.get("/{project_id}/export")
-async def export_project(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
-    """Download the current working copy of a project as a CSV file."""
+async def export_project(
+    project_id: uuid.UUID,
+    delimiter: schemas.ExportDelimiter = Query(default=schemas.ExportDelimiter.comma),
+    include_header: bool = Query(default=True),
+    encoding: schemas.ExportEncoding = Query(default=schemas.ExportEncoding.utf8),
+    db: Session = Depends(database.get_db),
+):
+    """Download the current working copy of a project as a CSV file.
+
+    Supports custom delimiter, header inclusion, and file encoding.
+    """
     project = get_project_or_404(project_id, db)
-    return FileResponse(project.file_path, media_type="text/csv", filename=f"{project.name}.csv")
+    df = read_csv_safe(project.file_path)
+
+    buffer = io.StringIO()
+    df.to_csv(buffer, index=False, sep=delimiter.to_char(), header=include_header)
+    csv_bytes = buffer.getvalue().encode(encoding.value)
+
+    filename = f"{project.name}.csv"
+    logger.info("Export: id=%s, delimiter=%s, header=%s, encoding=%s", project_id, delimiter, include_header, encoding)
+    return StreamingResponse(
+        io.BytesIO(csv_bytes),
+        media_type=f"text/csv; charset={encoding.value}",
+        headers={"Content-Disposition": f'attachment; filename="{filename}"'},
+    )
 
 
 @router.delete("/{project_id}")
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -5,9 +5,10 @@ Handles upload, retrieval, save (checkpoint), and revert operations.
 
 import io
 import uuid
+from urllib.parse import quote
 
 from fastapi import APIRouter, Depends, File, Form, HTTPException, Query, UploadFile
-from fastapi.responses import StreamingResponse
+from fastapi.responses import FileResponse, Response, StreamingResponse
 from sqlmodel import Session
 
 from app import database, models, schemas
@@ -21,7 +22,7 @@ from app.services.project_service import (
 from app.services.file_service import delete_project_files, get_original_path, store_upload
 from app.services.transformation_service import apply_logged_transformation
 from app.utils.logging import get_logger
-from app.utils.pandas_helpers import dataframe_to_response, read_csv_safe, save_csv_safe
+from app.utils.pandas_helpers import dataframe_to_response, read_csv_as_strings, read_csv_safe, save_csv_safe
 from app.utils.security import validate_upload_file
 
 logger = get_logger(__name__)
@@ -198,18 +199,63 @@ async def export_project(
     Supports custom delimiter, header inclusion, and file encoding.
     """
     project = get_project_or_404(project_id, db)
-    df = read_csv_safe(project.file_path)
+
+    # Build a safe Content-Disposition header per RFC 6266.
+    # Strip characters that would break the quoted-string token, then provide
+    # both a legacy ASCII fallback and a percent-encoded filename* parameter
+    # for non-ASCII names and modern browsers.
+    safe_name = project.name.replace('"', "").replace("\\", "")
+    encoded_name = quote(f"{safe_name}.csv", safe="")
+    content_disposition = (
+        f'attachment; filename="{safe_name}.csv"; '
+        f"filename*=UTF-8''{encoded_name}"
+    )
+    headers = {"Content-Disposition": content_disposition}
+
+    logger.info(
+        "Export: id=%s, delimiter=%s, header=%s, encoding=%s",
+        project_id, delimiter, include_header, encoding,
+    )
+
+    # Fast path: serve the raw file bytes for the default combination so that
+    # no pandas round-trip can silently alter data values (leading zeros, date
+    # reformatting, scientific notation, etc.).
+    if (
+        delimiter == schemas.ExportDelimiter.comma
+        and include_header
+        and encoding == schemas.ExportEncoding.utf8
+    ):
+        return FileResponse(
+            project.file_path,
+            media_type="text/csv; charset=utf-8",
+            headers=headers,
+        )
+
+    # Non-default: read every cell as a string to preserve original
+    # representations, then re-serialise with the requested options.
+    df = read_csv_as_strings(project.file_path)
 
     buffer = io.StringIO()
     df.to_csv(buffer, index=False, sep=delimiter.to_char(), header=include_header)
-    csv_bytes = buffer.getvalue().encode(encoding.value)
 
-    filename = f"{project.name}.csv"
-    logger.info("Export: id=%s, delimiter=%s, header=%s, encoding=%s", project_id, delimiter, include_header, encoding)
-    return StreamingResponse(
-        io.BytesIO(csv_bytes),
+    try:
+        csv_bytes = buffer.getvalue().encode(encoding.value)
+    except UnicodeEncodeError:
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                f"The data contains characters that cannot be encoded as "
+                f"{encoding.value}. Try UTF-8 instead."
+            ),
+        )
+
+    # The full content is already in memory (required to catch the encoding
+    # error above), so use Response directly rather than giving a false
+    # impression of streaming via StreamingResponse.
+    return Response(
+        content=csv_bytes,
         media_type=f"text/csv; charset={encoding.value}",
-        headers={"Content-Disposition": f'attachment; filename="{filename}"'},
+        headers=headers,
     )
 
 
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -126,6 +126,26 @@ class DataType(str, Enum):
     datetime = "datetime"
 
 
+class ExportDelimiter(str, Enum):
+    """Supported CSV delimiter options for export."""
+    comma = "comma"
+    tab = "tab"
+    semicolon = "semicolon"
+    pipe = "pipe"
+
+    def to_char(self) -> str:
+        """Return the actual delimiter character."""
+        return {"comma": ",", "tab": "\t", "semicolon": ";", "pipe": "|"}[self.value]
+
+
+class ExportEncoding(str, Enum):
+    """Supported file encoding options for export."""
+    utf8 = "utf-8"
+    latin1 = "latin-1"
+    ascii = "ascii"
+    utf16 = "utf-16"
+
+
 class CastDataTypeParams(BaseModel):
     """Parameters for casting a column to a different data type."""
     column: str
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -2,7 +2,7 @@
 
 import uuid
 from pydantic import BaseModel
-from enum import Enum
+from enum import Enum, StrEnum
 from typing import Optional, Union, Any, List
 import datetime
 
@@ -126,7 +126,7 @@ class DataType(str, Enum):
     datetime = "datetime"
 
 
-class ExportDelimiter(str, Enum):
+class ExportDelimiter(StrEnum):
     """Supported CSV delimiter options for export."""
     comma = "comma"
     tab = "tab"
@@ -135,15 +135,19 @@ class ExportDelimiter(str, Enum):
 
     def to_char(self) -> str:
         """Return the actual delimiter character."""
-        return {"comma": ",", "tab": "\t", "semicolon": ";", "pipe": "|"}[self.value]
+        match self:
+            case ExportDelimiter.comma:     return ","
+            case ExportDelimiter.tab:       return "\t"
+            case ExportDelimiter.semicolon: return ";"
+            case ExportDelimiter.pipe:      return "|"
 
 
-class ExportEncoding(str, Enum):
+class ExportEncoding(StrEnum):
     """Supported file encoding options for export."""
     utf8 = "utf-8"
     latin1 = "latin-1"
     ascii = "ascii"
-    utf16 = "utf-16"
+    utf16le = "utf-16-le"
 
 
 class CastDataTypeParams(BaseModel):
--- a/dataloom-backend/app/utils/pandas_helpers.py
+++ b/dataloom-backend/app/utils/pandas_helpers.py
@@ -26,6 +26,29 @@ def read_csv_safe(path: Path) -> pd.DataFrame:
         raise HTTPException(status_code=500, detail=f"Error reading CSV: {str(e)}")
 
 
+def read_csv_as_strings(path: Path) -> pd.DataFrame:
+    """Read a CSV preserving all values as their original string representations.
+
+    Uses dtype=str and keep_default_na=False to prevent pandas from silently
+    converting values (e.g. "00123"→123, "1.10"→"1.1", ""→NaN).
+
+    Args:
+        path: Path to the CSV file.
+
+    Returns:
+        DataFrame where every cell is a string.
+
+    Raises:
+        HTTPException: If the file cannot be read.
+    """
+    try:
+        return pd.read_csv(path, dtype=str, keep_default_na=False)
+    except FileNotFoundError:
+        raise HTTPException(status_code=404, detail=f"CSV file not found: {path}")
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=f"Error reading CSV: {str(e)}")
+
+
 def save_csv_safe(df: pd.DataFrame, path: Path) -> None:
     """Save a DataFrame to CSV safely.
 
--- a/dataloom-backend/tests/test_new_features.py
+++ b/dataloom-backend/tests/test_new_features.py
@@ -1,6 +1,7 @@
 """Tests for new features: rename column, cast data type, export, and delete project."""
 
 import csv
+import io
 
 import pandas as pd
 import pytest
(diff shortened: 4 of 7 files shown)


Denyme24:
hey @ivantha as discussed [here,](https://github.com/c2siorg/dataloom/issues/92#issuecomment-3959161211) , i have made the changes, PTAL.

ivantha (commented):
## ◎ Argus Review

This PR cleanly implements the feature from a UX and API-design perspective — backward-compatible defaults, automatic FastAPI enum validation for 422 errors, proper blob URL cleanup, and good JSDoc documentation. However, there are two high-severity issues that must be addressed before merging: an unhandled `UnicodeEncodeError` that causes silent 500s when data is incompatible with the chosen encoding, and unsanitized `project.name` usage in the `Content-Disposition` header which enables header injection and breaks non-ASCII filenames. Beyond those blockers, the DataFrame round-trip silently alters data values (leading zeros, date formats), the `StreamingResponse` wrapper provides no actual streaming benefit, and UTF-16 exports will include an unexpected BOM that breaks most CSV parsers. The complete absence of new tests for this functionality is also a significant gap given the number of distinct code paths introduced.

ivantha (commented):
## ◎ Argus Review

This is a well-structured and thoughtfully implemented PR that adds configurable CSV export options (delimiter, encoding, header toggle) with a clean backend/frontend split. Highlights include the fast-path optimization for the default case to avoid pandas data corruption, RFC 6266-compliant Content-Disposition headers, and a notably comprehensive test suite. However, there are two medium-severity issues that should be addressed before merging: (1) a potential HTTP header injection vulnerability because newline characters are not stripped from the project name before embedding it in the `Content-Disposition` header, and (2) a functional bug where the `loading` state is not reset in the `useEffect` cleanup, which can leave the Export button permanently disabled if the dialog is closed mid-request and then reopened. Several low-severity items around import hygiene, a missing `match` fallback, and a JSDoc inaccuracy are also worth fixing.

### ⚠️ Breaking Changes

- **[INFO]** Non-Breaking: `exportProject()` Signature Extended with Optional `options` Parameter: **Previous behavior**: `exportProject(projectId)` — single argument, always called the endpoint with no query parameters.

**New behavior**: `exportProject(projectId, { delimiter, include_header, encoding } = {})` — second argument is optional with defaults that match the old behavior.

**Who is affected**: Any code calling `exportProject(projectId)` without a second argument.

**Migration path**: No migration needed. All existing callers continue to work identically — the defaults (`comma`, `true`, `utf-8`) produce the same output as the previous no-parameter call, and the fast-path in the backend serves the raw file bytes for this combination.
  > **Suggestion:** No action required. This is a fully backward-compatible additive change.


### PR Description Issues

- **[MEDIUM]** "New Tests Added" Checkbox Is Unchecked Despite Extensive New Tests: The **How Has This Been Tested?** section has `- [ ] New tests added` unchecked, but the diff clearly adds **over 100 lines** of new tests across 15+ new test cases in `test_new_features.py`, including parametrized delimiter tests, encoding tests, Content-Disposition header tests, and Unicode edge-case tests. This is a direct factual mismatch between the PR description and the actual code changes.
  > **Suggestion:** Update the checklist to reflect reality:
```
- [X] Existing tests pass
- [X] New tests added  ← should be checked
- [ ] Manual testing
```
And briefly describe the new tests in the section body, e.g.: *"Added 15 new test cases covering delimiter variants, encoding variants (including incompatibility → 400), header toggle, Content-Disposition sanitization, and non-existent project handling."*
- **[LOW]** Duplicate 'Type of Change' Checklist Section: The PR description contains **two** "Type of Change" checkbox sections. The first has `Bug fix` and `New feature` checked; the second (immediately after the break line `---`) is a separate identical section with only `New feature` checked. This is leftover template boilerplate that was accidentally duplicated.
  > **Suggestion:** Remove the second (duplicate) checklist. The description should contain only one "Type of Change" section:
```markdown
## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
```
(The "Bug fix" tick is debatable — this is primarily a new feature, not a bug fix, so consider unchecking it.)
- **[LOW]** Empty '## Screenshots' Heading Under Testing Section: The **How Has This Been Tested?** section ends with a `## Screenshots (if applicable)` heading that has no content below it. Screenshots do exist in the PR — but they are placed in the **Description** section rather than here. This creates an empty heading that looks like an unfilled template placeholder.
  > **Suggestion:** Either remove the empty screenshots heading, or move/reference the screenshots to that section:
```markdown
## Screenshots

See images in the Description section above.
```

dataloom-backend/app/api/endpoints/projects.py
```diff
@@ -202,10 +204,76 @@ async def revert_to_checkpoint(
...
+    # for non-ASCII names and modern browsers.
+    safe_name = project.name.replace('"', "").replace("\\", "")
+    encoded_name = quote(f"{safe_name}.csv", safe="")
+    content_disposition = (
+        f'attachment; filename="{safe_name}.csv"; '
+        f"filename*=UTF-8''{encoded_name}"
+    )
+    headers = {"Content-Disposition": content_disposition}
+
+    logger.info(
```
ivantha, line 232: **[MEDIUM]** HTTP Header Injection via Unsanitized Newlines in Project Name

The `safe_name` sanitization strips `"` and `\` from the project name before embedding it in the `Content-Disposition` header, but it does **not** strip carriage returns (`\r`) or newlines (`\n`). An attacker who can control the project name (e.g., `evil\r\nX-Injected: pwned`) could inject arbitrary HTTP response headers.

```python
safe_name = project.name.replace('"', "").replace("\\", "")
# Missing: strip \r and \n
```

Although the `filename*` RFC 6266 encoded form would be safe, the legacy `filename="..."` quoted-string portion is not.

**Suggestion:**
```
Strip all control characters including CR/LF from the safe name before embedding in the header:
```python
import re
safe_name = re.sub(r'["\\\r\n]', '', project.name)
```
Or more broadly, allow only printable non-control characters:
```python
safe_name = "".join(c for c in project.name if c.isprintable() and c not in '"\\\r\n')
```
```

File: dataloom-frontend/src/Components/forms/ExportDialog.jsx [hunk 2]
```diff
@@ -0,0 +1,162 @@
...
+      setDelimiter("comma");
+      setIncludeHeader(true);
+      setEncoding("utf-8");
+      setError(null);
+    }
+  }, [isOpen]);
+
+  const handleExport = async () => {
+    setLoading(true);
+    setError(null);
```
ivantha, line 48: **[MEDIUM]** `loading` State Not Reset When Export Dialog Is Reopened

The `useEffect` in `ExportDialog` resets `delimiter`, `includeHeader`, `encoding`, and `error` when the dialog opens, but **omits `loading`**. This means if a user:
1. Opens the dialog and clicks **Export** (sets `loading = true`)
2. Clicks **Cancel** while the request is in-flight (closes dialog, but export fetch continues)
3. Reopens the dialog before the fetch resolves

The Export button will be permanently disabled because `loading` is still `true`, and the previous fetch's `finally { setLoading(false) }` will run on the still-mounted (but hidden) component — a race condition that may or may not rescue the state depending on timing.

```js
useEffect(() => {
  if (isOpen) {
    setDelimiter("comma");
    setIncludeHeader(true);
    setEncoding("utf-8");
    setError(null);
    // ← setLoading(false) is missing
  }
}, [isOpen]);
```

**Suggestion:**
```
Add `setLoading(false)` to the reset effect, and consider using an `AbortController` to cancel the in-flight request on close:
```js
useEffect(() => {
  if (isOpen) {
    setDelimiter("comma");
    setIncludeHeader(true);
    setEncoding("utf-8");
    setError(null);
    setLoading(false); // ← add this
  }
}, [isOpen]);
```
```

File: dataloom-backend/app/schemas.py [hunk 3]
```diff
@@ -141,6 +145,30 @@ class DataType(StrEnum):
...
+    pipe = "pipe"
+
+    def to_char(self) -> str:
+        """Return the actual delimiter character."""
+        match self:
+            case ExportDelimiter.comma:     return ","
+            case ExportDelimiter.tab:       return "\t"
+            case ExportDelimiter.semicolon: return ";"
+            case ExportDelimiter.pipe:      return "|"
+
```
ivantha, line 162: **[MEDIUM]** `to_char()` Has No Fallback Case — Returns `None` Silently on Unmatched Enum Values

The `match` statement in `ExportDelimiter.to_char()` covers all four current enum members, but contains no wildcard `case _:` guard. If a new enum member is ever added without a corresponding `case`, Python will fall through and implicitly return `None`. This `None` would then be passed as the `sep` argument to `df.to_csv()`, which accepts any single character — causing either a `TypeError` at runtime or, worse, silent incorrect output with no error raised for the caller.

```python
def to_char(self) -> str:
    match self:
        case ExportDelimiter.comma:     return ","
        case ExportDelimiter.tab:       return "\t"
        case ExportDelimiter.semicolon: return ";"
        case ExportDelimiter.pipe:      return "|"
        # no default!
```

**Suggestion:**
```
Add an explicit fallback case:
```python
def to_char(self) -> str:
    match self:
        case ExportDelimiter.comma:     return ","
        case ExportDelimiter.tab:       return "\t"
        case ExportDelimiter.semicolon: return ";"
        case ExportDelimiter.pipe:      return "|"
        case _:
            raise ValueError(f"No character mapping defined for delimiter: {self!r}")
```
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 4]
```diff
@@ -3,10 +3,12 @@
 Handles upload, retrieval, save (checkpoint), and revert operations.
 """
 
+import io
 import uuid
+from urllib.parse import quote
```
ivantha, line 8: **[LOW]** Unused `StreamingResponse` Import in `projects.py`

`StreamingResponse` is imported from `fastapi.responses` but is never used in the file. The implementation correctly chose `Response` (not `StreamingResponse`) for the non-default path, and this is even explained in an inline comment. However, the unused import was not cleaned up.

```python
from fastapi.responses import FileResponse, Response, StreamingResponse
#                                                       ^^^^^^^^^^^^^^^^ never used
```

This will likely trigger linters (e.g., `ruff`, `flake8`) and adds noise to the import block.

**Suggestion:**
```
Remove `StreamingResponse` from the import:
```python
from fastapi.responses import FileResponse, Response
```
```

File: dataloom-backend/app/schemas.py [hunk 5]
```diff
@@ -6,6 +6,10 @@
 from typing import Any
 
 from pydantic import BaseModel
+from enum import Enum
+from typing import Optional, Union, Any, List
+import datetime
+
```
ivantha, line 12: **[LOW]** Redundant and Potentially Unused Imports Added to `schemas.py`

The new additions to `schemas.py` introduce several import issues:

1. `from typing import Any` was already imported at the top of the file; the new block re-imports it alongside `Optional`, `Union`, and `List` — creating a duplicate.
2. `from enum import Enum` is imported but `Enum` is never referenced in the new code (the enums use `StrEnum`, which must already be available in the file).
3. `import datetime` is added but does not appear to be used by any of the new enums or models.

```python
# Already at top of file:
from typing import Any

# New additions (problematic):
from enum import Enum          # unused
from typing import Optional, Union, Any, List  # Any is duplicated
import datetime                # apparently unused
```

**Suggestion:**
```
Remove the redundant block entirely. If `Optional`, `Union`, or `List` are genuinely needed elsewhere in the file, merge them into the existing `from typing import ...` line at the top. Verify with a linter (`ruff check`) before merging.
```

dataloom-backend/app/utils/pandas_helpers.py
```diff
@@ -27,6 +27,29 @@ def read_csv_safe(path: Path) -> pd.DataFrame:
...
+        path: Path to the CSV file.
+
+    Returns:
+        DataFrame where every cell is a string.
+
+    Raises:
+        HTTPException: If the file cannot be read.
+    """
+    try:
+        return pd.read_csv(path, dtype=str, keep_default_na=False)
```
ivantha, line 46: **[LOW]** File System Path Exposed in 404 Error Message

In `read_csv_as_strings`, the `FileNotFoundError` handler constructs the detail string using the full filesystem path:

```python
except FileNotFoundError:
    raise HTTPException(status_code=404, detail=f"CSV file not found: {path}")
```

This leaks the server's internal directory structure to API clients in production, which is unnecessary information for the caller and could aid an attacker in understanding the deployment layout.

**Suggestion:**
```
Return a generic message to the client and log the path server-side:
```python
except FileNotFoundError:
    logger.error("CSV file not found: %s", path)
    raise HTTPException(status_code=404, detail="Project data file not found.")
```
```

File: dataloom-frontend/src/api/projects.js [hunk 7]
```diff
@@ -66,10 +66,15 @@ export const revertToCheckpoint = async (projectId, checkpointId) => {
 /**
  * Export the current working copy of a project as a CSV download.
  * @param {string} projectId - The project ID.
+ * @param {Object} [options] - Export format options.
```
ivantha, line 69: **[LOW]** JSDoc Comment Documents Wrong Encoding Value (`utf-16` vs `utf-16-le`)

The JSDoc comment in `projects.js` lists the supported encodings as `utf-8, latin-1, ascii, utf-16`, but the backend actually accepts `utf-16-le` (not `utf-16`). Sending `utf-16` as the encoding value would result in a `422` validation error.

```js
 * @param {string} [options.encoding="utf-8"] - File encoding: utf-8, latin-1, ascii, utf-16.
//                                                                                      ^^^^^^
//                                                                           Should be utf-16-le
```

**Suggestion:**
```
Correct the documented value:
```js
 * @param {string} [options.encoding="utf-8"] - File encoding: utf-8, latin-1, ascii, utf-16-le.
```
```

File: dataloom-frontend/src/Components/forms/ExportDialog.jsx [hunk 8]
```diff
@@ -0,0 +1,162 @@
...
+          <label htmlFor="export-encoding" className="text-sm font-medium text-gray-700">Encoding</label>
+          <select
+            id="export-encoding"
+            value={encoding}
+            onChange={(e) => setEncoding(e.target.value)}
+            className="border border-gray-300 rounded-md px-3 py-2 text-sm focus:outline-none focus:ring-2 focus:ring-blue-500"
+          >
+            {ENCODINGS.map((e) => (
+              <option key={e.value} value={e.value}>
+                {e.label}
```
ivantha, line 109: **[LOW]** Iterator Variable `e` Shadows Common `event` Convention in JSX

In the `ENCODINGS.map()` render block, the iterator parameter is named `e`:

```jsx
{ENCODINGS.map((e) => (
  <option key={e.value} value={e.value}>
    {e.label}
  </option>
))}
```

`e` is the universal shorthand for `event` objects in React handlers. Using it as a data iterator inside JSX is a common source of confusion during maintenance. The `DELIMITERS.map((d) => ...)` uses the more appropriate `d` pattern — the same should apply here.

**Suggestion:**
```
Rename the iterator to `enc` or `option` for clarity:
```jsx
{ENCODINGS.map((enc) => (
  <option key={enc.value} value={enc.value}>
    {enc.label}
  </option>
))}
```
```

File: dataloom-frontend/src/Components/forms/ExportDialog.jsx [hunk 9]
```diff
@@ -0,0 +1,162 @@
+import { useState, useEffect } from "react";
+import Modal from "../common/Modal";
+import { exportProject } from "../../api";
+import proptype from "prop-types";
```
ivantha, line 4: **[LOW]** PropTypes Import Variable Uses Non-Standard Casing (`proptype` vs `PropTypes`)

`ExportDialog.jsx` (and apparently `MenuNavbar.jsx` too, as an existing pattern) uses `import proptype from "prop-types"` with a lowercase `proptype`. The React community convention and the official prop-types package README use `PropTypes` (PascalCase) for this import, since it acts as a namespace/class.

This is a pre-existing style issue carried into the new file, but it's worth standardizing.

**Suggestion:**
```
Use the conventional casing:
```js
import PropTypes from "prop-types";
// ...
ExportDialog.propTypes = {
  isOpen: PropTypes.bool.isRequired,
  onClose: PropTypes.func.isRequired,
  projectId: PropTypes.string.isRequired,
  projectName: PropTypes.string,
};
```
```

File: dataloom-backend/app/utils/pandas_helpers.py [hunk 10]
```diff
@@ -27,6 +27,29 @@ def read_csv_safe(path: Path) -> pd.DataFrame:
...
+    except FileNotFoundError:
```
ivantha, line 47: **[LOW]** Exception Chain Lost in `read_csv_as_strings` for `FileNotFoundError`

In `read_csv_as_strings`, the `FileNotFoundError` handler does not use `raise ... from e` for exception chaining, unlike the `Exception` handler in the same function and the equivalent code in `read_csv_safe`:

```python
except FileNotFoundError:
    raise HTTPException(status_code=404, detail=f"CSV file not found: {path}")
    # ← should be: raise HTTPException(...) from e
except Exception as e:
    raise HTTPException(status_code=500, detail=f"Error reading CSV: {str(e)}")
    # ← this one correctly uses 'as e' but also missing 'from e'
```

Losing the exception chain makes debugging harder since the original traceback is suppressed in some Python configurations.

**Suggestion:**
```
Preserve the exception chain consistently:
```python
except FileNotFoundError as e:
    raise HTTPException(status_code=404, detail="Project data file not found.") from e
except Exception as e:
    raise HTTPException(status_code=500, detail=f"Error reading CSV: {str(e)}") from e
```
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 11]
```diff
@@ -202,10 +204,76 @@ async def revert_to_checkpoint(
...
+        )
+
+    # The full content is already in memory (required to catch the encoding
+    # error above), so use Response directly rather than giving a false
+    # impression of streaming via StreamingResponse.
+    return Response(
+        content=csv_bytes,
+        media_type=f"text/csv; charset={encoding.value}",
+        headers=headers,
+    )
```
ivantha, line 278: **[INFO]** Full In-Memory CSV Encoding for Large Files on Non-Default Paths

The non-default export path reads the entire CSV into a pandas DataFrame, serialises it to a `StringIO` buffer, then encodes the full string to bytes — all in memory. For large projects (e.g., hundreds of MBs), this triples peak memory usage. The comment explains why `StreamingResponse` was avoided (needing to catch `UnicodeEncodeError`), which is valid reasoning, but it's worth flagging for future scaling.

The fast path (default comma/UTF-8) correctly avoids this by serving the raw file, which is a good optimization.

**Suggestion:**
```
For a future improvement, consider chunked encoding with pre-validation of the encoding capability (e.g., test-encoding a sample of rows), or add file-size limits for non-default encoding requests. For now, documenting this behavior and its memory implications in the endpoint's docstring would be valuable.
```

dataloom-backend/tests/test_new_features.py
```diff
@@ -117,33 +118,146 @@ def test_replay_cast_data_type(self, sample_df):
...
+        assert r.status_code == 200
+        first_line = r.content.decode("utf-8").splitlines()[0]
+        # Data row starts with "Alice", not the column name "name"
+        assert first_line.startswith("Alice")
+
+    def test_export_with_header_includes_header_row(self, client, seeded_project):
+        r = client.get(f"/projects/{seeded_project}/export?include_header=true")
+        assert r.status_code == 200
+        first_line = r.content.decode("utf-8").splitlines()[0]
+        assert first_line.startswith("name")
```
ivantha, line 200: **[INFO]** Comma Delimiter Test Passes Trivially Due to Header Presence

The parametrized delimiter test checks that the expected separator byte is present in the response content:

```python
@pytest.mark.parametrize("delimiter,expected_sep", [
    ("comma",     b","),
    ...
])
def test_export_delimiter(self, client, seeded_project, delimiter, expected_sep):
    r = client.get(f"/projects/{seeded_project}/export?delimiter={delimiter}")
    assert r.status_code == 200
    assert expected_sep in r.content
```

For `comma`, `b","` will always appear in the content because the default fast-path serves the original file directly, which inherently contains commas. This test does not actually validate that the comma-delimited export path is functioning correctly — it would pass even if the delimiter parameter was completely ignored for the `comma` case.

**Suggestion:**
```
For a more rigorous comma test, parse the response as CSV and verify that each row has the expected number of fields:
```python
case ("comma", ...):
    rows = list(csv.reader(r.content.decode("utf-8").splitlines()))
    assert all(len(row) == 3 for row in rows)  # name, age, city
```
Alternatively, add a test with a dataset that contains no commas in values, ensuring the comma test is meaningful.
```

File: dataloom-frontend/src/Components/MenuNavbar.jsx [hunk 13]
```diff
@@ -296,6 +286,12 @@ const Menu_NavBar = ({ projectId, onTransform }) => {
           onTransform={onTransform}
         />
       )}
+      <ExportDialog
+        isOpen={showExportDialog}
+        onClose={() => setShowExportDialog(false)}
+        projectId={projectId}
```
ivantha, line 292: **[INFO]** `ExportDialog` Always Mounted Regardless of `isOpen` State

In `MenuNavbar.jsx`, `ExportDialog` is rendered unconditionally with `isOpen={showExportDialog}` rather than being conditionally mounted:

```jsx
<ExportDialog
  isOpen={showExportDialog}
  onClose={() => setShowExportDialog(false)}
  projectId={projectId}
  projectName={projectName}
/>
```

This is a common controlled-modal pattern and not a bug, but it means the component and its state (selects, error message, loading) persist in the DOM tree even when hidden. Since the `useEffect` properly resets state on open, this is functionally correct.

**Suggestion:**
```
No action required — the `useEffect` reset handles stale state. If bundle size or initial render time becomes a concern, conditional rendering with `{showExportDialog && <ExportDialog ... />}` would lazily mount/unmount the component.
```

ivantha:
Holding this PR — the latest CI run shows Frontend Lint and Backend Lint failing. Please fix the lint errors (`uv run ruff check .` / `uv run ruff format --check .` and `npm run lint`) and push so CI goes green before we can merge.

ivantha:
Thank you for implementing CSV export! This work has been superseded by #265 which provides multi-format export. Feel free to follow up on that PR.

PR #144, opened 2026-02-27
test: add comprehensive tests for Table component functionality

### LINKED ISSUE #84: [Chore] Add Table tests for context menu, Escape cancel, and row rendering
**Difficulty:** Medium
**Area:** frontend

## Description
Already covered, do not redo it: `dataloom-frontend/src/test/Table.columnOrder.test.jsx` renders `Table` with a mocked project context and a mocked `../api`, asserts column header order under a reordered `columnOrder`, and asserts that typing then pressing Enter saves a cell and calls `transformProject` with the backend column index. `dataloom-frontend/src/test/ContextMenu.test.jsx` covers the standalone `ContextMenu` component (open, Escape, click outside) but never the menu as the table opens it. What is left untested is the table's own interaction surface: the three `onContextMenu` handlers in `Table.tsx` and the menu entries they produce, Escape cancelling an in-progress cell edit with no API call, and the leading S.No. column plus the data rows actually rendering. Note that the component now enters edit mode on a single click through `handleCellClick`, not a double click, so any test written against the old double-click behaviour will not fire.

## Acceptance Criteria
- [ ] Right-clicking a column header opens the context menu showing Add Column, Delete Column, and Rename Column.
- [ ] Right-clicking a data cell opens the context menu showing Add Row and Delete Row.
- [ ] Pressing Escape while a cell is in edit mode clears the input and leaves `transformProject` uncalled.
- [ ] The header row renders `S.No.` ahead of the project's columns, and each data row renders its serial number followed by its cell values.
- [ ] Tests reuse the mock shape already established in `Table.columnOrder.test.jsx` (project context, `../api`, `HistoryRefreshContext`, `ToastProvider`) rather than introducing a second convention.
- [ ] `npm test` passes from `dataloom-frontend/`.

## Resources
- `dataloom-frontend/src/Components/Table.tsx:491-494` - the column header `onContextMenu`, gated on preview mode
- `dataloom-frontend/src/Components/Table.tsx:552` - the dtype row `onContextMenu`, which opens the same column menu
- `dataloom-frontend/src/Components/Table.tsx:579-583` - the data cell `onContextMenu` that opens the row menu
- `dataloom-frontend/src/Components/Table.tsx:640-656` - the menu entries the assertions should match, verbatim
- `dataloom-frontend/src/Components/Table.tsx:343-355` - `handleInputKeyDown`, where Enter saves and Escape clears without touching the API
- `dataloom-frontend/src/Components/Table.tsx:356-365` - `handleCellClick`, the single-click entry into edit mode
- `dataloom-frontend/src/Components/Table.tsx:140-150` - where `S.No.` is prepended to columns and the serial number to each row
- `dataloom-frontend/src/test/Table.columnOrder.test.jsx:9-58` - the existing mocks and `renderTable` helper to extend
- `dataloom-frontend/src/test/ContextMenu.test.jsx` - the standalone menu tests, for contrast with what is still missing
- `dataloom-frontend/src/hooks/useContextMenu.ts:13-25` - the hook that only holds state, so the listeners live in `ContextMenu`
- [@testing-library/react user events](https://testing-library.com/docs/user-event/intro)

Denyme24:
## Description

adds a comprehensive Vitest test suite for the `Table.jsx` component, which previously had zero test coverage despite being the most complex and central component in the codebase.

The test suite covers:
- **Rendering** -> `S.No.` column, column headers, row data, `externalData` prop override, and empty table
- **Cell editing** -> click to activate input, `S.No.` not editable, Enter saves + calls API, Escape cancels without API call, blur saves
- **Context menu (column)** ->  right-click opens menu with correct column options
- **Context menu (row)** -> right-click opens menu with correct row options
- **Context menu dismissal** -> clicking outside closes the menu
- **Context menu actions** -> Delete Row and Add Row correctly call `transformProject`

All dependencies (`useProjectContext`, `transformProject`) are properly mocked for isolated unit testing. React 18 + Vitest concurrent mode act() false-positives are suppressed with a targeted `console.error` filter.

Fixes #84

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [X] New tests added
- [ ] Manual testing

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/Table.jsx
+++ b/dataloom-frontend/src/Components/Table.jsx
@@ -223,7 +223,7 @@ const Table = ({ projectId, data: externalData }) => {
   };
 
   return (
-    <div className="px-8 pt-3" onClick={handleCloseContextMenu}>
+    <div className="px-8 pt-3" data-testid="table-container" onClick={handleCloseContextMenu}>
       <div
         className="overflow-x-scroll overflow-y-auto border border-gray-200 rounded-lg shadow-sm"
         style={{ maxHeight: "calc(100vh - 140px)" }}
--- a/dataloom-frontend/src/Components/__tests__/Table.test.jsx
+++ b/dataloom-frontend/src/Components/__tests__/Table.test.jsx
@@ -1,0 +1,313 @@
+import { render, screen, fireEvent, waitFor, act } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { vi, describe, it, expect, beforeAll, afterAll, beforeEach } from "vitest";
+import Table from "../Table";
+
+// --- Mocks ---
+
+vi.mock("../../context/ProjectContext", () => ({
+  useProjectContext: vi.fn(),
+}));
+
+vi.mock("../../api", () => ({
+  transformProject: vi.fn(),
+}));
+
+// React 18 + Vitest: concurrent mode can emit spurious act() warnings when
+// userEvent's pointer sequence (click → bubble → handleCloseContextMenu)
+// schedules a micro-task state update across two batching boundaries.
+// The tests themselves are correct; suppress only this known false-positive.
+let _origConsoleError;
+beforeAll(() => {
+  _origConsoleError = console.error.bind(console);
+  console.error = (...args) => {
+    if (typeof args[0] === "string" && args[0].includes("not wrapped in act")) return;
+    _origConsoleError(...args);
+  };
+});
+afterAll(() => {
+  console.error = _origConsoleError;
+});
+
+import { useProjectContext } from "../../context/ProjectContext";
+import { transformProject } from "../../api";
+
+// --- Helpers ---
+
+const COLUMNS = ["Name", "Score"];
+const ROWS = [
+  ["Alice", 90],
+  ["Bob", 85],
+];
+
+const mockContext = (overrides = {}) => {
+  useProjectContext.mockReturnValue({
+    columns: COLUMNS,
+    rows: ROWS,
+    dtypes: { Name: "str", Score: "int" },
+    updateData: vi.fn(),
+    ...overrides,
+  });
+};
+
+const renderTable = async (props = {}) => {
+  let result;
+  await act(async () => {
+    result = render(<Table projectId="test-project-id" {...props} />);
+  });
+  return result;
+};
+
+// --- Tests ---
+
+describe("Table", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockContext();
+  });
+
+  afterEach(async () => {
+    // Flush all pending React state updates / microtasks to avoid act() warnings
+    await act(async () => {});
+  });
+
+  // ── Rendering ──────────────────────────────────────────────────────────────
+
+  describe("rendering", () => {
+    it("renders S.No. column header", async () => {
+      await renderTable();
+      expect(screen.getByText("S.No.")).toBeInTheDocument();
+    });
+
+    it("renders all column headers from context", async () => {
+      await renderTable();
+      expect(screen.getByText("Name")).toBeInTheDocument();
+      expect(screen.getByText("Score")).toBeInTheDocument();
+    });
+
+    it("renders the correct number of data rows", async () => {
+      await renderTable();
+      // Each data row has a serial number cell
+      expect(screen.getByText("1")).toBeInTheDocument();
+      expect(screen.getByText("2")).toBeInTheDocument();
+    });
+
+    it("renders correct cell values from context", async () => {
+      await renderTable();
+      expect(screen.getByText("Alice")).toBeInTheDocument();
+      expect(screen.getByText("Bob")).toBeInTheDocument();
+      expect(screen.getByText("90")).toBeInTheDocument();
+      expect(screen.getByText("85")).toBeInTheDocument();
+    });
+
+    it("renders data from externalData prop, overriding context", async () => {
+      const externalData = {
+        columns: ["City"],
+        rows: [["London"]],
+        dtypes: {},
+      };
+      await renderTable({ data: externalData });
+      expect(screen.getByText("City")).toBeInTheDocument();
+      expect(screen.getByText("London")).toBeInTheDocument();
+      // context data should not be present
+      expect(screen.queryByText("Alice")).not.toBeInTheDocument();
+    });
+
+    it("renders an empty table body when context has no rows", async () => {
+      mockContext({ columns: [], rows: [] });
+      const { container } = await renderTable();
+      const rows = container.querySelectorAll("tbody tr");
+      expect(rows).toHaveLength(0);
+    });
+  });
+
+  // ── Context Menu – Column ──────────────────────────────────────────────────
+
+  describe("column context menu", () => {
+    it("opens with correct options on right-click of a column header", async () => {
+      await renderTable();
+      const nameHeader = screen.getByText("Name").closest("th");
+      fireEvent.contextMenu(nameHeader);
+
+      expect(screen.getByText("Add Column")).toBeInTheDocument();
+      expect(screen.getByText("Delete Column")).toBeInTheDocument();
+      expect(screen.getByText("Rename Column")).toBeInTheDocument();
+    });
+
+    it("does not show row menu options when column header is right-clicked", async () => {
+      await renderTable();
+      const nameHeader = screen.getByText("Name").closest("th");
+      fireEvent.contextMenu(nameHeader);
+
+      expect(screen.queryByText("Add Row")).not.toBeInTheDocument();
+      expect(screen.queryByText("Delete Row")).not.toBeInTheDocument();
+    });
+  });
+
+  // ── Context Menu – Row ─────────────────────────────────────────────────────
+
+  describe("row context menu", () => {
+    it("opens with correct options on right-click of a row cell", async () => {
+      await renderTable();
+      const aliceCell = screen.getByText("Alice").closest("td");
+      fireEvent.contextMenu(aliceCell);
+
+      expect(screen.getByText("Add Row")).toBeInTheDocument();
+      expect(screen.getByText("Delete Row")).toBeInTheDocument();
+    });
+
+    it("does not show column menu options when a row cell is right-clicked", async () => {
+      await renderTable();
+      const aliceCell = screen.getByText("Alice").closest("td");
+      fireEvent.contextMenu(aliceCell);
+
+      expect(screen.queryByText("Add Column")).not.toBeInTheDocument();
+      expect(screen.queryByText("Delete Column")).not.toBeInTheDocument();
+    });
+  });
+
+  // ── Context Menu – Dismiss ─────────────────────────────────────────────────
+
+  describe("context menu dismissal", () => {
+    it("closes the context menu when clicking outside", async () => {
+      await renderTable();
+      const nameHeader = screen.getByText("Name").closest("th");
+      fireEvent.contextMenu(nameHeader);
+      expect(screen.getByText("Add Column")).toBeInTheDocument();
+
+      // Click the outer container div
+      const container = screen.getByText("Name").closest(".px-8");
+      fireEvent.click(container);
+
+      expect(screen.queryByText("Add Column")).not.toBeInTheDocument();
+    });
+  });
+
+  // ── Cell Editing ───────────────────────────────────────────────────────────
+
+  describe("cell editing", () => {
+    it("activates input mode when a data cell is clicked", async () => {
+      await renderTable();
+      const aliceCell = screen.getByText("Alice");
+      await userEvent.click(aliceCell);
+
+      const input = screen.getByRole("textbox");
+      expect(input).toBeInTheDocument();
+      expect(input).toHaveValue("Alice");
+    });
+
+    it("does not activate editing when the S.No. cell is clicked", async () => {
+      await renderTable();
+      const serialCell = screen.getByText("1");
+      await userEvent.click(serialCell);
+
+      expect(screen.queryByRole("textbox")).not.toBeInTheDocument();
+    });
+
+    it("calls transformProject with correct params when Enter is pressed to save", async () => {
+      transformProject.mockResolvedValue({
+        columns: COLUMNS,
+        rows: ROWS,
+        dtypes: { Name: "str", Score: "int" },
+      });
+
+      await renderTable();
+      await userEvent.click(screen.getByText("Alice"));
+
+      const input = screen.getByRole("textbox");
+      await userEvent.clear(input);
+      await userEvent.type(input, "Charlie");
+      await userEvent.keyboard("{Enter}");
+
+      await waitFor(() => {
+        expect(transformProject).toHaveBeenCalledWith("test-project-id", {
+          operation_type: "changeCellValue",
+          change_cell_value: {
+            col_index: 1, // first editable column (index 0 is S.No.)
+            row_index: 0,
+            fill_value: "Charlie",
+          },
+        });
+      });
+    });
+
+    it("dismisses edit mode without calling the API when Escape is pressed", async () => {
+      await renderTable();
+      await userEvent.click(screen.getByText("Alice"));
+
+      const input = screen.getByRole("textbox");
+      expect(input).toBeInTheDocument();
+
+      fireEvent.keyDown(input, { key: "Escape" });
+
+      expect(screen.queryByRole("textbox")).not.toBeInTheDocument();
+      expect(transformProject).not.toHaveBeenCalled();
+    });
+
+    it("saves the edit on input blur", async () => {
+      transformProject.mockResolvedValue({
+        columns: COLUMNS,
+        rows: ROWS,
+        dtypes: { Name: "str", Score: "int" },
+      });
+
+      await renderTable();
+      await userEvent.click(screen.getByText("Alice"));
+
+      const input = screen.getByRole("textbox");
+      await userEvent.clear(input);
+      await userEvent.type(input, "Dave");
+      fireEvent.blur(input);
+
+      await waitFor(() => {
+        expect(transformProject).toHaveBeenCalledTimes(1);
+      });
+    });
+  });
+
+  // ── Context Menu Actions ───────────────────────────────────────────────────
+
+  describe("context menu actions", () => {
+    it("calls transformProject when Delete Row is clicked", async () => {
+      transformProject.mockResolvedValue({
+        columns: COLUMNS,
+        rows: [["Bob", 85]],
+        dtypes: { Name: "str", Score: "int" },
+      });
+
+      await renderTable();
+      const aliceCell = screen.getByText("Alice").closest("td");
+      fireEvent.contextMenu(aliceCell);
+
+      await userEvent.click(screen.getByText("Delete Row"));
+
+      await waitFor(() => {
+        expect(transformProject).toHaveBeenCalledWith("test-project-id", {
+          operation_type: "delRow",
+          row_params: { index: 0 },
+        });
+      });
+    });
+
+    it("calls transformProject when Add Row is clicked", async () => {
+      transformProject.mockResolvedValue({
+        columns: COLUMNS,
+        rows: [...ROWS, ["", ""]],
+        dtypes: {},
+      });
+
+      await renderTable();
+      const aliceCell = screen.getByText("Alice").closest("td");
+      fireEvent.contextMenu(aliceCell);
+
+      await userEvent.click(screen.getByText("Add Row"));
+
+      await waitFor(() => {
+        expect(transformProject).toHaveBeenCalledWith("test-project-id", {
+          operation_type: "addRow",
+          row_params: { index: 0 },
+        });
+      });
+    });
+  });
+});
--- a/dataloom-frontend/src/Components/__tests__/Table.test.jsx
+++ b/dataloom-frontend/src/Components/__tests__/Table.test.jsx
@@ -1,9 +1,13 @@
-import { render, screen, fireEvent, waitFor, act } from "@testing-library/react";
+import { render, screen, waitFor, act } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { vi, describe, it, expect, beforeAll, afterAll, beforeEach } from "vitest";
+import { vi, describe, it, expect, beforeEach } from "vitest";
 import Table from "../Table";
+import { useProjectContext } from "../../context/ProjectContext";
+import { transformProject } from "../../api";
 
 // --- Mocks ---
+// vi.mock() calls are hoisted to the top of the file by Vitest at compile time,
+// so mocks take effect before any imports are evaluated regardless of source order.
 
 vi.mock("../../context/ProjectContext", () => ({
   useProjectContext: vi.fn(),
@@ -13,32 +17,12 @@ vi.mock("../../api", () => ({
   transformProject: vi.fn(),
 }));
 
-// React 18 + Vitest: concurrent mode can emit spurious act() warnings when
-// userEvent's pointer sequence (click → bubble → handleCloseContextMenu)
-// schedules a micro-task state update across two batching boundaries.
-// The tests themselves are correct; suppress only this known false-positive.
-let _origConsoleError;
-beforeAll(() => {
-  _origConsoleError = console.error.bind(console);
-  console.error = (...args) => {
-    if (typeof args[0] === "string" && args[0].includes("not wrapped in act")) return;
-    _origConsoleError(...args);
-  };
-});
-afterAll(() => {
-  console.error = _origConsoleError;
-});
-
-import { useProjectContext } from "../../context/ProjectContext";
-import { transformProject } from "../../api";
-
 // --- Helpers ---
 
-const COLUMNS = ["Name", "Score"];
-const ROWS = [
-  ["Alice", 90],
-  ["Bob", 85],
-];
+// Object.freeze prevents accidental mutation of shared test fixtures that
+// would cause hard-to-debug cross-test pollution.
+const COLUMNS = Object.freeze(["Name", "Score"]);
+const ROWS = Object.freeze([Object.freeze(["Alice", 90]), Object.freeze(["Bob", 85])]);
 
 const mockContext = (overrides = {}) => {
   useProjectContext.mockReturnValue({
@@ -61,14 +45,14 @@ const renderTable = async (props = {}) => {
 // --- Tests ---
 
 describe("Table", () => {
+  let user;
+
   beforeEach(() => {
     vi.clearAllMocks();
     mockContext();
-  });
-
-  afterEach(async () => {
-    // Flush all pending React state updates / microtasks to avoid act() warnings
-    await act(async () => {});
+    // userEvent.setup() wraps every interaction in act() automatically,
+    // which properly flushes React state updates and avoids act() warnings.
+    user = userEvent.setup();
   });
 
   // ── Rendering ──────────────────────────────────────────────────────────────
@@ -86,10 +70,11 @@ describe("Table", () => {
     });
 
     it("renders the correct number of data rows", async () => {
-      await renderTable();
-      // Each data row has a serial number cell
-      expect(screen.getByText("1")).toBeInTheDocument();
-      expect(screen.getByText("2")).toBeInTheDocument();
+      const { container } = await renderTable();
+      // Query the DOM structure directly: avoids false positives from any other
+      // element containing the text "1" or "2" (e.g., data cells or pagination).
+      const dataRows = container.querySelectorAll("tbody tr");
+      expect(dataRows).toHaveLength(ROWS.length);
     });
 
     it("renders correct cell values from context", async () => {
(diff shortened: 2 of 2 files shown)


Denyme24:
hey @ivantha as [discussed](https://github.com/c2siorg/dataloom/issues/84#issuecomment-3966787018), have made the changes.
PTAL.

dataloom-frontend/src/Components/__tests__/Table.test.jsx
```diff
@@ -0,0 +1,313 @@
...
+    updateData: vi.fn(),
+    ...overrides,
+  });
+};
+
+const renderTable = async (props = {}) => {
+  let result;
+  await act(async () => {
+    result = render(<Table projectId="test-project-id" {...props} />);
+  });
```
ivantha: **[INFO]** Consider extracting `renderTable` helper to a shared test utility

The `renderTable` helper and the `mockContext` function are excellent patterns for reducing boilerplate. However, if other test files are added for sibling components that also need to render `Table` (e.g., an integration test for a parent page), these helpers would need to be duplicated.

This is a forward-looking note rather than a current bug.

**Suggestion:**
```
When the test suite grows, move `renderTable`, `mockContext`, `COLUMNS`, and `ROWS` into a shared test utility file (e.g., `src/Components/__tests__/testUtils.jsx`) and import them into each test file that needs them.
```

ivantha (commented):
## ◎ Argus Review

This PR introduces solid, well-organized test coverage for the `Table` component — a meaningful improvement over zero tests. The describe-block hierarchy is clean, the mock setup is largely correct, and the positive/negative pairing in context menu tests is thorough. However, two issues require attention before merging: (1) the global suppression of `act()` warnings actively hides potential real React concurrency issues and must be replaced with proper `userEvent.setup()` usage, and (2) the `col_index: 1` assertion in the cell-editing test needs verification against the actual API contract, as an off-by-one here would mean the test passes while the component sends wrong indices to the backend. Several medium-severity issues (fragile CSS-class selector, missing API error tests, no column action tests) and low-severity style inconsistencies (mixed `fireEvent`/`userEvent`, non-standard import order) should also be addressed to keep the test suite maintainable long-term.

ivantha (commented):
## ◎ Argus Review

This PR adds a well-structured, comprehensive test suite for the `Table` component with solid coverage across rendering, cell editing, context menus, and error handling. The test architecture (frozen fixtures, `mockContext`/`renderTable` helpers, `userEvent.setup()`) demonstrates good testing instincts and the inline comments are genuinely valuable. However, there are two significant technical issues that likely cause flaky or incorrect behavior in CI: (1) `waitFor` is repeatedly nested inside `act()` — an explicit anti-pattern that can mask real warnings and create timing issues — and (2) `userEvent.setup()` interactions are redundantly double-wrapped in manual `act()` despite the author's own comment noting this is unnecessary. There is also a Rename Column test that likely submits the wrong value due to not clearing the pre-filled input, and fragile `getByText('1')`/`getByText('2')` selectors that are coupled to fixture values. Finally, the PR description claims a `console.error` filter was added that is entirely absent from the diff.

### ⚠️ Breaking Changes

- **[LOW]** Non-breaking: `data-testid` attribute added to Table component's root element: **Previous behavior**: The outermost `<div>` in `Table.jsx` had `className="px-8 pt-3"` and an `onClick` handler but no `data-testid`.

**New behavior**: The element now also carries `data-testid="table-container"`.

**Who is affected**: No consumers are broken — adding a HTML attribute is purely additive. Snapshot tests or DOM-diffing tests that assert on the exact HTML structure of Table's root element could unexpectedly fail.

**Migration path**: If any existing snapshot tests cover `Table.jsx`, run `vitest --update-snapshots` to accept the diff.
  > **Suggestion:** Check for existing snapshot tests (`*.snap` files) containing Table's rendered HTML and update them. This is a one-time `--update-snapshots` run.


### PR Description Issues

- **[MEDIUM]** PR description claims `console.error` suppression exists but it's absent from the diff: The PR description states: *"React 18 + Vitest concurrent mode act() false-positives are suppressed with a targeted `console.error` filter."* However, scanning the entire diff, there is no `console.error` spy, mock, or filter anywhere in the test file. This is a factual mismatch between the description and the code.

This is concerning because either:
1. The author intended to add this and forgot, meaning act() warnings are firing in CI.
2. The description was copied from a template or draft and not updated.
  > **Suggestion:** Either add the console.error filter to the test file:
```js
beforeAll(() => {
  vi.spyOn(console, 'error').mockImplementation((msg, ...args) => {
    if (typeof msg === 'string' && /act\(\.\.\.\/not wrapped in act/.test(msg)) return;
    originalConsoleError(msg, ...args);
  });
});
```
Or remove the claim from the PR description if the filter is not needed.
- **[LOW]** PR incorrectly categorizes new tests as a 'New feature': The PR description has `[X] New feature` checked under "Type of Change". A test-only PR (no new user-facing functionality is added) is typically categorized as `tests`, `chore`, or `refactor` — not `New feature`. This affects changelog generation tools that read these checkboxes and could cause the PR to be misclassified in release notes.
  > **Suggestion:** Update the PR description:
```
## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] Tests (new test coverage for existing functionality)
```
If the template doesn't have a Tests option, consider adding one or checking `chore`/`refactor` if available.
- **[LOW]** Several checklist items left unchecked without explanation: The following checklist items are unchecked:
- `[ ] My code follows the project's style guidelines`
- `[ ] I have added/updated documentation as needed`

The style guidelines item is notable for a test-only PR — it's reasonable to expect these tests would follow the project's testing conventions, and leaving it unchecked without explanation suggests an oversight.
  > **Suggestion:** Either check the applicable items, or add a note explaining why they don't apply:
```
- [X] I have added/updated documentation as needed (N/A — test-only change, no public API modified)
```
- **[INFO]** PR description is detailed and well-structured overall: The PR description is genuinely strong: it provides a clear summary of what was covered, enumerates tested scenarios in organized bullet groups, explains mock strategy and rationale, and references the issue it fixes (`#84`). The inline explanation of the act() concern (even though the implementation is absent) shows the author understands React 18 testing nuances. This level of detail makes review significantly easier.
  > **Suggestion:** No structural change needed. Address the `console.error` claim mismatch and the type-of-change checkbox as noted in other findings.

dataloom-frontend/src/Components/__tests__/Table.test.jsx
```diff
@@ -0,0 +1,545 @@
...
+      });
+
+      await renderTable();
+      // Open the cell for editing first, then grab the input from the DOM
+      // before starting the type+save interaction in a separate act() scope.
+      await act(async () => {
+        await user.click(screen.getByText("Alice"));
+      });
+
+      const input = screen.getByRole("textbox");
```
ivantha, line 230: **[HIGH]** `waitFor` nested inside `act()` is an anti-pattern that can hide real bugs

Throughout the test file, the pattern `await act(async () => { await user.click(...); await waitFor(() => { expect(...) }) })` is used repeatedly. This is explicitly discouraged by the Testing Library maintainers.

`waitFor` is itself a polling mechanism that flushes microtasks internally. Nesting it inside `act()` can suppress real "not wrapped in act()" warnings that signal genuine test-infrastructure issues, mask real race conditions, and create subtle timing deadlocks where `act()` drains the queue while `waitFor` tries to re-enter it.

The correct pattern is to:
1. `await` the userEvent interaction (which, when using `userEvent.setup()`, already wraps itself in act).
2. Use `await waitFor(...)` **outside** any surrounding `act()` block.

```js
// ❌ Current (anti-pattern)
await act(async () => {
  await user.keyboard('{Enter}');
  await waitFor(() => {
    expect(transformProject).toHaveBeenCalled();
  });
});

// ✅ Correct
await user.keyboard('{Enter}');
await waitFor(() => {
  expect(transformProject).toHaveBeenCalled();
});
```

This pattern recurs at approximately lines 207–226, 244–260, 315–332, 340–360, 388–400, 420–435, 442–455, 476–490, and 502–520.

**Suggestion:**
```
Remove the outer `act()` wrappers from interactions that use `userEvent.setup()` — `userEvent.setup()` already handles act() internally. Keep `waitFor` calls at the top level of the test body. Only use manual `act()` if you are triggering an imperative React state update outside of a user interaction.
```

dataloom-frontend/src/Components/__tests__/Table.test.jsx
```diff
@@ -0,0 +1,545 @@
...
+        await user.keyboard("{Enter}");
+        await waitFor(() => {
+          expect(transformProject).toHaveBeenCalledWith("test-project-id", {
+            operation_type: "changeCellValue",
+            change_cell_value: {
+              col_index: 2, // Score is at display index 2 (S.No.=0, Name=1, Score=2)
+              row_index: 0,
+              fill_value: "95",
+            },
+          });
```
ivantha, line 284: **[HIGH]** Fragile `getByText('1')` and `getByText('2')` selectors will break if any data cell contains those values

Two tests use `screen.getByText('1')` and `screen.getByText('2')` to locate S.No. cells:

```js
// line ~281
await user.click(screen.getByText("1")); // intended: S.No. row 1

// line ~390
await user.click(screen.getByText("2")); // intended: S.No. of Bob
```

If any data cell in `ROWS` contains the strings `"1"` or `"2"`, `getByText` will throw `TestingLibraryElementError: Found multiple elements with text '1'`. The current `ROWS` fixture has `90` and `85`, which are safe — but this is an invisible coupling between test fixtures. Renaming a ROWS value to `1` or `2` would silently break entirely different tests, making failures confusing to diagnose.

**Suggestion:**
```
Use a more specific query that targets the S.No. column explicitly, such as a `data-testid` on the S.No. cells or a within() scope:
```js
// Option A: add data-testid="sno-cell" to the S.No. <td> in Table.jsx
const snoCell = screen.getAllByTestId('sno-cell')[0];
await user.click(snoCell);

// Option B: scope the query to the first tbody row
import { within } from '@testing-library/react';
const firstRow = container.querySelectorAll('tbody tr')[0];
const snoCell = within(firstRow).getByText('1');
await user.click(snoCell);
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 4]
```diff
@@ -0,0 +1,545 @@
...
+
+    it("calls transformProject when Delete Column is clicked", async () => {
+      transformProject.mockResolvedValue({
+        columns: ["Score"],
+        rows: [[90], [85]],
+        dtypes: { Score: "int" },
+      });
+
+      await renderTable();
+      // Right-click on the "Name" header (display index 1); the handler strips
```
ivantha, line 402: **[HIGH]** Rename Column test likely submits 'NameFullName' because the dialog input is not cleared first

The Rename Column test does:
```js
const dialogInput = screen.getByRole('textbox');
await act(async () => {
  await user.type(dialogInput, 'FullName');
  await user.click(screen.getByRole('button', { name: /ok/i }));
  await waitFor(() => {
    expect(transformProject).toHaveBeenCalledWith('test-project-id', {
      operation_type: 'renameCol',
      rename_col_params: { col_index: 0, new_name: 'FullName' },
    });
  });
});
```

If `InputDialog` pre-fills the existing column name (a very common UX pattern for rename dialogs), the input will start with `'Name'`. `userEvent.type` **appends** characters to the existing value, producing `'NameFullName'`. The API expectation of `new_name: 'FullName'` would then fail — meaning this test would never pass in a real run, or it only accidentally passes because the dialog starts empty (in which case the lack of a `clear` before typing is still a maintenance hazard).

**Suggestion:**
```
Always clear the input before typing the new value in rename tests:
```js
await user.clear(dialogInput);
await user.type(dialogInput, 'FullName');
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 5]
```diff
@@ -0,0 +1,545 @@
...
+
+describe("Table", () => {
+  let user;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockContext();
+    // userEvent.setup() wraps every interaction in act() automatically,
+    // which properly flushes React state updates and avoids act() warnings.
+    user = userEvent.setup();
```
ivantha, line 55: **[MEDIUM]** Redundant `act()` wrappers around `userEvent.setup()` interactions

The test file wraps nearly every `userEvent` interaction in a manual `act()` call:
```js
await act(async () => {
  await user.click(screen.getByText('Alice'));
});
```

However, `user = userEvent.setup()` (as set up in `beforeEach`) already wraps every interaction in `act()` internally — this is a primary feature of the setup API vs the legacy `userEvent.v13` API. The double-wrapping is redundant noise and can, in some edge cases, cause subtle issues with React's batched state updates being flushed at the wrong boundary.

The comment in `beforeEach` correctly states this: *"userEvent.setup() wraps every interaction in act() automatically"* — but then proceeds to wrap them again everywhere.

**Suggestion:**
```
Remove the outer `act()` calls from all `userEvent` interactions. The tests become cleaner and more readable:
```js
// ❌ Before
await act(async () => {
  await user.click(screen.getByText('Alice'));
});

// ✅ After
await user.click(screen.getByText('Alice'));
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 6]
```diff
@@ -0,0 +1,545 @@
...
+        await user.click(screen.getByRole("button", { name: /ok/i }));
+        await waitFor(() => {
+          expect(transformProject).toHaveBeenCalledWith("test-project-id", {
+            operation_type: "renameCol",
+            rename_col_params: { col_index: 0, new_name: "FullName" },
+          });
+        });
+      });
+    });
+
```
ivantha, line 455: **[MEDIUM]** Inconsistent `col_index` semantics between `addCol` and `delCol` API payloads in tests

The two column-operation tests document conflicting index semantics without a clear resolution:

- **Delete Column** (Name at display index 1): The comment says "the handler strips the S.No. offset so the API receives `col_params.index = 1 - 1 = 0`". Expectation: `col_params: { index: 0 }`.
- **Add Column** (also triggered from Name's context menu at display index 1): Expectation: `col_params: { index: 1, name: 'City' }` — the S.No. offset is **not** subtracted.

This inconsistency suggests either:
1. The actual Table.jsx component sends different index conventions to the API for `addCol` vs `delCol`, which would be a bug in the component.
2. One of the test expectations is wrong and would fail against the real component.

The tests document and cement an asymmetric API contract without flagging it, which is misleading to future maintainers.

**Suggestion:**
```
Add an explicit comment to the Add Column test explaining why the index semantics differ from Delete Column, OR verify the actual Table.jsx source to confirm the correct indices, and adjust the test expectations (and a bug report if needed):
```js
// In Add Column test — if the API truly uses display index for addCol:
// col_params: { index: 1, name: 'City' }  // display index, S.No. NOT subtracted
// vs delCol which subtracts 1 → col_params: { index: 0 }
// This asymmetry should be confirmed against Table.jsx source and
// ideally normalized in the component to avoid confusion.
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 7]
```diff
@@ -0,0 +1,545 @@
...
+    it("does not activate editing when the S.No. cell is clicked", async () => {
+      await renderTable();
+      await act(async () => {
+        await user.click(screen.getByText("1"));
+      });
+
+      expect(screen.queryByRole("textbox")).not.toBeInTheDocument();
+    });
+
```
ivantha, line 215: **[MEDIUM]** Cell-edit `col_index` documents a potential off-by-one API bug without flagging it

The test comments explicitly note that `changeCellValue` sends `col_index: 1` for the 'Name' column (display index, including S.No. at 0), while column operations like `delCol` subtract 1 to get a 0-based data index. This asymmetry means the backend API must handle two different indexing schemes depending on operation type — a fragile design.

If the backend actually expects a 0-based column index for `changeCellValue` (consistent with `delCol`), then the tests are documenting a real bug in `Table.jsx`. The tests treat these as expected correct behavior rather than flagging the inconsistency.

The comment says: *"Note: column operations (delCol/renameCol) subtract 1 to get a 0-based data index — cell editing does not; it sends the display index."* This is a critical observation buried in a comment rather than raised as a question.

**Suggestion:**
```
Add a TODO or question in the PR/issue tracker about whether the backend API truly expects `col_index: 1` (display) for `changeCellValue` vs `col_index: 0` (data) for `delCol`. Consider adding a unit test of the backend endpoint contract or referencing the API spec. If this is a bug, the test is encoding the wrong behavior.
```

dataloom-frontend/src/Components/__tests__/Table.test.jsx
```diff
@@ -0,0 +1,545 @@
...
+      // fires after the above act() exits.
+      await act(async () => {
+        await waitFor(() => {
+          expect(screen.getByRole("alert")).toBeInTheDocument();
+          expect(screen.getByText("Failed to delete row. Please try again.")).toBeInTheDocument();
+        });
+      });
+    });
+  });
+});
```
ivantha, line 545: **[MEDIUM]** API error handling tests only cover two operations out of many; error toast content is hardcoded

The error handling section tests `changeCellValue` and `delRow` failures, but the component presumably has similar error handling for `addRow`, `delCol`, `addCol`, `renameCol` as well. If those operations' catch blocks have different error messages or broken toast calls, no test would catch them.

Additionally, the tests assert on very specific error message strings:
```js
expect(screen.getByText('Failed to edit cell. Please try again.')).toBeInTheDocument();
expect(screen.getByText('Failed to delete row. Please try again.')).toBeInTheDocument();
```

This tightly couples the tests to the exact error message wording — a copy change in the component would break the tests. Considering that the test file is intended to be the primary test for this component, the missing coverage and message-string coupling are notable gaps.

**Suggestion:**
```
Either:
1. Add error tests for `addRow`, `delCol`, `addCol`, `renameCol`.
2. Or assert on the presence of the alert role without the exact message: `expect(screen.getByRole('alert')).toBeInTheDocument()`. This is more resilient to copy changes while still verifying the toast appears.
```js
// More resilient
expect(screen.getByRole('alert')).toBeInTheDocument();
// If you must check content, use a partial match:
expect(screen.getByRole('alert')).toHaveTextContent(/failed/i);
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 9]
```diff
@@ -0,0 +1,545 @@
...
+      });
+
+      await renderTable();
+      await act(async () => {
+        await user.pointer({
+          target: screen.getByText("Alice").closest("td"),
+          keys: "[MouseRight]",
+        });
+      });
+
```
ivantha, line 356: **[MEDIUM]** Cell-edit API call tests do not verify that `editingCell` is cleared after save

The tests for Enter-saves and blur-saves only assert that `transformProject` was called with the correct arguments. They do not verify that the editing UI (the `<input>`) is dismissed after a successful save. This means if the component calls the API correctly but fails to call `setEditingCell(null)` (e.g., a regression), the test would still pass.

A complete test for a save action should verify both the side effect (API call) and the resulting UI state (input disappears, cell text reappears).

**Suggestion:**
```
After the API-call assertion, add a UI state assertion:
```js
await waitFor(() => {
  expect(transformProject).toHaveBeenCalledWith(...);
});
// Also verify the input is dismissed
expect(screen.queryByRole('textbox')).not.toBeInTheDocument();
// And the updated value is rendered
expect(screen.getByText('Charlie')).toBeInTheDocument();
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 10]
```diff
@@ -0,0 +1,545 @@
+import { render, screen, waitFor, act } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { vi, describe, it, expect, beforeEach } from "vitest";
+import Table from "../Table";
+import { useProjectContext } from "../../context/ProjectContext";
+import { transformProject } from "../../api";
```
ivantha, line 6: **[LOW]** Missing `afterEach` or `afterAll` cleanup for `console.error` mock if it were added

The PR description states that act() false-positive warnings are suppressed with a targeted `console.error` filter. No such filter exists in the actual code. If such a filter were added in the future (e.g., `beforeAll(() => { vi.spyOn(console, 'error').mockImplementation(...) })`), it would require a corresponding `afterAll` or `afterEach` to restore `console.error` — otherwise it would leak across test files and silently suppress real errors in other test suites.

**Suggestion:**
```
If adding a console.error spy in the future:
```js
let consoleErrorSpy;
beforeAll(() => {
  consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation((msg, ...args) => {
    if (typeof msg === 'string' && msg.includes('act(...)')) return; // suppress false positive
    console.error(msg, ...args); // pass through real errors
  });
});
afterAll(() => { consoleErrorSpy.mockRestore(); });
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 11]
```diff
@@ -0,0 +1,545 @@
...
+
+  // ── Context Menu – Row ─────────────────────────────────────────────────────
+
+  describe("row context menu", () => {
+    it("opens with correct options on right-click of a row cell", async () => {
+      await renderTable();
+      await act(async () => {
+        await user.pointer({
+          target: screen.getByText("Alice").closest("td"),
+          keys: "[MouseRight]",
```
ivantha, line 148: **[LOW]** Missing negative assertions for context menu option exclusivity

The column context menu tests only verify the presence of column-specific options. The negative assertions that row-only options are absent (`Add Row`, `Delete Row`) are correctly present in `does not show row menu options`. However, the row context menu tests do not check that `Rename Column` is absent when a row cell is right-clicked — only `Add Column` and `Delete Column` are checked. If the component accidentally renders all options for rows, `Rename Column` would be undetected.

**Suggestion:**
```
Add `Rename Column` to the negative assertions in the row context menu test:
```js
expect(screen.queryByText('Add Column')).not.toBeInTheDocument();
expect(screen.queryByText('Delete Column')).not.toBeInTheDocument();
expect(screen.queryByText('Rename Column')).not.toBeInTheDocument(); // add this
```
```

File: dataloom-frontend/src/Components/Table.jsx [hunk 12]
```diff
@@ -223,7 +223,7 @@ const Table = ({ projectId, data: externalData }) => {
   };
 
   return (
-    <div className="px-8 pt-3" onClick={handleCloseContextMenu}>
+    <div className="px-8 pt-3" data-testid="table-container" onClick={handleCloseContextMenu}>
```
ivantha, line 226: **[LOW]** `data-testid` added to production component code couples test infrastructure to shipping code

Adding `data-testid="table-container"` directly to `Table.jsx` introduces a test-only attribute into the production bundle. While this is an accepted and common pattern in many codebases (and is fine for most projects), some teams prefer to avoid test attributes in production HTML, either for cleanliness or to prevent external consumers from accidentally relying on them.

The specific use-case here — clicking outside to dismiss the context menu — could instead use a pre-existing structural selector (e.g., the `overflow-x-scroll` wrapper div) via `container.firstChild` or a role-based selector.

**Suggestion:**
```
If the team prefers to keep test attributes out of production, consider using the container reference from `render()` instead:
```js
// No data-testid needed
const { container } = await renderTable();
await user.click(container.firstElementChild); // outermost div
```
Alternatively, document a team-level decision to accept `data-testid` in production as a convention and move on — this is the simpler choice and widely practiced.
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 13]
```diff
@@ -0,0 +1,545 @@
...
+        });
+      });
+
+      expect(screen.getByText("Add Row")).toBeInTheDocument();
+      expect(screen.getByText("Delete Row")).toBeInTheDocument();
+    });
+
```
ivantha, line 155: **[LOW]** No tests for context menu keyboard dismissal (Escape key)

The context menu dismissal test only covers the click-outside path. Most accessible context menu implementations also close on `Escape`. If the component supports this (common for keyboard accessibility), it should be tested. If it does not, that may be an accessibility gap worth a comment or follow-up.

**Suggestion:**
```
Consider adding:
```js
it('closes the context menu when Escape is pressed', async () => {
  await renderTable();
  await user.pointer({
    target: screen.getByText('Name').closest('th'),
    keys: '[MouseRight]',
  });
  expect(screen.getByText('Add Column')).toBeInTheDocument();

  await user.keyboard('{Escape}');

  expect(screen.queryByText('Add Column')).not.toBeInTheDocument();
});
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 14]
```diff
@@ -0,0 +1,545 @@
...
+      await renderTable();
+      // Right-click on the "Name" header (display index 1); the handler strips
+      // the S.No. offset so the API receives col_params.index = 1 - 1 = 0.
+      await act(async () => {
+        await user.pointer({
+          target: screen.getByText("Name").closest("th"),
+          keys: "[MouseRight]",
+        });
+      });
+
```
ivantha, line 410: **[LOW]** No test for cancelling the Rename Column or Add Column dialog

The context menu action tests cover the happy path for both Rename Column and Add Column (typing a name and clicking OK), but there are no tests for clicking Cancel on the InputDialog. If the component calls `transformProject` even when the dialog is cancelled (a potential bug), no test would catch it.

**Suggestion:**
```
Add cancel tests:
```js
it('does not call transformProject when Rename Column dialog is cancelled', async () => {
  await renderTable();
  await user.pointer({ target: screen.getByText('Name').closest('th'), keys: '[MouseRight]' });
  await user.click(screen.getByText('Rename Column'));
  const dialogInput = screen.getByRole('textbox');
  await user.type(dialogInput, 'ShouldNotSave');
  await user.click(screen.getByRole('button', { name: /cancel/i }));
  expect(transformProject).not.toHaveBeenCalled();
});
```
```

File: dataloom-frontend/src/Components/__tests__/Table.test.jsx [hunk 15]
```diff
@@ -0,0 +1,545 @@
...
+// Object.freeze prevents accidental mutation of shared test fixtures that
+// would cause hard-to-debug cross-test pollution.
+const COLUMNS = Object.freeze(["Name", "Score"]);
+const ROWS = Object.freeze([Object.freeze(["Alice", 90]), Object.freeze(["Bob", 85])]);
+
+const mockContext = (overrides = {}) => {
+  useProjectContext.mockReturnValue({
+    columns: COLUMNS,
+    rows: ROWS,
```
ivantha, line 30: **[INFO]** Good use of `Object.freeze` on shared fixtures to prevent cross-test mutation

Using `Object.freeze` on `COLUMNS` and `ROWS` is a thoughtful defensive practice. Shared mutable fixtures are a common source of hard-to-diagnose test pollution, where one test mutates a fixture and causes a subsequent test to fail for seemingly unrelated reasons. Freezing at both the array and nested array level correctly prevents both `ROWS.push(...)` and `ROWS[0][0] = 'mutated'`.

**Suggestion:**
```
No change needed — this is a good pattern worth acknowledging and potentially propagating to other test files in the project.
```

dataloom-frontend/src/Components/__tests__/Table.test.jsx
ivantha, line 55: **[INFO]** Good use of `renderTable` and `mockContext` helpers for DRY test setup

The `mockContext` and `renderTable` helper functions centralize setup logic, accept overrides cleanly, and make individual tests concise and readable. The `overrides = {}` default parameter and spread pattern are idiomatic and allow individual tests to customize only what they need without duplicating full context objects.

**Suggestion:**
```
No change needed — this is a well-structured helper pattern. Consider extracting it to a shared `test-utils.jsx` if other component test files are added in the future.
```

dataloom-frontend/src/Components/__tests__/Table.test.jsx
```diff
@@ -0,0 +1,545 @@
...
+  });
+
+  // ── Rendering ──────────────────────────────────────────────────────────────
+
+  describe("rendering", () => {
+    it("renders S.No. column header", async () => {
+      await renderTable();
+      expect(screen.getByText("S.No.")).toBeInTheDocument();
+    });
+
```
ivantha, line 65: **[INFO]** Good inline code comments explaining non-obvious testing decisions

The test file includes numerous well-written comments that explain *why* a particular approach was chosen (e.g., why `container.querySelectorAll` is used instead of text-based queries for row count, why `data-testid` is preferred over Tailwind class selectors, why `userEvent.setup()` handles act internally). These comments significantly improve maintainability for future contributors.

**Suggestion:**
```
No change needed — this is a positive quality signal. The commenting style is appropriately targeted at non-obvious decisions without over-documenting obvious code.
```

dataloom-frontend/src/Components/Table.jsx
ivantha, line 226: **[LOW]** Non-breaking: `data-testid` attribute added to Table component's root element

**Previous behavior**: The outermost `<div>` in `Table.jsx` had `className="px-8 pt-3"` and an `onClick` handler but no `data-testid`.

**New behavior**: The element now also carries `data-testid="table-container"`.

**Who is affected**: No consumers are broken — adding a HTML attribute is purely additive. Snapshot tests or DOM-diffing tests that assert on the exact HTML structure of Table's root element could unexpectedly fail.

**Migration path**: If any existing snapshot tests cover `Table.jsx`, run `vitest --update-snapshots` to accept the diff.

**Suggestion:**
```
Check for existing snapshot tests (`*.snap` files) containing Table's rendered HTML and update them. This is a one-time `--update-snapshots` run.
```

ivantha:
This PR currently has merge conflicts with `main`. Please rebase onto latest `main` and resolve the conflicts before we can merge.

ivantha:

Issue #148, opened 2026-02-28
[Bug] Non-CSV file upload triggers unnecessary API request


vikas-cse123:
**Describe the bug**
When a non-CSV file is selected then  submitting the form still sends an API request to the server. The backend responds with a 400 error that says only CSV files are allowed. 
This leads to an unnecessary network request that could be avoided with client-side file type validation.

**Expected behavior**
No API request should be sent when a non-CSV file is chosen.

**Screenshots**
1. Non-CSV file selected. Form submission has started.
<img width="1594" height="649" alt="Image" src="https://github.com/user-attachments/assets/cd2e11f2-46ce-4b71-98a1-8dbeebd3b928" />

2. A needless API request is made 
<img width="860" height="246" alt="Image" src="https://github.com/user-attachments/assets/479af566-bd08-43df-8d0d-4b25845f87df" />

ivantha:
Closing: this is already fixed on `main`.

Submission runs client-side extension and size validation and returns with a toast before any network call, so an unsupported file never reaches the API.

Evidence: dataloom-frontend/src/Components/Homescreen.jsx:335-338 calls validateFile and returns early; dataloom-frontend/src/utils/fileUtils.ts:22-33 does the extension check against ACCEPTED_EXTENSIONS.

Thanks for the report. If you still see this on `main`, comment here and we will reopen.

PR #149, opened 2026-02-28
fix: prevent invalid file upload from triggering API request

vikas-cse123:
## Description

Prevents non-CSV files from triggering an API request in the Create Project modal.

Fixes #(issue number)
#148 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Screenshots 

1. File picker restricted to CSV files
image

2. Client-side validation blocks request and displays error toast
image

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/Homescreen.jsx
+++ b/dataloom-frontend/src/Components/Homescreen.jsx
@@ -95,6 +95,11 @@ const HomeScreen = () => {
       return;
     }
 
+    if (!fileUpload.name.endsWith(".csv")) {
+      showToast("Only CSV files are allowed.", "warning");
+      return;
+    }
+
     if (!projectName.trim()) {
       showToast("Project Name cannot be empty", "warning");
       return;
@@ -203,6 +208,7 @@ const HomeScreen = () => {
             <h2 className="text-2xl font-semibold text-gray-900 mb-4">Upload Dataset</h2>
             <input
               type="file"
+              accept=".csv"
               className="block w-full text-lg text-gray-900 border border-gray-300 rounded-md px-3 py-2 bg-white cursor-pointer focus:outline-none mb-4"
               onChange={handleFileUpload}
             />


dataloom-frontend/src/Components/Homescreen.jsx
```diff
@@ -95,6 +95,11 @@ const HomeScreen = () => {
       return;
     }
 
+    if (!fileUpload.name.endsWith(".csv")) {
+      showToast("Only CSV files are allowed.", "warning");
+      return;
+    }
```
ivantha: **[LOW]** No guard against fileUpload being null/undefined before accessing .name

The check `fileUpload.name.endsWith(".csv")` assumes `fileUpload` is a valid `File` object with a `.name` property. There appears to be an existing null-check guard just above (line ~95 of the original), but it's worth confirming that guard covers the case where `fileUpload` is `null` or `undefined`. If `fileUpload` is a non-null but unexpected object shape (e.g., an empty FileList entry), accessing `.name` could still throw a runtime TypeError.

**Suggestion:**
```
Confirm the existing guard at line 95 covers `null`/`undefined` for `fileUpload`. If it does, add a brief comment to indicate intent:
```js
// Guard against no file selected (handled above at line 95)
if (!fileUpload.name.toLowerCase().endsWith(".csv")) { ... }
```
If there is any path where `fileUpload` could be a non-null non-File object, add:
```js
if (!(fileUpload instanceof File) || !fileUpload.name.toLowerCase().endsWith(".csv")) { ... }
```
```

File: dataloom-frontend/src/Components/Homescreen.jsx [hunk 2]
```diff
@@ -203,6 +208,7 @@ const HomeScreen = () => {
             <h2 className="text-2xl font-semibold text-gray-900 mb-4">Upload Dataset</h2>
             <input
               type="file"
+              accept=".csv"
```
ivantha: **[INFO]** accept=".csv" is a UX hint, not a security control

The `accept=".csv"` attribute on the `<input type="file">` element is a welcome UX improvement — it filters the OS file picker dialog to show only CSV files by default. However, it is important to document and understand that `accept` can be trivially bypassed by the user selecting "All Files" in the OS dialog, or by drag-and-drop, or programmatically via test tooling.

This is correctly paired with the JavaScript guard, which provides a second layer. This is good defence-in-depth for the UI layer. The observation here is only to ensure the team understands neither check is a security boundary — only server-side validation provides real protection.

**Suggestion:**
```
No code change required. Recommended documentation practice: add a comment near the validation logic:
```js
// Client-side guard: prevents accidental non-CSV uploads.
// NOTE: Server-side validation is the authoritative enforcement.
if (!fileUpload.name.toLowerCase().endsWith(".csv")) { ... }
```
Also confirm with the backend team that the API endpoint validates file content/type independently.
```

File: dataloom-frontend/src/Components/Homescreen.jsx (hunk 1 above)
Comment (ivantha) [inline dataloom-frontend/src/Components/Homescreen.jsx]: **[INFO]** Positive: Early-return guard pattern is clean and consistent

The new validation block follows the existing early-return validation pattern already used throughout the function (checking `fileUpload`, `projectName`, etc.). This is idiomatic, readable, and consistent with the surrounding code. Good pattern usage.

**Suggestion:**
```
No change needed — acknowledge this as a positive pattern.
```

ivantha (commented):
## ◎ Argus Review

This PR adds client-side validation to prevent non-CSV files from triggering an API request in the Create Project modal. The change is directionally correct — combining a UI-level `accept` filter with a JavaScript guard — but the implementation has a real bug (case-insensitive extension handling), relies solely on the file extension rather than MIME type, and lacks automated test coverage. The fix should be strengthened before merging.

PR #156, opened 2026-03-01
feat(ui): replace Filter/Sort column text input with dropdown

dhruv1955:
## Summary
Replace manual **Column** text inputs with **dropdown selectors** 
populated from the current dataset columns in Filter and Sort forms.

## Why
- Prevents typos in column names
- Faster and more beginner-friendly
- Makes Filter and Sort consistent with each other

## Changes
- `FilterForm.jsx` — Column field replaced with dropdown
- `SortForm.jsx` — Column field replaced with dropdown
- Dropdown options sourced live from project context columns

## Validation
- `npm run lint` ✅
- `npm run test` ✅ (41/41)

### Filter Form
| Before | After |
|--------|-------|
| filter-before | filter-after |

### Sort Form
| Before | After |
|--------|-------|
| sort-before | sort-after |

## Notes
- No backend changes
- No API contract changes

### DIFF
--- a/dataloom-frontend/src/Components/common/ColumnSelect.jsx
+++ b/dataloom-frontend/src/Components/common/ColumnSelect.jsx
@@ -1,0 +1,53 @@
+import PropTypes from "prop-types";
+
+const ColumnSelect = ({
+  id,
+  label,
+  name = "column",
+  value,
+  onChange,
+  columns,
+  required = false,
+}) => {
+  return (
+    <>
+      <label htmlFor={id} className="block mb-1 text-sm font-medium text-gray-700">
+        {label}
+      </label>
+      <select
+        id={id}
+        name={name}
+        value={value}
+        onChange={onChange}
+        disabled={!columns || columns.length === 0}
+        className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none disabled:bg-gray-100 disabled:text-gray-500"
+        required={required}
+      >
+        {!columns || columns.length === 0 ? (
+          <option value="">No columns available — load a dataset first</option>
+        ) : (
+          <>
+            <option value="">Select column…</option>
+            {columns.map((columnName, index) => (
+              <option key={`${index}-${columnName}`} value={columnName}>
+                {columnName}
+              </option>
+            ))}
+          </>
+        )}
+      </select>
+    </>
+  );
+};
+
+ColumnSelect.propTypes = {
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  name: PropTypes.string,
+  value: PropTypes.string.isRequired,
+  onChange: PropTypes.func.isRequired,
+  columns: PropTypes.arrayOf(PropTypes.string).isRequired,
+  required: PropTypes.bool,
+};
+
+export default ColumnSelect;
--- a/dataloom-frontend/src/Components/forms/FilterForm.jsx
+++ b/dataloom-frontend/src/Components/forms/FilterForm.jsx
@@ -2,11 +2,15 @@ import { useState } from "react";
 import PropTypes from "prop-types";
 import { transformProject } from "../../api";
 import { FILTER } from "../../constants/operationTypes";
+import { useProjectContext } from "../../context/ProjectContext";
 import TransformResultPreview from "./TransformResultPreview";
 import useError from "../../hooks/useError";
 import FormErrorAlert from "../common/FormErrorAlert";
+import ColumnSelect from "../common/ColumnSelect";
 
 const FilterForm = ({ projectId, onClose }) => {
+  const { columns } = useProjectContext();
+  const safeColumns = columns ?? [];
   const [filterParams, setFilterParams] = useState({
     column: "",
     condition: "=",
@@ -49,19 +53,25 @@ const FilterForm = ({ projectId, onClose }) => {
         <h3 className="font-semibold text-gray-900 mb-2">Filter Dataset</h3>
         <div className="flex flex-wrap mb-4">
           <div className="w-full sm:w-1/3 mb-2">
-            <label className="block mb-1 text-sm font-medium text-gray-700">Column:</label>
-            <input
-              type="text"
+            <ColumnSelect
+              id="filter-column"
+              label="Column:"
               name="column"
               value={filterParams.column}
               onChange={handleInputChange}
-              className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+              columns={safeColumns}
               required
             />
           </div>
           <div className="w-full sm:w-1/3 mb-2 pl-2">
-            <label className="block mb-1 text-sm font-medium text-gray-700">Condition:</label>
+            <label
+              htmlFor="filter-condition"
+              className="block mb-1 text-sm font-medium text-gray-700"
+            >
+              Condition:
+            </label>
             <select
+              id="filter-condition"
               name="condition"
               value={filterParams.condition}
               onChange={handleInputChange}
@@ -78,8 +88,11 @@ const FilterForm = ({ projectId, onClose }) => {
             </select>
           </div>
           <div className="w-full sm:w-1/3 mb-2 pl-2">
-            <label className="block mb-1 text-sm font-medium text-gray-700">Value:</label>
+            <label htmlFor="filter-value" className="block mb-1 text-sm font-medium text-gray-700">
+              Value:
+            </label>
             <input
+              id="filter-value"
               type="text"
               name="value"
               value={filterParams.value}
@@ -93,7 +106,7 @@ const FilterForm = ({ projectId, onClose }) => {
           <button
             type="submit"
             className="bg-blue-500 hover:bg-blue-600 text-white px-4 py-2 rounded-md font-medium transition-colors duration-150"
-            disabled={loading}
+            disabled={loading || safeColumns.length === 0}
           >
             Apply Filter
           </button>
--- a/dataloom-frontend/src/Components/forms/SortForm.jsx
+++ b/dataloom-frontend/src/Components/forms/SortForm.jsx
@@ -2,11 +2,15 @@ import { useState } from "react";
 import PropTypes from "prop-types";
 import { transformProject } from "../../api";
 import { SORT } from "../../constants/operationTypes";
+import { useProjectContext } from "../../context/ProjectContext";
 import TransformResultPreview from "./TransformResultPreview";
 import useError from "../../hooks/useError";
 import FormErrorAlert from "../common/FormErrorAlert";
+import ColumnSelect from "../common/ColumnSelect";
 
 const SortForm = ({ projectId, onClose }) => {
+  const { columns } = useProjectContext();
+  const safeColumns = columns ?? [];
   const [column, setColumn] = useState("");
   const [ascending, setAscending] = useState(true);
   const [result, setResult] = useState(null);
@@ -42,18 +46,22 @@ const SortForm = ({ projectId, onClose }) => {
         <h3 className="font-semibold text-gray-900 mb-2">Sort Dataset</h3>
         <div className="flex flex-wrap mb-4">
           <div className="w-full sm:w-1/2 mb-2">
-            <label className="block mb-1 text-sm font-medium text-gray-700">Column:</label>
-            <input
-              type="text"
+            <ColumnSelect
+              id="sort-column"
+              label="Column:"
+              name="column"
               value={column}
               onChange={(e) => setColumn(e.target.value)}
-              className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+              columns={safeColumns}
               required
             />
           </div>
           <div className="w-full sm:w-1/2 mb-2 pl-2">
-            <label className="block mb-1 text-sm font-medium text-gray-700">Order:</label>
+            <label htmlFor="sort-order" className="block mb-1 text-sm font-medium text-gray-700">
+              Order:
+            </label>
             <select
+              id="sort-order"
               value={ascending}
               onChange={(e) => setAscending(e.target.value === "true")}
               className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
@@ -67,9 +75,9 @@ const SortForm = ({ projectId, onClose }) => {
           <button
             type="submit"
             className="bg-blue-500 hover:bg-blue-600 text-white px-4 py-2 rounded-md font-medium transition-colors duration-150"
-            disabled={loading}
+            disabled={loading || safeColumns.length === 0}
           >
-            Submit
+            Apply Sort
           </button>
           <button
             type="button"
--- a/dataloom-frontend/src/Components/common/__tests__/ColumnSelect.test.jsx
+++ b/dataloom-frontend/src/Components/common/__tests__/ColumnSelect.test.jsx
@@ -1,0 +1,67 @@
+import { render, screen, fireEvent } from "@testing-library/react";
+import { vi, beforeEach, expect, describe, test } from "vitest";
+import ColumnSelect from "../ColumnSelect";
+
+describe("ColumnSelect", () => {
+  const defaultProps = {
+    id: "test-select",
+    label: "Test Label",
+    value: "",
+    onChange: vi.fn(),
+    columns: ["col1", "col2", "col3"],
+  };
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  test("renders placeholder and disabled select when columns is empty array", () => {
+    render(<ColumnSelect {...defaultProps} columns={[]} />);
+
+    expect(screen.getByText("No columns available — load a dataset first")).toBeInTheDocument();
+    expect(screen.getByRole("combobox")).toBeDisabled();
+  });
+
+  test("renders all column names as options when columns provided", () => {
+    render(<ColumnSelect {...defaultProps} />);
+
+    expect(screen.getByText("Select column…")).toBeInTheDocument();
+    expect(screen.getByText("col1")).toBeInTheDocument();
+    expect(screen.getByText("col2")).toBeInTheDocument();
+    expect(screen.getByText("col3")).toBeInTheDocument();
+  });
+
+  test("select is disabled when columns is null", () => {
+    render(<ColumnSelect {...defaultProps} columns={null} />);
+
+    expect(screen.getByRole("combobox")).toBeDisabled();
+  });
+
+  test("calls onChange when option selected", () => {
+    const mockOnChange = vi.fn();
+    render(<ColumnSelect {...defaultProps} onChange={mockOnChange} />);
+
+    const select = screen.getByRole("combobox");
+    fireEvent.change(select, { target: { value: "col2" } });
+
+    expect(mockOnChange).toHaveBeenCalledTimes(1);
+  });
+
+  test("required prop is forwarded to select element", () => {
+    render(<ColumnSelect {...defaultProps} required={true} />);
+
+    expect(screen.getByRole("combobox")).toBeRequired();
+  });
+
+  test("uses default name when not provided", () => {
+    const { container } = render(<ColumnSelect {...defaultProps} />);
+
+    expect(container.querySelector("select")).toHaveAttribute("name", "column");
+  });
+
+  test("uses custom name when provided", () => {
+    const { container } = render(<ColumnSelect {...defaultProps} name="custom-name" />);
+
+    expect(container.querySelector("select")).toHaveAttribute("name", "custom-name");
+  });
+});


ivantha (commented):
## ◎ Argus Review

This PR is a well-scoped, user-experience improvement that replaces free-text column inputs with dropdowns sourced from project context in `FilterForm.jsx` and `SortForm.jsx`. The core approach is sound and the two files are implemented consistently. However, there is a meaningful runtime crash risk if `columns` is ever `null` or `undefined` from context (e.g., during loading or on error), and several accessibility and UX gaps deserve attention before merging.

dataloom-frontend/src/Components/forms/FilterForm.jsx
```diff
@@ -49,14 +51,20 @@ const FilterForm = ({ projectId, onClose }) => {
...
-            />
+            >
+              <option value="">Select column...</option>
+              {columns.map((column) => (
+                <option key={column} value={column}>
+                  {column}
+                </option>
+              ))}
+            </select>
           </div>
```
ivantha: **[MEDIUM]** `<label>` not programmatically associated with `<select>` (accessibility)

The `<label>` element reads `Column:` but has no `htmlFor` attribute, and the `<select>` has no matching `id`. Without this association, assistive technologies (screen readers, voice control) cannot correctly link the label to its control.

This was a pre-existing issue on the original `<input>`, but the PR is an ideal moment to fix it since the element is already being replaced. The WCAG 2.1 success criterion 1.3.1 (Info and Relationships) requires this pairing.

**Suggestion:**
```
```jsx
<label htmlFor="filter-column" className="block mb-1 text-sm font-medium text-gray-700">
  Column:
</label>
<select
  id="filter-column"
  name="column"
  value={filterParams.column}
  onChange={handleInputChange}
  className="..."
  required
>
```
```

File: dataloom-frontend/src/Components/forms/SortForm.jsx [hunk 2]
```diff
@@ -42,13 +44,19 @@ const SortForm = ({ projectId, onClose }) => {
...
-            />
+            >
+              <option value="">Select column...</option>
+              {columns.map((columnName) => (
+                <option key={columnName} value={columnName}>
+                  {columnName}
+                </option>
+              ))}
+            </select>
           </div>
```
ivantha: **[MEDIUM]** `<label>` not programmatically associated with `<select>` (accessibility, SortForm)

Same accessibility gap as in `FilterForm.jsx`. The `<label>Column:</label>` and `<select>` are visually co-located but not programmatically linked via `htmlFor`/`id`.

**Suggestion:**
```
```jsx
<label htmlFor="sort-column" className="block mb-1 text-sm font-medium text-gray-700">
  Column:
</label>
<select
  id="sort-column"
  value={column}
  onChange={(e) => setColumn(e.target.value)}
  className="..."
  required
>
```
```

File: dataloom-frontend/src/Components/forms/FilterForm.jsx [hunk 3]
```diff
@@ -49,14 +51,20 @@ const FilterForm = ({ projectId, onClose }) => {
...
               name="column"
               value={filterParams.column}
               onChange={handleInputChange}
               className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
               required
-            />
+            >
+              <option value="">Select column...</option>
+              {columns.map((column) => (
+                <option key={column} value={column}>
```
ivantha: **[MEDIUM]** No empty-state message when dataset has no columns

When `columns` is an empty array (e.g., before data is loaded, or if the dataset is genuinely empty), the dropdown renders only the `Select column…` placeholder. The user has no way to distinguish between "data is still loading", "you must load a dataset first", or "this dataset has no columns". This will be confusing, especially for the beginner audience the PR description explicitly targets.

**Suggestion:**
```
Consider disabling the select and providing an explicit message when no columns are available:
```jsx
<select
  name="column"
  value={filterParams.column}
  onChange={handleInputChange}
  className="..."
  required
  disabled={columns.length === 0}
>
  {columns.length === 0 ? (
    <option value="">No columns available — load a dataset first</option>
  ) : (
    <>
      <option value="">Select column…</option>
      {columns.map((column) => (
        <option key={column} value={column}>{column}</option>
      ))}
    </>
  )}
</select>
```
```

File: dataloom-frontend/src/Components/forms/FilterForm.jsx (hunk 1 above)
Comment (ivantha) [inline dataloom-frontend/src/Components/forms/FilterForm.jsx]: **[MEDIUM]** Previously entered column value becomes silently invisible after switch

If a user had previously submitted a filter and the form is reopened with a pre-populated `column` value (or if any parent component passes an initial value that doesn't match any option in `columns`), the `<select>` will silently fall back to the placeholder option. Unlike a text `<input>`, a `<select>` cannot display an arbitrary string that isn't one of its `<option>` values.

This is subtle: if `filterParams.column` holds a column name not present in the current `columns` array, the controlled value appears as the empty-string option, which could mislead the user into thinking no column was selected, and a new submission would send an empty string.

**Suggestion:**
```
Validate or reset the pre-populated value against the current columns list when the component mounts or when `columns` changes:
```jsx
useEffect(() => {
  if (filterParams.column && !columns.includes(filterParams.column)) {
    setFilterParams((prev) => ({ ...prev, column: "" }));
  }
}, [columns]);
```
Alternatively, if there is no pre-population scenario, this can be documented as a non-issue.
```

File: dataloom-frontend/src/Components/forms/SortForm.jsx [hunk 4]
```diff
@@ -42,13 +44,19 @@ const SortForm = ({ projectId, onClose }) => {
...
             <label className="block mb-1 text-sm font-medium text-gray-700">Column:</label>
-            <input
-              type="text"
+            <select
               value={column}
               onChange={(e) => setColumn(e.target.value)}
               className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
               required
-            />
+            >
```
ivantha: **[LOW]** SortForm `<select>` is missing a `name` attribute

The `<select>` in `SortForm` does not carry a `name` attribute, unlike the `<select>` in `FilterForm` which has `name="column"`. While the current implementation manages state via a direct `onChange` setter (`setColumn`) and doesn't rely on form serialisation, the missing `name` is an inconsistency that could confuse future maintainers, and it would silently omit the field if anyone later tries to read `FormData` from this form.

**Suggestion:**
```
Add `name="column"` to the SortForm `<select>` for consistency and forwards-compatibility:
```jsx
<select
  name="column"
  value={column}
  onChange={(e) => setColumn(e.target.value)}
  ...
>
```
```

File: dataloom-frontend/src/Components/forms/FilterForm.jsx [hunk 5]
```diff
@@ -49,14 +51,20 @@ const FilterForm = ({ projectId, onClose }) => {
...
+                  {column}
+                </option>
+              ))}
+            </select>
```
ivantha: **[LOW]** Column name used as React `key` — fragile if column names are non-unique or special characters

Both dropdowns use the column name string as the React `key` prop (`key={column}`, `key={columnName}`). This is generally fine for datasets with unique column names, but two edge cases are worth considering:
1. **Duplicate column names** — Some data formats (e.g., CSVs with repeated headers) can produce duplicate column names. Duplicate keys cause silent React reconciliation bugs.
2. **Special characters / whitespace** — While not a React issue per se, very long or exotic column names (newlines, Unicode control characters) could cause subtle rendering issues in some browsers.

**Suggestion:**
```
If columns are guaranteed unique (typical for relational datasets), the current approach is acceptable. If not, use an index-based fallback for the key while keeping the value as the name:
```jsx
{columns.map((column, index) => (
  <option key={`${index}-${column}`} value={column}>
    {column}
  </option>
))}
```
```

File: dataloom-frontend/src/Components/forms/FilterForm.jsx [hunk 6]
```diff
@@ -1,11 +1,13 @@
 import { useState } from "react";
 import PropTypes from "prop-types";
 import { transformProject } from "../../api";
+import { useProjectContext } from "../../context/ProjectContext";
 import TransformResultPreview from "./TransformResultPreview";
 import useError from "../../hooks/useError";
 import FormErrorAlert from "../common/FormErrorAlert";
 
 const FilterForm = ({ projectId, onClose }) => {
+  const { columns } = useProjectContext();
```
ivantha: **[LOW]** Columns from context may be stale if the dataset changes while the form is open

The component reads `columns` from context once at render time and re-reads it only when the context value changes. If the underlying dataset is mutated (e.g., a column is renamed or dropped) while the Filter/Sort form is open, the dropdown could show column names that no longer exist, leading to a server-side error on submission.

This is a low-risk concern for most UX flows (users typically don't mutate the dataset while a modal form is open), but it's worth flagging as an architectural consideration.

**Suggestion:**
```
Consider whether the context is kept live (reactive) or if it represents a snapshot. If snapshots are used, document this limitation. If live updates are expected, ensure the context provider invalidates and re-fetches columns after mutations. No code change is strictly required if the UX flow makes concurrent mutations impossible.
```

dataloom-frontend/src/Components/forms/FilterForm.jsx
ivantha: **[INFO]** Variable naming collision: outer `column` prop shadows inner `column` map variable

In `FilterForm.jsx`, the `.map()` callback uses `(column) => ...` as the parameter name. While the outer scope also references a `column` prop on `filterParams` (as `filterParams.column`), there is no actual variable named `column` in the outer scope — so there is no true shadowing. However, this could cause confusion for readers who see `column` used both in the `.map()` and expect a broader-scope `column` variable.

In `SortForm.jsx`, this was correctly avoided by using `columnName` as the map parameter, since the outer scope *does* have a `column` state variable. This inconsistency between the two files could trip up future contributors.

**Suggestion:**
```
Rename the map parameter in `FilterForm.jsx` to match `SortForm.jsx` for consistency:
```jsx
{columns.map((columnName) => (
  <option key={columnName} value={columnName}>
    {columnName}
  </option>
))}
```
```

File: dataloom-frontend/src/Components/forms/FilterForm.jsx (hunk 1 above)
Comment (ivantha) [inline dataloom-frontend/src/Components/forms/FilterForm.jsx]: **[INFO]** Positive: Dropdown options block is a good candidate for a shared component

The pattern of rendering a `<select>` from `columns` context is now used identically in two places. If a third consumer is added (e.g., a GroupBy form, an Aggregate form), this pattern should be extracted into a shared `<ColumnSelect />` component to avoid copy-paste drift.

This is not a blocker for this PR, but flagging it now so the team can plan ahead.

**Suggestion:**
```
Consider creating a reusable component in `src/Components/common/ColumnSelect.jsx`:
```jsx
const ColumnSelect = ({ value, onChange, name, id, required }) => {
  const { columns = [] } = useProjectContext();
  return (
    <select
      id={id}
      name={name}
      value={value}
      onChange={onChange}
      required={required}
      disabled={columns.length === 0}
    >
      <option value="">
        {columns.length === 0 ? 'No columns available' : 'Select column…'}
      </option>
      {columns.map((col) => (
        <option key={col} value={col}>{col}</option>
      ))}
    </select>
  );
};
```
Both `FilterForm` and `SortForm` would then use `<ColumnSelect ... />`, keeping all column-display logic (empty state, guard, accessibility) in one place.
```

**Comment by dhruv1955** [comment]
@ivantha  Thanks for the thorough review! Addressed all feedback:

- Added `columns = []` default guard in both forms (HIGH)
- Added `htmlFor`/`id` associations on labels in both forms (MEDIUM)
- Added disabled state + "No columns available" empty message (MEDIUM)
- Added `name="column"` to SortForm select (LOW)
- Renamed map param to `columnName` in FilterForm (INFO)
- Extracted shared `<ColumnSelect />` component used by both forms (INFO)

`npm run lint` ✅
`npm run test` ✅ (41/41)

**Comment by dhruv1955** [comment]
Regarding the test coverage note — the existing 41 tests cover the 
happy path. I can add unit tests for:
- Rendering when `columns` is undefined
- Rendering when `columns` is empty array

Let me know if you'd like me to add these before merging.

File: dataloom-frontend/src/Components/forms/FilterForm.jsx [hunk 7]
```diff
@@ -30,7 +32,7 @@ const FilterForm = ({ projectId, onClose }) => {
     clearError();
     try {
       const response = await transformProject(projectId, {
-        operation_type: FILTER,
+        operation_type: "filter",
```
ivantha: **[MEDIUM]** Magic strings replace named constants for operation_type

The PR removes `import { FILTER } from '../../constants/operationTypes'` and replaces the constant reference with the inline string `"filter"`. The same pattern appears in `SortForm.jsx` with `"sort"`. While functionally equivalent (assuming the constants held those exact strings), this change:

1. **Breaks the single-source-of-truth principle** — if `FILTER`/`SORT` ever need to change their values, the constants file update would silently fail to update these files.
2. **Creates inconsistency** — other parts of the codebase may still use the constants, producing two different authoring styles for the same concept.
3. **Risks a silent bug** — if the constant values were ever something other than lowercase strings (e.g., `"FILTER"` or `"filter_op"`), the hardcoded string would silently diverge.

The constants file should remain the authoritative source for these string values.

**Suggestion:**
```
Revert to using the constants:
```jsx
import { FILTER } from '../../constants/operationTypes';
// ...
operation_type: FILTER,
```
If the intent is to remove the constants module entirely, that should be a separate, intentional PR with explicit deprecation tracking.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This is a clean, well-motivated UI improvement PR that introduces a reusable `ColumnSelect` component and correctly integrates it into both `FilterForm` and `SortForm`. The component design is solid — good PropTypes, accessible label/input linkage, and graceful empty-state handling. The main concerns are: (1) the replacement of `FILTER`/`SORT` named constants with magic strings, which is a maintainability regression inconsistent with having a constants module; (2) the use of the deprecated `defaultProps` API which will break in React 19; and (3) the absence of any new unit tests for the `ColumnSelect` component despite it being new, standalone logic. No breaking changes to external APIs or contracts are introduced.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes detected — suggestedSemver: minor: **Previous behavior**: Column inputs in FilterForm and SortForm were free-text `<input type="text">` elements allowing arbitrary string entry.

**New behavior**: Column inputs are `<select>` dropdowns populated from `ProjectContext`. Users can only select from existing dataset columns.

**Who is affected**: End users who previously relied on typing custom/non-existent column names (e.g., for testing or speculative operations). No external API, exported function, or backend contract changes.

**Migration path**: No code migration required. End users must select a column from the dropdown instead of typing it.

The `FILTER`/`SORT` constants are removed from the imports of `FilterForm` and `SortForm` but are not removed from their source module (`constants/operationTypes`), so other consumers remain unaffected.


### PR Description Issues

- **[LOW]** PR description omits the removal of FILTER/SORT constant usage: The description lists only the UI changes (dropdown replacement, live column sourcing) but does not mention that `FILTER` and `SORT` named constants are removed from the imports of both forms and replaced with inline magic strings `"filter"` and `"sort"`. This is a meaningful maintainability change that reviewers should be aware of, especially if the codebase has a convention of using the constants file.
  > **Suggestion:** Add a bullet to the **Changes** section:
> - `FilterForm.jsx` and `SortForm.jsx` — replaced `FILTER`/`SORT` constant references with inline string literals `"filter"`/`"sort"` (see note: consider reverting to constants for consistency)
- **[LOW]** PR description does not mention the submit button behavioral change: Both forms now disable the submit button when `columns.length === 0`, which is a behavioral change beyond the column input replacement. A user with no dataset loaded will see a disabled submit button — this is new behavior that the description does not call out.
  > **Suggestion:** Add to the **Changes** section:
> - Submit button in both forms is now disabled when no dataset columns are available, preventing empty submissions
- **[INFO]** PR description is well-structured with strong motivation and visual validation: The description clearly articulates the **why** (typo prevention, UX improvement, consistency), the **what** (three files changed, live column sourcing), and includes before/after screenshots for both forms. The validation section (lint + test counts) and explicit "no backend changes" note are good practices that reduce reviewer uncertainty.
  > **Suggestion:** No critical improvement needed. Optionally add a **Breaking changes: None** subsection for completeness, since reviewers often scan for this explicitly.

File: dataloom-frontend/src/Components/common/ColumnSelect.jsx [hunk 8]
```diff
@@ -0,0 +1,50 @@
...
+  label: PropTypes.string.isRequired,
+  name: PropTypes.string,
+  value: PropTypes.string.isRequired,
+  onChange: PropTypes.func.isRequired,
+  columns: PropTypes.arrayOf(PropTypes.string).isRequired,
+  required: PropTypes.bool,
+};
+
+ColumnSelect.defaultProps = {
+  name: "column",
```
ivantha: **[MEDIUM]** Use of deprecated `defaultProps` API

`ColumnSelect.defaultProps` was deprecated in React 18.3 and is scheduled for removal in a future major version. React now recommends using ES6 default parameter values directly in the function signature. Using `defaultProps` generates deprecation warnings in development builds with React 18.3+, and will stop working entirely in React 19+.

This is especially relevant for functional components — `defaultProps` was primarily kept for class component compatibility.

**Suggestion:**
```
Replace `defaultProps` with default parameter values:
```jsx
const ColumnSelect = ({
  id,
  label,
  name = "column",
  value,
  onChange,
  columns,
  required = false,
}) => {
  // ...
};

// Remove the ColumnSelect.defaultProps block entirely
```
```

File: dataloom-frontend/src/Components/common/ColumnSelect.jsx [hunk 9]
```diff
@@ -0,0 +1,50 @@
...
+  required: false,
+};
+
+export default ColumnSelect;
```
ivantha, line 53: **[MEDIUM]** No tests added for new ColumnSelect component

The PR introduces a new reusable `ColumnSelect` component but does not include any unit tests. The PR description states "41/41" tests pass, which implies only pre-existing tests were run — no new coverage was added for the new component or its integration into `FilterForm`/`SortForm`.

Key behaviors that should be tested:
- Renders placeholder option when `columns` is empty
- Renders all column names as `<option>` elements
- Disables the `<select>` when `columns.length === 0`
- Calls `onChange` correctly when an option is selected
- `required` prop is forwarded to the underlying `<select>`
- Integration: Filter/Sort submit button is disabled when columns are empty

**Suggestion:**
```
Add a test file `ColumnSelect.test.jsx`:
```jsx
import { render, screen } from '@testing-library/react';
import ColumnSelect from './ColumnSelect';

it('shows empty-state message when no columns', () => {
  render(<ColumnSelect id="c" label="Col" value="" onChange={() => {}} columns={[]} />);
  expect(screen.getByRole('combobox')).toBeDisabled();
  expect(screen.getByText(/No columns available/i)).toBeInTheDocument();
});

it('renders column options', () => {
  render(<ColumnSelect id="c" label="Col" value="" onChange={() => {}} columns={['age', 'name']} />);
  expect(screen.getByRole('option', { name: 'age' })).toBeInTheDocument();
  expect(screen.getByRole('option', { name: 'name' })).toBeInTheDocument();
});
```
```

File: dataloom-frontend/src/Components/common/ColumnSelect.jsx [hunk 10]
```diff
@@ -0,0 +1,50 @@
...
+        {columns.length === 0 ? (
+          <option value="">No columns available — load a dataset first</option>
+        ) : (
+          <>
+            <option value="">Select column…</option>
+            {columns.map((columnName) => (
+              <option key={columnName} value={columnName}>
+                {columnName}
+              </option>
+            ))}
```
ivantha, line 35: **[LOW]** Non-unique column names cause duplicate React keys

The `key={columnName}` prop assumes all column names in the dataset are unique. Some datasets (e.g., CSVs with duplicate headers, or DataFrames with repeated column names) may contain duplicate column names. When duplicate keys appear, React will:

1. Log a warning in development mode
2. Potentially render incorrectly (only one of the duplicates may be committed to the DOM)

This is a subtle data-integrity risk that depends entirely on upstream data quality.

**Suggestion:**
```
Use a composite key combining index and column name to guarantee uniqueness:
```jsx
{columns.map((columnName, index) => (
  <option key={`${index}-${columnName}`} value={columnName}>
    {columnName}
  </option>
))}
```
```

File: dataloom-frontend/src/Components/forms/FilterForm.jsx [hunk 11]
```diff
@@ -1,12 +1,14 @@
...
-import { FILTER } from "../../constants/operationTypes";
+import { useProjectContext } from "../../context/ProjectContext";
 import TransformResultPreview from "./TransformResultPreview";
 import useError from "../../hooks/useError";
 import FormErrorAlert from "../common/FormErrorAlert";
+import ColumnSelect from "../common/ColumnSelect";
 
 const FilterForm = ({ projectId, onClose }) => {
+  const { columns = [] } = useProjectContext();
   const [filterParams, setFilterParams] = useState({
```
ivantha, line 14: **[LOW]** Context `columns` may be `null`, bypassing the default parameter

The destructuring `const { columns = [] } = useProjectContext()` only applies the `[]` default when `columns` is `undefined`. If the context provides `columns: null` (e.g., before data loads, or as an explicit reset state), the default won't kick in and `columns` will be `null`.

This would cause `columns.length === 0` in `ColumnSelect` to throw a `TypeError: Cannot read properties of null (reading 'length')`, crashing both forms.

The same issue exists in `SortForm.jsx`.

**Suggestion:**
```
Use a nullish coalescing guard to handle both `null` and `undefined`:
```jsx
const { columns } = useProjectContext();
const safeColumns = columns ?? [];
// then pass safeColumns to ColumnSelect
```
Or alternatively, normalise the value in `ColumnSelect` itself:
```jsx
const safeColumns = columns ?? [];
```
```

File: dataloom-frontend/src/Components/forms/SortForm.jsx [hunk 12]
```diff
@@ -42,18 +44,22 @@ const SortForm = ({ projectId, onClose }) => {
...
-            <input
-              type="text"
+            <ColumnSelect
+              id="sort-column"
+              label="Column:"
+              name="column"
               value={column}
               onChange={(e) => setColumn(e.target.value)}
-              className="border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
+              columns={columns}
```
ivantha: **[LOW]** Redundant explicit `name` prop on ColumnSelect in SortForm

`name="column"` is passed explicitly to `ColumnSelect` in `SortForm.jsx`, but `"column"` is already the default value defined in `ColumnSelect.defaultProps`. This is a minor redundancy that adds no value and could mislead future readers into thinking the default is something other than `"column"`.

Note: The same redundancy exists in `FilterForm.jsx` but is less harmful there because `FilterForm` uses the controlled `handleInputChange` handler which specifically reads `e.target.name`. Removing it there would require additional testing.

**Suggestion:**
```
Remove the explicit `name` prop from `SortForm`'s usage since the value matches the default:
```jsx
<ColumnSelect
  id="sort-column"
  label="Column:"
  value={column}
  onChange={(e) => setColumn(e.target.value)}
  columns={columns}
  required
/>
```
```

File: dataloom-frontend/src/Components/forms/SortForm.jsx [hunk 13]
```diff
@@ -67,7 +73,7 @@ const SortForm = ({ projectId, onClose }) => {
           <button
             type="submit"
             className="bg-blue-500 hover:bg-blue-600 text-white px-4 py-2 rounded-md font-medium transition-colors duration-150"
```
ivantha, line 77: **[LOW]** Inconsistent submit button label: "Submit" vs "Apply Filter"

The Filter form uses the descriptive label `"Apply Filter"` while the Sort form continues to use the generic `"Submit"`. While this is a pre-existing issue not introduced by this PR, the PR touches the button's `disabled` logic and was an opportunity to align the labels. A consistent, action-specific label (e.g., `"Apply Sort"`) improves usability and accessibility.

**Suggestion:**
```
Update the Sort form's button label to be consistent with the Filter form:
```jsx
<button
  type="submit"
  className="bg-blue-500 hover:bg-blue-600 text-white px-4 py-2 rounded-md font-medium transition-colors duration-150"
  disabled={loading || columns.length === 0}
>
  Apply Sort
</button>
```
```

File: dataloom-frontend/src/Components/common/ColumnSelect.jsx [hunk 14]
```diff
@@ -0,0 +1,50 @@
+import PropTypes from "prop-types";
+
+const ColumnSelect = ({ id, label, name, value, onChange, columns, required }) => {
+  return (
+    <>
+      <label htmlFor={id} className="block mb-1 text-sm font-medium text-gray-700">
```
ivantha, line 14: **[INFO]** Fragment root element limits parent styling flexibility

The component returns a React fragment (`<>...</>`). This works correctly in the current usage since the consuming parent divs provide all necessary layout wrappers. However, if a future consumer wants to add a `className` or other props to the outermost wrapper of `ColumnSelect` (e.g., for spacing overrides), they cannot do so without wrapping the component themselves.

A `<div>` wrapper with an optional `className` prop would make the component more composable.

**Suggestion:**
```
Add an optional `className` prop for wrapper customisation:
```jsx
const ColumnSelect = ({ id, label, name = "column", value, onChange, columns, required = false, className = "" }) => {
  return (
    <div className={className}>
      <label htmlFor={id} className="block mb-1 text-sm font-medium text-gray-700">
        {label}
      </label>
      {/* ... */}
    </div>
  );
};
```
This is optional and the current approach is acceptable for the current use case.
```

File: dataloom-frontend/src/Components/common/ColumnSelect.jsx [hunk 15]
```diff
@@ -0,0 +1,50 @@
...
+        {label}
+      </label>
+      <select
+        id={id}
+        name={name}
+        value={value}
+        onChange={onChange}
+        disabled={columns.length === 0}
```
ivantha: **[INFO]** Disabled select provides no tooltip or contextual guidance

When `columns.length === 0`, the `<select>` is disabled and shows `"No columns available — load a dataset first"`. While helpful, users may not notice the placeholder text. The submit button is also disabled but there's no tooltip or `aria-describedby` explaining *why* it is disabled. Screen reader users in particular may not understand the cause without additional cues.

**Suggestion:**
```
Add a `title` tooltip to the disabled button so users understand why submission is blocked:
```jsx
<button
  type="submit"
  title={columns.length === 0 ? "Load a dataset to enable filtering" : undefined}
  disabled={loading || columns.length === 0}
  className="..."
>
  Apply Filter
</button>
```
```

File: dataloom-frontend/src/Components/common/ColumnSelect.jsx (hunk 9 above)
Comment (ivantha) [inline dataloom-frontend/src/Components/common/ColumnSelect.jsx:53]: **[INFO]** Positive: Well-structured reusable component with good PropTypes

The extraction of `ColumnSelect` as a shared component follows good component design principles:
- Correct `PropTypes` definitions with `isRequired` constraints
- Consistent disabled state handling in both the `<select>` and its placeholder option
- Accessible `htmlFor`/`id` linkage between label and control
- Handles edge cases (empty columns array) gracefully with a descriptive placeholder
- The PR also correctly adds `htmlFor`/`id` pairs to the pre-existing `Condition`, `Value`, and `Order` fields — a nice accessibility improvement beyond the stated scope.

**Suggestion:**
```
No change needed — this is a positive observation. Consider adding a JSDoc comment above the component to document its expected use case for future contributors.
```

**Comment by dhruv1955** [comment]
@ivantha Addressed all feedback from the second review:

- Reverted `"filter"`/`"sort"` back to `FILTER`/`SORT` named constants
- Replaced deprecated `defaultProps` with ES6 default parameters in `ColumnSelect`
- Added `ColumnSelect.test.jsx` with 7 tests covering empty state, null columns, options rendering, onChange, required prop, and default/custom name - **55/55 passing**
- Changed null guard to `!columns || columns.length === 0` to handle both `null` and `undefined`
- Using composite key `${index}-${columnName}` for duplicate-safe option rendering
- Removed redundant explicit `name="column"` from SortForm usage
- Changed "Submit" → "Apply Sort" for consistency with FilterForm

**Comment by ivantha** [comment]

```
git fetch origin
git rebase origin/main
git push --force-with-lease
```

One small note: the `disabled` button condition uses `columns.length` directly rather than `safeColumns.length` — worth guarding in case `columns` is `null` when the context hasn't loaded yet. Otherwise the PR is clean — reusable `ColumnSelect` component, good tests, and consistent UX improvement across Filter and Sort forms.

dhruv1955:
@ivantha Rebased onto upstream/main - branch is now up to date.

To summarise what's in this PR:
- `safeColumns = columns ?? []` in both FilterForm and SortForm (guards against null and undefined)
- `safeColumns` passed to `<ColumnSelect>` and used in submit button `disabled` condition
- `ColumnSelect.jsx`: ES6 default params (no `defaultProps`), composite keys, null-safe `disabled` check
- FILTER/SORT named constants restored (no magic strings)
- `ColumnSelect.test.jsx`: 7 tests - 55/55 passing
- "Apply Sort" button label for consistency
- Redundant `name="column"` removed from SortForm

Ready for merge!

ivantha:
Thank you for replacing the column text inputs with dropdowns! This work has been superseded by #241 which provides dropdowns across all transform forms. Feel free to follow up on that PR.

PR #157, opened 2026-03-01
feat(export): add xlsx export support and Excel menu action

dhruv1955:
## Summary
Add Excel (.xlsx) export support end-to-end alongside the existing CSV export.

## Why
- CSV is the only export format currently supported
- Many users work primarily in Excel or Google Sheets
- `.xlsx` is a standard format expected in data tooling

## Backend Changes
- Updated `/projects/{project_id}/export` to accept `?format=csv|xlsx` query param
- CSV export unchanged — no breaking changes
- XLSX generated from current project data using `pandas` + `openpyxl`
- Cleaned deprecation warnings in touched backend code (`ConfigDict`, `session.exec(select(...))`)

## Frontend Changes
- Added **Export Excel** button in File tab alongside **Export CSV**
- Reuses same export API with `?format=xlsx`
- File downloads as `{project_name}.xlsx`

## Validation
- `pytest tests/test_new_features.py::TestExportEndpoint -q` ✅
- `npm run lint` ✅
- `npm run test` ✅

### Export Excel button in toolbar
Screenshot 2026-03-02 022659

### Downloaded .xlsx opens correctly in Excel
Screenshot 2026-03-02 022725

## Notes
- `openpyxl` added as dependency for Excel file generation
- No breaking changes to existing CSV export

### DIFF
--- a/dataloom-backend/app/api/dependencies.py
+++ b/dataloom-backend/app/api/dependencies.py
@@ -3,7 +3,7 @@
 import uuid
 
 from fastapi import Depends, HTTPException
-from sqlmodel import Session
+from sqlmodel import Session, select
 
 from app import database, models
 from app.utils.logging import get_logger
@@ -24,7 +24,8 @@ def get_project_or_404(project_id: uuid.UUID, db: Session = Depends(database.get
     Raises:
         HTTPException: 404 if project not found.
     """
-    project = db.query(models.Project).filter(models.Project.project_id == project_id).first()
+    statement = select(models.Project).where(models.Project.project_id == project_id)
+    project = db.exec(statement).first()
     if not project:
         raise HTTPException(status_code=404, detail=f"Project with ID {project_id} not found")
     return project
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -3,9 +3,13 @@
 Handles upload, retrieval, save (checkpoint), and revert operations.
 """
 
+import os
 import uuid
+from io import BytesIO
+from typing import Literal
+from urllib.parse import quote
 
-from fastapi import APIRouter, Depends, File, Form, HTTPException, UploadFile
+from fastapi import APIRouter, Depends, File, Form, HTTPException, Query, Response, UploadFile
 from fastapi.responses import FileResponse
 from sqlmodel import Session
 
@@ -207,10 +211,53 @@ async def revert_to_checkpoint(
 
 
 @router.get("/{project_id}/export")
-async def export_project(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
-    """Download the current working copy of a project as a CSV file."""
+async def export_project(
+    project_id: uuid.UUID,
+    export_format: Literal["csv", "xlsx"] = Query(default="csv", alias="format"),
+    db: Session = Depends(database.get_db),
+):
+    """Download the current working copy of a project as CSV or XLSX."""
     project = get_project_or_404(project_id, db)
-    return FileResponse(project.file_path, media_type="text/csv", filename=f"{project.name}.csv")
+
+    if export_format == "csv":
+        ascii_name = project.name.encode("ascii", errors="replace").decode().replace('"', "'")
+        csv_headers = {
+            "Content-Disposition": (
+                f"attachment; filename=\"{ascii_name}.csv\"; filename*=UTF-8''{quote(project.name)}.csv"
+            )
+        }
+        return FileResponse(
+            project.file_path,
+            media_type="text/csv",
+            headers=csv_headers,
+        )
+
+    MAX_EXPORT_BYTES = 100 * 1024 * 1024  # 100 MB
+    file_size = os.path.getsize(project.file_path)
+    if file_size > MAX_EXPORT_BYTES:
+        raise HTTPException(status_code=413, detail="File too large to export as XLSX")
+
+    ascii_name = project.name.encode("ascii", errors="replace").decode().replace('"', "'")
+    headers = {
+        "Content-Disposition": (
+            f"attachment; filename=\"{ascii_name}.xlsx\"; filename*=UTF-8''{quote(project.name)}.xlsx"
+        )
+    }
+
+    try:
+        df = read_csv_safe(project.file_path)
+        output = BytesIO()
+        df.to_excel(output, index=False, engine="openpyxl")
+    except FileNotFoundError as exc:
+        raise HTTPException(status_code=404, detail="Project data file not found on disk") from exc
+    except Exception as exc:
+        raise HTTPException(status_code=500, detail="Failed to generate XLSX export") from exc
+
+    return Response(
+        content=output.getvalue(),
+        media_type="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+        headers=headers,
+    )
 
 
 @router.delete("/{project_id}")
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -5,7 +5,7 @@ import uuid
 from enum import StrEnum
 from typing import Any
 
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, ConfigDict, field_validator
 
 # --- Enums ---
 
@@ -306,6 +306,4 @@ class LastResponse(BaseModel):
     name: str
     description: str | None
     last_modified: datetime.datetime
-
-    class Config:
-        from_attributes = True
+    model_config = ConfigDict(from_attributes=True)
--- a/dataloom-backend/tests/test_new_features.py
+++ b/dataloom-backend/tests/test_new_features.py
@@ -1,6 +1,7 @@
 """Tests for new features: rename column, cast data type, export, and delete project."""
 
 import csv
+from io import BytesIO
 
 import pandas as pd
 import pytest
@@ -215,8 +216,9 @@ class TestAddDeleteColumnEndpoint:
 
 
 class TestExportEndpoint:
-    def test_export_project(self, client, sample_csv, db):
-        # Upload a project first
+    @pytest.fixture(scope="class")
+    def uploaded_project(self, client, sample_csv, db):
+        """Upload a project once for all tests in this class."""
         with open(sample_csv, "rb") as f:
             response = client.post(
                 "/projects/upload",
@@ -224,7 +226,11 @@ class TestExportEndpoint:
                 data={"projectName": "Export Test", "projectDescription": "Test export"},
             )
         assert response.status_code == 200
-        project_id = response.json()["project_id"]
+        return response.json()["project_id"]
+
+    def test_export_project(self, client, uploaded_project):
+        # Upload a project first (now using fixture)
+        project_id = uploaded_project
 
         # Export the project
         export_response = client.get(f"/projects/{project_id}/export")
@@ -241,6 +247,30 @@ class TestExportEndpoint:
     def test_export_nonexistent_project(self, client):
         response = client.get("/projects/00000000-0000-0000-0000-000000000000/export")
         assert response.status_code == 404
+
+    def test_export_project_xlsx(self, client, uploaded_project):
+        project_id = uploaded_project
+
+        export_response = client.get(f"/projects/{project_id}/export?format=xlsx")
+        assert export_response.status_code == 200
+        assert (
+            "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
+            in export_response.headers["content-type"]
+        )
+
+        exported_df = pd.read_excel(BytesIO(export_response.content))
+        assert list(exported_df.columns) == ["name", "age", "city"]
+        assert len(exported_df) == 4
+        assert exported_df.iloc[0]["name"] == "Alice"
+        assert exported_df.iloc[0]["age"] == 30
+        assert "filename=" in export_response.headers.get("content-disposition", "")
+
+    @pytest.mark.parametrize("bad_format", ["pdf", "json", "xls", ""])
+    def test_export_invalid_format(self, client, uploaded_project, bad_format):
+        project_id = uploaded_project
+
+        response = client.get(f"/projects/{project_id}/export?format={bad_format}")
+        assert response.status_code == 422
 
 
 # --- Delete Endpoint Tests ---
--- a/dataloom-frontend/src/Components/MenuNavbar.jsx
+++ b/dataloom-frontend/src/Components/MenuNavbar.jsx
@@ -83,17 +83,24 @@ const MenuNavbar = ({ projectId, onTransform }) => {
     }
   };
 
-  const handleExport = async () => {
+  const handleExport = async (format = "csv") => {
     try {
-      const blob = await exportProject(projectId);
+      const response = await exportProject(projectId, format);
+      const blob = response.data;
       const url = URL.createObjectURL(blob);
       const a = document.createElement("a");
+      const disposition = response.headers?.["content-disposition"] ?? "";
+      const rfc5987Match = disposition.match(/filename\*=UTF-8''([^;\n]+)/i);
+      const fallbackMatch = disposition.match(/filename[^;=\n]*=(["']?)([^'"\n;]*)\1/);
+      const filename = rfc5987Match
+        ? decodeURIComponent(rfc5987Match[1])
+        : (fallbackMatch?.[2] ?? (format === "xlsx" ? "export.xlsx" : "export.csv"));
       a.href = url;
-      a.download = "export.csv";
+      a.download = filename;
       document.body.appendChild(a);
       a.click();
       a.remove();
-      URL.revokeObjectURL(url);
+      setTimeout(() => URL.revokeObjectURL(url), 100);
     } catch {
       setToast({ message: "Failed to export project.", type: "error" });
     }
@@ -171,7 +178,8 @@ const MenuNavbar = ({ projectId, onTransform }) => {
         group: "Save",
         items: [
           { label: "Save", icon: LuSave, onClick: handleSave },
-          { label: "Export", icon: LuDownload, onClick: handleExport },
+          { label: "Export CSV", icon: LuDownload, onClick: () => handleExport("csv") },
+          { label: "Export Excel", icon: LuDownload, onClick: () => handleExport("xlsx") },
         ],
       },
       {
--- a/dataloom-frontend/src/api/projects.js
+++ b/dataloom-frontend/src/api/projects.js
@@ -4,6 +4,8 @@
  */
 import client from "./client";
 
+const VALID_FORMATS = ["csv", "xlsx"];
+
 /**
  * Upload a new project CSV file.
  * @param {File} file - The CSV file to upload.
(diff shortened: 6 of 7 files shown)


ivantha (commented):
## ◎ Argus Review

This PR cleanly adds end-to-end XLSX export support alongside a welcome set of housekeeping fixes (Pydantic v2 `ConfigDict` migration, SQLModel `select()` API migration). The backend implementation is functionally correct for typical cases, but has several medium-severity concerns: the `Content-Disposition` filename is not RFC 6266-encoded (breaks on special characters, potential header injection), no error handling wraps the XLSX conversion path, and the entire XLSX payload is materialized in RAM before streaming. On the frontend, blob URLs are never revoked (memory leak), and the download filename is hardcoded as `export.xlsx` regardless of the project name — contradicting the backend's carefully set `Content-Disposition` header. Test coverage is functional but shallow: it checks column names and row count but not cell values, and there is no test for invalid `format` values. The overall quality is good and the approach is sound; the issues identified are straightforward to fix.

dataloom-frontend/src/Components/MenuNavbar.jsx
```diff
@@ -83,17 +83,21 @@ const MenuNavbar = ({ projectId, onTransform }) => {
...
   };
 
-  const handleExport = async () => {
+  const handleExport = async (format = "csv") => {
     try {
-      const blob = await exportProject(projectId);
+      const response = await exportProject(projectId, format);
+      const blob = response.data;
       const url = URL.createObjectURL(blob);
       const a = document.createElement("a");
```
ivantha, line 91: **[MEDIUM]** Filename regex ignores RFC 5987 filename* parameter, returns ASCII fallback with '?' chars

The backend sets two Content-Disposition filename tokens:
```
attachment; filename="Donn?es.xlsx"; filename*=UTF-8''Donn%C3%A9es.xlsx
```
The `filename*=` (RFC 5987) token is the standards-compliant, percent-encoded UTF-8 name. The regex used in `MenuNavbar.jsx` is:
```js
disposition.match(/filename[^;=\n]*=(["']?)([^'"\n;]*)\1/)
```
This matches the *first* occurrence — which is the plain `filename=` token — because it appears first in the string. For project names containing non-ASCII characters (e.g. `Données`), the backend intentionally replaces them with `?` in the ASCII fallback. The regex captures `Donn?es.xlsx` rather than the correct `Données.xlsx`, so downloaded XLSX files for such projects will have corrupted filenames.

**Suggestion:**
```
Prefer the `filename*` parameter when present. A robust extraction pattern:
```js
const disposition = response.headers?.["content-disposition"] ?? "";
const rfc5987Match = disposition.match(/filename\*=UTF-8''([^;\n]+)/i);
const fallbackMatch = disposition.match(/filename[^;=\n]*=(["']?)([^'"\n;]*)\1/);
const filename = rfc5987Match
  ? decodeURIComponent(rfc5987Match[1])
  : (fallbackMatch?.[2] ?? (format === "xlsx" ? "export.xlsx" : "export.csv"));
```
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 2]
```diff
@@ -207,10 +210,39 @@ async def revert_to_checkpoint(
...
+    try:
+        df = read_csv_safe(project.file_path)
+        output = BytesIO()
+        df.to_excel(output, index=False)
+        output.seek(0)
+    except FileNotFoundError as exc:
+        raise HTTPException(status_code=404, detail="Project data file not found on disk") from exc
+    except Exception as exc:
+        raise HTTPException(status_code=500, detail="Failed to generate XLSX export") from exc
+
```
ivantha, line 255: **[MEDIUM]** Entire CSV loaded into memory as DataFrame for XLSX conversion — no size guard

The export handler reads the full CSV file into a pandas DataFrame and then serialises it into an in-memory `BytesIO` buffer before returning. For very large project files this approach:
1. Holds two full in-memory copies simultaneously (DataFrame + BytesIO buffer).
2. Blocks the async event loop while pandas does the I/O and openpyxl serialises the workbook.
3. Can exhaust server memory if multiple large exports run concurrently.

There is no file-size check, row-count limit, or streaming mechanism in place.

**Suggestion:**
```
Add a pre-flight size check and consider running the pandas conversion in a thread pool executor to avoid blocking the event loop:
```python
import asyncio
from functools import partial

MAX_EXPORT_BYTES = 100 * 1024 * 1024  # 100 MB

file_size = os.path.getsize(project.file_path)
if file_size > MAX_EXPORT_BYTES:
    raise HTTPException(status_code=413, detail="File too large to export as XLSX")

loop = asyncio.get_event_loop()
output = await loop.run_in_executor(None, partial(_build_xlsx, project.file_path))
```
Extract `_build_xlsx` as a pure synchronous helper.
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 3]
```diff
@@ -207,10 +210,39 @@ async def revert_to_checkpoint(
...
+    project_id: uuid.UUID,
+    export_format: Literal["csv", "xlsx"] = Query(default="csv", alias="format"),
+    db: Session = Depends(database.get_db),
+):
+    """Download the current working copy of a project as CSV or XLSX."""
     project = get_project_or_404(project_id, db)
-    return FileResponse(project.file_path, media_type="text/csv", filename=f"{project.name}.csv")
+    ascii_name = project.name.encode("ascii", errors="replace").decode().replace('"', "'")
+
+    if export_format == "csv":
```
ivantha, line 222: **[LOW]** CSV branch missing RFC 5987 Content-Disposition header — non-ASCII filenames are silently corrupted

The XLSX branch carefully sets both `filename=` (ASCII fallback) and `filename*=UTF-8''` (RFC 5987). However, the CSV branch uses FastAPI's `FileResponse` with a bare `filename=f"{project.name}.csv"`. FastAPI passes this directly as an HTTP header value without encoding, which produces an invalid header for project names containing non-ASCII characters (emojis, accented letters, CJK, etc.). Some browsers will reject or corrupt the filename.

This is a pre-existing issue exposed by the contrast with the new XLSX path.

**Suggestion:**
```
Apply the same RFC 5987 treatment to the CSV response:
```python
if export_format == "csv":
    ascii_name = project.name.encode("ascii", errors="replace").decode().replace('"', "'")
    csv_headers = {
        "Content-Disposition": (
            f"attachment; filename=\"{ascii_name}.csv\"; "
            f"filename*=UTF-8''{quote(project.name)}.csv"
        )
    }
    return FileResponse(
        project.file_path,
        media_type="text/csv",
        headers=csv_headers,
    )
```
(Note: FastAPI's `FileResponse` accepts a `headers` kwarg.)
```

File: dataloom-frontend/src/api/projects.js [hunk 4]
```diff
@@ -64,15 +66,20 @@ export const revertToCheckpoint = async (projectId, checkpointId) => {
...
- * Export the current working copy of a project as a CSV download.
+ * Export the current working copy of a project as a CSV or XLSX download.
  * @param {string} projectId - The project ID.
- * @returns {Promise<Blob>} The CSV file as a Blob.
+ * @param {"csv"|"xlsx"} format - Export format.
+ * @returns {Promise<Blob>} The exported file as a Blob.
  */
-export const exportProject = async (projectId) => {
-  const response = await client.get(`/projects/${projectId}/export`, {
+export const exportProject = async (projectId, format = "csv") => {
```
ivantha, line 74: **[LOW]** JSDoc @returns still documents Blob but function now returns full Axios response

The function signature changed from returning `Promise<Blob>` to returning `Promise<AxiosResponse>`, but the JSDoc was not updated:
```js
 * @returns {Promise<Blob>} The exported file as a Blob.
```
This misleads consumers of the function and any IDE tooling performing type inference from JSDoc.

**Suggestion:**
```
Update the JSDoc:
```js
/**
 * Export the current working copy of a project as a CSV or XLSX download.
 * @param {string} projectId - The project ID.
 * @param {"csv"|"xlsx"} format - Export format.
 * @returns {Promise<import('axios').AxiosResponse<Blob>>} The full Axios response; access `.data` for the Blob.
 */
```
```

File: dataloom-backend/tests/test_new_features.py [hunk 5]
```diff
@@ -242,6 +243,44 @@ def test_export_nonexistent_project(self, client):
...
+                data={"projectName": "Export Invalid Format", "projectDescription": "Test bad format"},
+            )
+        assert response.status_code == 200
+        project_id = response.json()["project_id"]
+
+        response = client.get(f"/projects/{project_id}/export?format={bad_format}")
+        assert response.status_code == 422
+
 
 # --- Delete Endpoint Tests ---
```
ivantha, line 277: **[LOW]** Parametrized invalid-format test creates a new project for every parameter value

`test_export_invalid_format` is parametrized over `["pdf", "json", "xls", ""]`. Each parametrized run performs a full `/projects/upload` POST before testing the export endpoint. The uploaded project is never used during the actual assertion — only the `project_id` matters. This means four identical uploads are made where one would suffice, slowing the test suite and polluting the test database.

**Suggestion:**
```
Extract the upload into a class-level or module-level fixture and share it:
```python
@pytest.fixture(scope="class")
def uploaded_project_id(self, client, sample_csv):
        response = client.post(
            "/projects/upload",
            data={"projectName": "Format Validation", "projectDescription": ""},
        )
    return response.json()["project_id"]

@pytest.mark.parametrize("bad_format", ["pdf", "json", "xls", ""])
def test_export_invalid_format(self, client, uploaded_project_id, bad_format):
    response = client.get(f"/projects/{uploaded_project_id}/export?format={bad_format}")
```
```

File: dataloom-backend/tests/test_new_features.py [hunk 6]
```diff
@@ -242,6 +243,44 @@ def test_export_nonexistent_project(self, client):
...
+                files={"file": ("test.csv", f, "text/csv")},
+                data={"projectName": "Export XLSX Test", "projectDescription": "Test xlsx export"},
+            )
+        assert response.status_code == 200
+        project_id = response.json()["project_id"]
+
+        export_response = client.get(f"/projects/{project_id}/export?format=xlsx")
+        assert export_response.status_code == 200
+        assert (
+            export_response.headers["content-type"]
```
ivantha: **[LOW]** Content-type assertion uses exact equality — will fail if charset is appended

```python
assert (
    export_response.headers["content-type"]
    == "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
)
```
FastAPI (via Starlette) may append `; charset=utf-8` or similar parameters to `Content-Type` headers depending on the `Response` class used. This exact-equality check will produce a false negative if the content-type header value includes any suffix.

**Suggestion:**
```
Use `startswith` or check for the MIME type substring:
```python
assert "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" in export_response.headers["content-type"]
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR cleanly adds XLSX export support end-to-end and includes worthwhile housekeeping (Pydantic v2 `ConfigDict`, SQLModel `select`/`exec` migration). The backend implementation is solid, but there are two noteworthy correctness issues: the frontend filename-extraction regex ignores the RFC 5987 `filename*=` parameter (meaning non-ASCII project names produce filenames with `?` characters), and the `exportProject` API function's return type changed from `Blob` to full `AxiosResponse` without being flagged as a breaking change in the PR description. There are also minor concerns around large-file memory usage and test efficiency.

### ⚠️ Breaking Changes

- **[MEDIUM]** exportProject return type changed from Blob to full Axios response: **Previous behavior:** `exportProject(projectId)` returned `Promise<Blob>` — callers used `const blob = await exportProject(projectId)` directly.

**New behavior:** The function now returns `Promise<AxiosResponse>`, requiring callers to access `response.data` for the blob.

**Who is affected:** Any module that imports `exportProject` from `src/api/projects.js` and expects a Blob. The sole known consumer (`MenuNavbar.jsx`) has been updated in this PR, but if any other component, hook, or test file calls `exportProject` and treats the result as a Blob, it will silently receive an Axios response object instead — causing `URL.createObjectURL()` to fail or behave incorrectly at runtime.

**Migration path:** Update all callers from `const blob = await exportProject(id)` to `const { data: blob } = await exportProject(id)`, or destructure `response.data` explicitly.
  > **Suggestion:** Search the codebase for all `exportProject` usages and update accordingly. Alternatively, maintain backward compatibility by keeping the old single-argument signature returning just the Blob and introducing a new `exportProjectRaw` that returns the full response, or accept an options object: `exportProject(projectId, { format, rawResponse })`.


### PR Description Issues

- **[INFO]** PR description omits the breaking API change to exportProject: The description correctly documents all backend and UI changes and even notes the deprecation cleanups. However, it does not mention the **breaking change** to the `exportProject` frontend API function — its return type changed from `Promise<Blob>` to `Promise<AxiosResponse>`. Any caller that hasn't been updated in this PR (or that is introduced in the future without reading the diff) will receive a silent regression.
  > **Suggestion:** Add a note under **Frontend Changes** such as:
> - `exportProject()` in `src/api/projects.js` now returns the full Axios response object rather than just the Blob. Callers must access `response.data` for the file content. The sole existing callsite (`MenuNavbar.jsx`) has been updated in this PR.
- **[INFO]** Test command references a specific test file path, not a broader suite: The validation section shows `pytest tests/test_new_features.py::TestExportEndpoint -q`, which runs only the export-specific tests. Given that the PR also touched `dependencies.py` and `schemas.py`, running only the export class may miss regressions in related flows (project listing, checkpoint revert, etc.) introduced by the shared dependency changes.
  > **Suggestion:** Update the validation section to reflect a broader run, e.g.:
```
pytest tests/ -q  # full suite
# or at minimum:
pytest tests/test_new_features.py -q  # all new-feature tests
```
This gives reviewers more confidence that shared utilities (`get_project_or_404`, `LastResponse`) weren't inadvertently broken.

File: dataloom-backend/app/api/endpoints/projects.py (hunk 3 above)
Comment (ivantha) [inline dataloom-backend/app/api/endpoints/projects.py:222]: **[LOW]** ascii_name computed unconditionally but only used in the XLSX branch

`ascii_name` is computed immediately after `get_project_or_404` regardless of the requested format. For CSV requests, this string-encoding work is entirely wasted. While the overhead is negligible, moving the computation inside the `xlsx` branch improves readability and makes the intent clear.

**Suggestion:**
```
Move `ascii_name` inside the XLSX branch:
```python
if export_format == "csv":
    return FileResponse(...)

ascii_name = project.name.encode("ascii", errors="replace").decode().replace('"', "'")
headers = { ... }
```
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 7]
```diff
@@ -207,10 +210,39 @@ async def revert_to_checkpoint(
...
+        "Content-Disposition": (
+            f"attachment; filename=\"{ascii_name}.xlsx\"; filename*=UTF-8''{quote(project.name)}.xlsx"
+        )
+    }
+
+    try:
+        df = read_csv_safe(project.file_path)
+        output = BytesIO()
+        df.to_excel(output, index=False)
+        output.seek(0)
```
ivantha: **[LOW]** openpyxl engine not explicitly specified in to_excel call

`df.to_excel(output, index=False)` relies on pandas selecting openpyxl automatically for `.xlsx`. While this is currently correct behaviour, it creates an implicit coupling: if `xlsxwriter` is later added to the environment (it has higher priority for some pandas versions), the engine silently switches. Explicit is better than implicit for reproducibility.

**Suggestion:**
```
```python
df.to_excel(output, index=False, engine="openpyxl")
```
```

File: dataloom-frontend/src/api/projects.js [hunk 8]
```diff
@@ -64,15 +66,20 @@ export const revertToCheckpoint = async (projectId, checkpointId) => {
...
+  if (!VALID_FORMATS.includes(format)) {
+    throw new Error(`Unsupported export format: "${format}"`);
+  }
+  const params = new URLSearchParams({ format });
+  const response = await client.get(`/projects/${projectId}/export?${params.toString()}`, {
     responseType: "blob",
   });
-  return response.data;
+  return response;
```
ivantha, line 82: **[LOW]** Manual URLSearchParams construction instead of Axios params option

The current code manually constructs the query string:
```js
const params = new URLSearchParams({ format });
const response = await client.get(`/projects/${projectId}/export?${params.toString()}`, { responseType: "blob" });
```
Axios natively handles query parameters via the `params` config option, which is more idiomatic and avoids potential double-encoding issues if the base URL already contains a query string.

**Suggestion:**
```
```js
const response = await client.get(`/projects/${projectId}/export`, {
  params: { format },
  responseType: "blob",
});
```
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 9]
```diff
@@ -207,10 +210,39 @@ async def revert_to_checkpoint(
...
+    except FileNotFoundError as exc:
+        raise HTTPException(status_code=404, detail="Project data file not found on disk") from exc
+    except Exception as exc:
```
ivantha, line 253: **[INFO]** output.seek(0) before getvalue() is a no-op

`BytesIO.getvalue()` always returns the entire buffer contents regardless of the current stream position — unlike `read()`, which respects position. The `seek(0)` call is therefore redundant here. It's harmless but could mislead a future maintainer into thinking a `seek` is required before `getvalue()`.

**Suggestion:**
```
Remove the `seek(0)` line, or add a comment explaining why it is present if you anticipate switching to `.read()` in the future:
```python
df.to_excel(output, index=False, engine="openpyxl")
# BytesIO.getvalue() reads the full buffer; seek(0) is not required here
```
```

File: dataloom-backend/app/api/dependencies.py [hunk 10]
```diff
@@ -24,7 +24,8 @@ def get_project_or_404(project_id: uuid.UUID, db: Session = Depends(database.get
     Raises:
         HTTPException: 404 if project not found.
     """
-    project = db.query(models.Project).filter(models.Project.project_id == project_id).first()
+    statement = select(models.Project).where(models.Project.project_id == project_id)
+    project = db.exec(statement).first()
```
ivantha, line 28: **[INFO]** Positive: SQLModel select/exec migration is correct and idiomatic

Replacing the legacy `db.query(...).filter(...).first()` ORM pattern with `select(...).where(...)` + `db.exec(...).first()` is the correct SQLModel v0.0.14+ idiom. This eliminates the SQLAlchemy 2.x deprecation warning and aligns with SQLModel's intended usage. The change is minimal, safe, and improves long-term compatibility.

dataloom-backend/app/schemas.py
```diff
@@ -306,6 +306,4 @@ class LastResponse(BaseModel):
     name: str
     description: str | None
     last_modified: datetime.datetime
-
-    class Config:
-        from_attributes = True
+    model_config = ConfigDict(from_attributes=True)
```
ivantha, line 309: **[INFO]** Positive: Pydantic v2 ConfigDict migration is correct

Replacing the inner `class Config` with `model_config = ConfigDict(from_attributes=True)` is the correct Pydantic v2 migration. The inner class approach is deprecated in Pydantic v2 and will be removed in a future major version. This is a clean, targeted fix.

dhruv1955:
@ivantha Addressed all feedback from the second review:

- Fixed filename regex to prefer RFC 5987 `filename*` parameter - non-ASCII project names now download with correct filenames
- Added 100MB size guard before XLSX generation to prevent memory exhaustion
- Applied RFC 5987 Content-Disposition to CSV branch as well for consistency
- Moved `ascii_name` computation inside XLSX branch only
- Added explicit `engine="openpyxl"` to `df.to_excel()`
- Removed redundant `seek(0)` before `getvalue()`
- Replaced manual `URLSearchParams` with Axios `params` option
- Updated JSDoc `@returns` to reflect full `AxiosResponse<Blob>`
- Fixed content-type assertion to use `in` instead of `==`
- Extracted upload into class-level fixture for parametrized invalid-format tests

ivantha:

```
git fetch origin
git rebase origin/main
git push --force-with-lease
```

The implementation is excellent — RFC5987 filename encoding, 100 MB size guard, proper error handling, parametrized invalid-format tests, and clean frontend integration. Ready to merge once rebased.

ivantha:
Thank you for adding xlsx export! This work has been superseded by #265 which provides multi-format export including xlsx. Feel free to follow up on that PR.

PR #159, opened 2026-03-02
fix(schema): make `name` optional in AddOrDeleteColumn to fix delCol 422 error

### LINKED ISSUE #158: [Bug] Delete Column fails due to schema requiring `name` field for delCol operation
## Bug Description
Right-clicking a column and selecting "Delete Column" fails with 
"Failed to delete column. Please try again." toast.

## Root Cause
`AddOrDeleteColumn` schema requires both `index` and `name` fields.
Delete Column only sends `index` — no `name` needed for deletion.
FastAPI/Pydantic rejects the request with 422 before business logic runs.

The `delCol` handler in `transformations.py` only uses `index`, 
confirming `name` is not needed for delete operations.

## Fix
Two valid approaches:
1. Split schema into `AddColumn` (index + name) and `DeleteColumn` (index only)
2. Make `name` optional in `AddOrDeleteColumn` with `name: str | None = None`

Option 2 is less invasive and safer.

## Steps to Reproduce
1. Open any project
2. Right-click any column header
3. Click "Delete Column"
4. Error toast appears

## Screenshots

Image

Image

dhruv1955:
## Summary
Fix Delete Column failing with 422 Pydantic validation error.

## Root Cause
`AddOrDeleteColumn` schema required both `index` and `name`.
Delete requests only send `index` — name is not needed for deletion.
FastAPI rejected the request with 422 before business logic ran.

## Fix
Made `name` optional in `AddOrDeleteColumn` schema:
- `name: str` → `name: str | None = None`

The `delCol` handler only uses `index`, so this is safe.
Add Column still works as `name` is always provided for add operations.

## Validation
- `pytest` ✅
- `npm run lint` ✅
- `npm run test` ✅

### Before — error toast on delete
Screenshot 2026-03-02 122923

### After — column deleted successfully
Screenshot 2026-03-02 125538

Fixes #158

### DIFF
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -104,11 +104,26 @@ class AddOrDeleteRow(BaseModel):
     index: int
 
 
-class AddOrDeleteColumn(BaseModel):
-    """Parameters for adding or deleting a column by index and name."""
+class AddColumn(BaseModel):
+    """Parameters for adding a column.
+
+    Attributes:
+        index: Zero-based column index where column will be inserted.
+        name: Column name; required for add operations.
+    """
 
     index: int
-    name: str | None = None
+    name: str
+
+
+class DeleteColumn(BaseModel):
+    """Parameters for deleting a column.
+
+    Attributes:
+        index: Zero-based column index to delete.
+    """
+
+    index: int
 
 
 class ChangeCellValue(BaseModel):
@@ -226,7 +241,8 @@ class TransformationInput(BaseModel):
     parameters: FilterParameters | None = None
     sort_params: SortParameters | None = None
     row_params: AddOrDeleteRow | None = None
-    col_params: AddOrDeleteColumn | None = None
+    add_col_params: AddColumn | None = None
+    del_col_params: DeleteColumn | None = None
     fill_empty_params: FillEmptyParams | None = None
     drop_duplicate: DropDuplicates | None = None
     adv_query: AdvQuery | None = None
--- a/dataloom-backend/app/api/endpoints/transformations.py
+++ b/dataloom-backend/app/api/endpoints/transformations.py
@@ -57,15 +57,15 @@ def _handle_basic_transform(df, transformation_input, project, db, project_id):
         return ts.delete_row(df, transformation_input.row_params.index), True
 
     elif op == "addCol":
-        if not transformation_input.col_params:
+        if not transformation_input.add_col_params:
             raise HTTPException(status_code=400, detail="Column parameters required")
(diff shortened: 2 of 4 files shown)


ivantha (commented):
## ◎ Argus Review

This PR correctly resolves a 422 Pydantic validation error on delete-column requests by making `name` optional in the shared `AddOrDeleteColumn` schema. The fix is minimal and surgical. However, the root cause is a design smell — a single schema serving two operations with different field requirements — and the fix trades one problem (delete always fails) for a subtler one (add column silently loses schema-level validation for `name`). No new tests accompany the change, and the docstring is now slightly misleading.

dataloom-backend/app/schemas.py
```diff
@@ -106,7 +106,7 @@ class AddOrDeleteColumn(BaseModel):
     """Parameters for adding or deleting a column by index and name."""
 
     index: int
-    name: str
+    name: str | None = None
```
ivantha: **[MEDIUM]** No tests added for the bug fix or regression guard

The PR description claims `pytest ✅` but the diff contains no new test code. Two test cases are missing:

1. **Bug fix test**: A request to delete a column (providing only `index`, no `name`) should now return `2xx` instead of `422`.
2. **Regression test**: A request to add a column without `name` should still return a meaningful error (currently this is unguarded — see the high-severity finding above).

Without these, the CI green-check provides false confidence: the existing tests may not exercise the delete path at all, and the regression on add is invisible.

**Suggestion:**
```
Add tests such as:

```python
def test_delete_column_without_name_returns_200(client):
    response = client.delete("/columns", json={"index": 2})

def test_add_column_without_name_returns_422(client):
    response = client.post("/columns", json={"index": 2})
```

The second test will currently **fail**, which surfaces the high-severity issue described above.
```

File: dataloom-backend/app/schemas.py [hunk 2]
```diff
@@ -106,7 +106,7 @@ class AddOrDeleteColumn(BaseModel):
     """Parameters for adding or deleting a column by index and name."""
```
ivantha: **[LOW]** Docstring no longer accurately describes the schema

The existing docstring reads:

```
"""Parameters for adding or deleting a column by index and name."""
```

With `name` now optional, the phrase "by index and name" is misleading for delete operations where `name` is irrelevant and not sent by the client.

**Suggestion:**
```
Update the docstring to reflect the optionality:

```python
"""Parameters for adding or deleting a column.

Attributes:
    index: Zero-based column index (required for both add and delete).
    name: Column name; required for add operations, ignored for delete.
"""
```
```

PR #160, opened 2026-03-02
test(schema): add regression test for delCol index-only payload

dhruv1955:
## Summary
Add a regression test to ensure Delete Column (`delCol`) accepts 
index-only payloads and does not fail validation.

## Why
A recent bug caused Delete Column to fail with a 422 error because 
the schema previously required a `name` field even for delete operations.  
This test prevents that regression from returning.

## What Changed
- Added backend regression test in `dataloom-backend/tests/test_new_features.py`
- New test:
  - Uploads a sample project
  - Sends transform request with:
```json
    {
      "operation_type": "delCol",
      "col_params": { "index": 0 }
    }
```
  - Asserts `200 OK`
  - Asserts resulting columns are updated correctly

## Validation
```
pytest -q
71 passed, 11 warnings in 7.22s
```

## Scope
- Test-only PR
- No API or frontend behavior changes

Relates to #158, #159

### DIFF
--- a/dataloom-backend/tests/test_new_features.py
+++ b/dataloom-backend/tests/test_new_features.py
@@ -35,9 +35,7 @@ def uploaded_project(client, sample_csv, db):
                 "projectDescription": "Regression fixture upload",
             },
         )
-    assert response.status_code == 200, (
-        f"Project upload failed with {response.status_code}: {response.text}"
-    )
+    assert response.status_code == 200, f"Project upload failed with {response.status_code}: {response.text}"
     return response.json()["project_id"]
 
 


dataloom-backend/tests/test_new_features.py
```diff
@@ -150,6 +150,24 @@ def test_export_nonexistent_project(self, client):
 
 
 class TestDeleteEndpoint:
+    def test_delete_column_accepts_index_only_params(self, client, sample_csv, db):
+        with open(sample_csv, "rb") as f:
+            response = client.post(
+                "/projects/upload",
+                files={"file": ("test.csv", f, "text/csv")},
+                data={"projectName": "Delete Column Test", "projectDescription": "Regression test"},
```
ivantha: **[INFO]** Consider line length compliance for CI linting

The line `data={"projectName": "Delete Column Test", "projectDescription": "Regression test"},` is likely to exceed the 88- or 79-character limit enforced by Black/Flake8 in many Python projects. If the CI pipeline runs a linter, this could generate a warning or failure that clutters the test output.

**Suggestion:**
```
Break the data dict across lines for readability and compliance:

```python
data={
    "projectName": "Delete Column Test",
    "projectDescription": "Regression test",
},
```
```

ivantha (commented):
## ◎ Argus Review

This PR introduces a schema change making `name` optional in the shared `AddOrDeleteColumn` model and adds a regression test for the `delCol` index-only path. While the test intent is sound and follows existing patterns, the schema change is broader than necessary — it silently relaxes validation for **both** `addCol` and `delCol` operations through a shared model, which risks introducing a different class of runtime bug for add-column paths. The test itself has brittle hardcoded column assertions and is placed in the wrong test class. The PR would benefit from either splitting the schema or adding a discriminating validator, plus a complementary test that confirms `addCol` still enforces a column name.

ivantha:
Please fix lint errors.

PR #161, opened 2026-03-02
Fix delRow replay inconsistency causing KeyError during checkpoint save/revert

anusha19murthy:
## Description

Fixes a replay inconsistency in `apply_logged_transformation()` where the `delRow` action called `df.drop(index)` directly without resetting the DataFrame index.

The live execution path (`delete_row()`) performs:
`df.drop(index).reset_index(drop=True)`

However, the replay path during checkpoint save and revert did not reset the index. This caused index gaps (e.g., `[0, 1, 3, 4]`) after the first deletion. Subsequent logged transformations referencing the same row index could:

- Raise `KeyError` during save/revert
- Prevent checkpoints from being created
- Break revert operations involving multiple row deletions

This PR delegates the `delRow` replay branch to the existing `delete_row()` function, ensuring replay behavior matches live execution and preserving consistent index normalization.

All other transformation types in `apply_logged_transformation()` already delegate to their corresponding helper functions; `delRow` was the only exception.

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

---

## How Has This Been Tested?

- Reproduced failure locally by:
  1. Uploading a CSV with 5+ rows
  2. Deleting the same row index twice
  3. Attempting to save → previously resulted in `KeyError`
- Verified that after fix:
  - Multiple `delRow` operations replay correctly
  - Checkpoint save succeeds
  - Revert to checkpoint works as expected
  - DataFrame index remains sequential after replay

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

---

N/A – backend logic fix.

---

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -392,6 +392,29 @@ def pivot_table(df: pd.DataFrame, index: str, value: str, column: str = None, ag
     return result.reset_index()
 
 
+def drop_na(df: pd.DataFrame, columns: list[str] | None = None) -> pd.DataFrame:
+    """Drop rows with missing/NaN values.
+
+    Args:
+        df: Source DataFrame.
+        columns: Optional list of column names to check for NaN.
+                 If None, drops rows where ANY column has NaN.
+
+    Returns:
+        DataFrame with NaN rows removed.
+    """
+    if columns is not None:
+        if len(columns) == 0:
+            raise TransformationError("columns list must not be empty")
+        missing = [c for c in columns if c not in df.columns]
+        if missing:
+            raise TransformationError(f"Columns not found in dataset: {', '.join(missing)}")
+        df = df.copy()
+        return df.dropna(subset=columns).reset_index(drop=True)
+    df = df.copy()
+    return df.dropna().reset_index(drop=True)
+
+
 def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_details: dict) -> pd.DataFrame:
     """Replay a logged transformation against a DataFrame.
 
@@ -425,12 +448,14 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
             raise TransformationError(f"Row index {index} out of range") from e
 
     elif action_type == "addCol":
-        index = action_details["col_params"]["index"]
-        column_name = action_details["col_params"]["name"]
+        params = action_details.get("add_col_params") or action_details.get("col_params")
+        index = params["index"]
+        column_name = params["name"]
         return add_column(df, index, column_name)
 
     elif action_type == "delCol":
-        index = action_details["col_params"]["index"]
+        params = action_details.get("del_col_params") or action_details.get("col_params")
+        index = params["index"]
         return delete_column(df, index)
 
     elif action_type == "changeCellValue":
@@ -462,6 +487,10 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
     elif action_type == "trimWhitespace":
         column = action_details["trim_whitespace_params"]["column"]
         return trim_whitespace(df, column)
+
+    elif action_type == "dropNa":
+        columns = action_details.get("drop_na_params", {}).get("columns")
+        return drop_na(df, columns)
 
     else:
         logger.warning("Unknown action type in log replay: %s", action_type)
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -416,9 +416,7 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
 
     elif action_type == "delRow":
         index = action_details["row_params"]["index"]
-        if index < 0 or index >= len(df):
-            raise TransformationError(f"Row index {index} out of range")
-        return df.drop(index)
+        return delete_row(df, index)
 
     elif action_type == "addCol":
         index = action_details["col_params"]["index"]
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -232,13 +232,26 @@ def rename_column(df: pd.DataFrame, col_index: int, new_name: str) -> pd.DataFra
 
     Returns:
         DataFrame with the column renamed.
+
+    Raises:
+        TransformationError: If new_name is empty or whitespace.
+        TransformationError: If col_index is out of range.
+        TransformationError: If new_name already exists in df.columns
+                             (unless new_name equals the current column name).
     """
     if col_index < 0 or col_index >= len(df.columns):
-        raise TransformationError(f"Column index {col_index} out of range (0-{len(df.columns) - 1})")
+        raise TransformationError(
+            f"Column index {col_index} is out of range (DataFrame has {len(df.columns)} columns)."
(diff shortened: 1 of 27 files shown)


dataloom-backend/app/services/transformation_service.py
```diff
@@ -416,9 +416,7 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
 
     elif action_type == "delRow":
         index = action_details["row_params"]["index"]
-        if index < 0 or index >= len(df):
-            raise TransformationError(f"Row index {index} out of range")
-        return df.drop(index)
+        return delete_row(df, index)
```
ivantha: **[INFO]** Positive: Fix correctly follows the established delegation pattern

The PR correctly identifies that `delRow` was the **only** action branch in `apply_logged_transformation()` not delegating to a helper function, and resolves the inconsistency by aligning it with every other action type. This is textbook DRY/single-responsibility refactoring: the fix eliminates a hidden divergence between two code paths that were intended to produce identical results. The root cause analysis in the PR description is accurate and the fix is the minimal correct change.

**Suggestion:**
```
No action needed. This is a good pattern to enforce going forward — consider adding a linting rule or code review checklist item that flags any new `action_type` branch in `apply_logged_transformation()` that does not delegate to a named helper function.
```

ivantha: **[HIGH]** Removed explicit bounds check may silently change exception type

The old code performed an explicit bounds check before dropping:

```python
if index < 0 or index >= len(df):
    raise TransformationError(f"Row index {index} out of range")
```

This guard is now removed entirely, delegating all validation responsibility to `delete_row()`. If `delete_row()` does **not** translate out-of-range indices into a `TransformationError` (and instead lets pandas raise a raw `KeyError` or `IndexError`), any caller that catches `TransformationError` for error recovery will now receive an unhandled exception instead — a silent regression in the error contract.

Additionally, there is a subtle semantic inconsistency in the old bounds check itself: it checked `index >= len(df)` (a positional/count-based check) but then called `df.drop(index)` (a label-based operation). After index gaps these two checks are not equivalent, which was partly the root cause of this bug. The fix is correct in delegating to `delete_row()`, but the error-handling contract must be verified.

**Suggestion:**
```
Confirm that `delete_row()` raises `TransformationError` (not a raw pandas exception) when given an out-of-range index. If it does not, add a wrapper or re-raise inside the delegation:

```python
elif action_type == "delRow":
    index = action_details["row_params"]["index"]
    try:
        return delete_row(df, index)
    except (KeyError, IndexError) as e:
        raise TransformationError(f"Row index {index} out of range") from e
```

Alternatively, if `delete_row()` already raises `TransformationError`, document that explicitly in its docstring so future maintainers don't re-introduce the local guard.
```

Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py]: **[MEDIUM]** No automated regression test added for the specific bug scenario

The PR checklist explicitly notes "New tests added" is **unchecked**, despite the bug being fully reproducible with a deterministic, narrow sequence:

1. Create a DataFrame with 5+ rows.
2. Log and replay two consecutive `delRow` actions using the same sequential index.
3. Assert no `KeyError` and that the resulting DataFrame index is contiguous `[0, 1, 2, ...]`.

Without a regression test, this exact failure mode can be silently reintroduced if `delete_row()` is ever refactored to omit `reset_index(drop=True)`, or if a new developer copies the old pattern for a different action type. The bug was subtle and the fix is a one-liner — these are exactly the cases where a regression test has the highest ROI.

**Suggestion:**
```
Add a unit test in the test suite for `apply_logged_transformation`:

```python
def test_delRow_replay_resets_index_after_multiple_deletions():
    df = pd.DataFrame({"A": [10, 20, 30, 40, 50]})
    log = [
        {"action_type": "delRow", "action_details": {"row_params": {"index": 1}}},
        {"action_type": "delRow", "action_details": {"row_params": {"index": 1}}},
    ]
    for entry in log:
        df = apply_logged_transformation(df, entry["action_type"], entry["action_details"])
    assert list(df.index) == list(range(len(df))), "Index must be contiguous after replay"
    assert list(df["A"]) == [10, 40, 50]
```

Also add a test that verifies a `TransformationError` (not `KeyError`) is raised for an out-of-range index replay.
```

Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py]: **[MEDIUM]** Implicit coupling to `delete_row()` internals without a documented contract

The correctness of `apply_logged_transformation()` now depends entirely on `delete_row()` always calling `reset_index(drop=True)`. This is an invisible invariant — nothing in the code or comments communicates that `delete_row()` **must** return a contiguous-index DataFrame for the replay system to work correctly.

If a future developer modifies `delete_row()` (e.g., adds an optional `reset` parameter, or changes behavior for a different use case) they may not realize they are breaking the replay contract. The `apply_logged_transformation` function has no safety net.

**Suggestion:**
```
Add a comment on the delegation line that makes the dependency explicit:

```python
elif action_type == "delRow":
    index = action_details["row_params"]["index"]
    # Delegates to delete_row() to ensure reset_index(drop=True) is applied,
    # keeping replay behavior consistent with live execution and maintaining
    # a contiguous integer index for subsequent logged transformations.
    return delete_row(df, index)
```

Alternatively, add a postcondition assertion (or assert in tests) that the returned DataFrame always has a range-based index after this operation.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR correctly identifies and fixes a real replay inconsistency bug where `delRow` in `apply_logged_transformation()` was the only action type not delegating to its helper function, causing index gaps that led to `KeyError` on subsequent replays. The fix is minimal, elegant, and follows the established DRY pattern used by all other action types. However, the removal of the explicit bounds-check guard introduces a potential regression in error handling semantics, and the absence of automated tests for the specific bug scenario is a notable gap given how reproducible the failure was.

File: dataloom-backend/app/services/transformation_service.py (hunk 1 above)
Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py]: **[LOW]** Old bounds check was semantically inconsistent with label-based `df.drop()`

The removed guard checked `index >= len(df)` which is a **positional/count-based** check, but `df.drop(index)` is a **label-based** operation. These two are only equivalent when the DataFrame has a clean 0-based contiguous integer index. After index gaps (the exact scenario this bug creates), `len(df)` may be 3 while valid labels are `[0, 2, 4]` — so an `index=3` would pass the guard but still raise `KeyError` from pandas.

This means the old guard was never fully reliable as a safety net in the replay scenario, which further supports delegating to `delete_row()`. This is worth noting for historical understanding, but the fix handles it correctly.

**Suggestion:**
```
No immediate action needed beyond the current fix. However, this is worth documenting in a PR comment or inline note to clarify why the seemingly protective bounds check was removed — it was providing false confidence in a state where labels may not match positional indices.
```

**Comment by ivantha** [comment]

```
git fetch origin
git rebase origin/main
git push --force-with-lease
```

The fix is correct and well-explained — delegating `delRow` replay to `delete_row()` ensures `reset_index(drop=True)` runs, matching live execution. The two targeted tests confirm index contiguity and proper error handling. Ready to merge once rebased.

ivantha:
Thank you for fixing the delRow replay inconsistency! This work has been superseded by #237 which provides the same fix. Feel free to follow up on that PR.

PR #167, opened 2026-03-04
fix: clear unapplied logs when reverting to original state

### LINKED ISSUE #166: [Bug] Revert to original does not clear pending unapplied logs, subsequent save re-applies stale transformations
## Describe the bug
When reverting a project to its original state (checkpoint_id=None), 
the `revert_to_checkpoint` endpoint correctly reloads the original file 
but does not delete or clear the pending unapplied logs from the 
`user_logs` table.

This means a subsequent save will replay those stale logs on top of 
the reverted state, silently corrupting the data.

## Steps to Reproduce
1. Upload a CSV project
2. Apply several transformations (e.g. addRow, delCol)
3. Do NOT save (leave logs as unapplied)
4. Call POST `/{project_id}/revert` without a checkpoint_id
5. Verify the working copy is restored 
6. Now call POST `/{project_id}/save`
7. Observe: the previously reverted transformations are re-applied 

## Expected behavior
Reverting to original should also delete all unapplied logs, since 
they represent transformations the user explicitly chose to discard.

## Root Cause
In `app/api/endpoints/projects.py`, the `revert_to_checkpoint` 
endpoint when `checkpoint_id is None`:
- Saves the original file to the working copy 
- Does NOT delete unapplied logs from `user_logs` 

## Suggested Fix
Add a step to delete unapplied logs after reverting:
```python
# Delete unapplied logs after full revert
db.query(models.ProjectChangeLog).filter(
    models.ProjectChangeLog.project_id == project_id,
    models.ProjectChangeLog.applied == False,
).delete()
db.commit()
```

VishakBaddur:
## Summary
Fixes #166

When reverting a project to its original state (no checkpoint_id),  the endpoint was correctly restoring the original file but leaving unapplied logs in the database. A subsequent save would then re-apply those stale transformations on top of the reverted state, silently corrupting the data.

## Root Cause
`revert_to_checkpoint` with `checkpoint_id=None` did not clear unapplied logs after restoring the working copy.

## Fix
Added a delete step to remove all unapplied logs for the project before committing the revert, ensuring the log state matches the restored file state.

## Testing
1. Upload a CSV project
2. Apply several transformations without saving
3. Call revert with no checkpoint_id
4. Call save - previously stale transformations are no longer re-applied

### DIFF
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -7,6 +7,7 @@ import uuid
 
 from fastapi import APIRouter, Depends, File, Form, HTTPException, UploadFile
 from fastapi.responses import FileResponse
+from sqlalchemy.exc import SQLAlchemyError
 from sqlmodel import Session
 
 from app import database, models, schemas
@@ -193,8 +194,20 @@ async def revert_to_checkpoint(
         for log in logs:
             df = apply_logged_transformation(df, log.action_type, log.action_details)
 
+    # Write file first — if this fails, DB is unchanged and state remains consistent.
     save_csv_safe(df, project.file_path)
-    db.commit()
+    # Clear unapplied logs so a subsequent save cannot re-apply stale
+    # transformations on top of the reverted file state.
+    # Applies to all reverts (full and partial) to prevent stale log replay.
+    db.query(models.ProjectChangeLog).filter(
+        models.ProjectChangeLog.project_id == project_id,
+        models.ProjectChangeLog.applied.is_(False),
+    ).delete(synchronize_session="evaluate")
+    try:
+        db.commit()
+    except SQLAlchemyError:
+        db.rollback()
+        raise
 
     resp = dataframe_to_response(df)
     logger.info("Project reverted: id=%s, checkpoint_id=%s", project_id, checkpoint_id)


ivantha (commented):
## ◎ Argus Review

The fix correctly identifies and addresses the root cause of the stale-log corruption bug (issue #166): unapplied logs were not cleared when reverting to the original state, allowing a subsequent save to re-apply discarded transformations. The change is minimal and targets the right layer. However, two high-severity concerns temper the assessment: (1) the delete runs unconditionally for all revert paths, not just `checkpoint_id=None` as the PR implies, making intent ambiguous for checkpoint reverts; and (2) there is no explicit error handling or rollback guard around the `save_csv_safe` call, leaving a crash window where the database and filesystem can fall out of sync. Additional medium-severity issues include the absence of `synchronize_session=False` on the bulk delete and no cleanup of applied logs that postdate the target checkpoint. An automated regression test is also missing, leaving the fix unguarded against future regressions.

dataloom-backend/app/api/endpoints/projects.py
```diff
@@ -188,6 +188,12 @@ async def revert_to_checkpoint(
...
             df = apply_logged_transformation(df, log.action_type, log.action_details)
 
+    # Clear unapplied logs — they represent transformations the user chose to discard
+    db.query(models.ProjectChangeLog).filter(
+        models.ProjectChangeLog.project_id == project_id,
+        models.ProjectChangeLog.applied == False,  # noqa: E712
+    ).delete()
+
     save_csv_safe(df, project.file_path)
     db.commit()
```
ivantha: **[HIGH]** File-system write and DB delete are not atomic — crash window between save and commit

The sequence is:
1. `db.query(...).delete()` — stages the delete in the SQLAlchemy session (not yet committed)
2. `save_csv_safe(df, project.file_path)` — writes bytes to disk
3. `db.commit()` — persists the delete

Two failure modes exist:

**Mode A — `save_csv_safe` raises:** The exception propagates, `db.commit()` is never called. Whether the staged delete is rolled back depends entirely on how the database session is managed by the dependency-injection layer (e.g., a FastAPI `Depends` that calls `db.rollback()` on error). If that guard is absent, the session is left dirty and the next operation on it will behave unpredictably.

**Mode B — process crash after `save_csv_safe` succeeds but before `db.commit()`:** The file on disk now reflects the reverted state, but the database still contains the stale unapplied logs. The data corruption the PR aims to prevent will still occur on the next save.

The safest ordering is **commit first, write file second**, or use a try/except with explicit rollback:

```python
try:
    db.query(models.ProjectChangeLog).filter(
        models.ProjectChangeLog.project_id == project_id,
        models.ProjectChangeLog.applied == False,
    ).delete()
    db.commit()
except Exception:
    db.rollback()
    raise
save_csv_safe(df, project.file_path)  # after commit
```

Or, ideally, wrap the entire endpoint in a transaction context manager.

**Suggestion:**
```
Add an explicit `try/except` with `db.rollback()` in the `except` clause, and consider whether `db.commit()` should precede `save_csv_safe` to reduce the crash window.
```

dataloom-backend/app/api/endpoints/projects.py
```diff
@@ -188,6 +188,12 @@ async def revert_to_checkpoint(
         for log in logs:
             df = apply_logged_transformation(df, log.action_type, log.action_details)
 
+    # Clear unapplied logs — they represent transformations the user chose to discard
+    db.query(models.ProjectChangeLog).filter(
+        models.ProjectChangeLog.project_id == project_id,
+        models.ProjectChangeLog.applied == False,  # noqa: E712
+    ).delete()
```
ivantha: **[MEDIUM]** Applied logs after the target checkpoint are not cleaned up

When reverting to a specific `checkpoint_id`, the function reconstructs `df` by replaying only the applied logs **up to** that checkpoint. However, any `ProjectChangeLog` rows that were `applied=True` but were created **after** the target checkpoint remain in the database untouched. After this revert, those rows are now stale: they describe transformations that are no longer reflected in the working file, yet they are still marked `applied=True`.

The fix addresses only `applied=False` rows. On a subsequent save or replay, those orphaned `applied=True` rows could re-corrupt the file, simply via a different code path than the one being fixed here.

Example scenario:
1. Apply transformations T1, T2, T3 (all `applied=True`)
2. Revert to the checkpoint at T1
3. T2 and T3 remain in the DB as `applied=True`
4. Depending on how save/replay logic works, T2/T3 could be re-applied

**Suggestion:**
```
Determine whether applied logs after the target checkpoint also need to be deleted or marked as `applied=False`. Consider a query like:

```python
db.query(models.ProjectChangeLog).filter(
    models.ProjectChangeLog.project_id == project_id,
    models.ProjectChangeLog.created_at > checkpoint.created_at,
).delete(synchronize_session=False)
```
```

Comment (ivantha) [inline dataloom-backend/app/api/endpoints/projects.py]: **[MEDIUM]** Missing `synchronize_session` argument on bulk delete — potential stale identity map

SQLAlchemy's `Query.delete()` defaults to `synchronize_session='evaluate'`. This strategy attempts to evaluate the Python-side filter expression and remove matching objects from the session's identity map. With ORM-level filters referencing model attributes (like `applied == False`), this evaluation can silently fail or produce incorrect session state if any `ProjectChangeLog` instances were already loaded into the session earlier in the request lifecycle.

For a bulk delete that intentionally bypasses object-level lifecycle events, the safest and most explicit approach is `synchronize_session=False`, which skips session synchronization entirely (the objects are assumed to be irrelevant after the delete):

```python
db.query(models.ProjectChangeLog).filter(
    models.ProjectChangeLog.project_id == project_id,
    models.ProjectChangeLog.applied == False,  # noqa: E712
).delete(synchronize_session=False)
```

Without this, if logs were eagerly loaded earlier (e.g., in the loop above that applies transformations), the session may return stale cached objects in subsequent queries within the same request.

**Suggestion:**
```
Add `synchronize_session=False` to `.delete()` since the endpoint does not use the deleted log objects after this point.
```

dataloom-backend/app/api/endpoints/projects.py
```diff
@@ -188,6 +188,12 @@ async def revert_to_checkpoint(
         for log in logs:
             df = apply_logged_transformation(df, log.action_type, log.action_details)
 
+    # Clear unapplied logs — they represent transformations the user chose to discard
+    db.query(models.ProjectChangeLog).filter(
+        models.ProjectChangeLog.project_id == project_id,
+        models.ProjectChangeLog.applied == False,  # noqa: E712
```
ivantha: **[LOW]** Use `.is_(False)` instead of `== False` for SQLAlchemy filter expressions

While `models.ProjectChangeLog.applied == False` generates valid SQL (`applied = false`) and the `# noqa: E712` suppresses the flake8 warning, SQLAlchemy's idiomatic and semantically clearer form is `.is_(False)`, which generates `applied IS false`. This is:

1. Linter-clean without needing a `# noqa` suppression
2. More explicit about NULL handling (important if `applied` is nullable)
3. The documented SQLAlchemy best practice for boolean comparisons

```python
# Preferred
models.ProjectChangeLog.applied.is_(False)

# Or using the unary NOT operator:
~models.ProjectChangeLog.applied
```

Removing the `# noqa` comment also reduces technical debt in the linting configuration.

**Suggestion:**
```
Replace `models.ProjectChangeLog.applied == False  # noqa: E712` with `models.ProjectChangeLog.applied.is_(False)` and remove the noqa comment.
```

dataloom-backend/app/api/endpoints/projects.py
```diff
@@ -188,6 +188,12 @@ async def revert_to_checkpoint(
...
+    ).delete()
+
```
ivantha: **[LOW]** No automated regression test for the fixed behavior

The PR description outlines a 4-step **manual** test plan. There is no automated test asserting that:

- Unapplied logs are deleted after a revert to original state
- A subsequent save does not re-apply stale transformations
- Unapplied logs are **not** deleted for paths that should not be affected (if any)

Without an automated test, this regression can silently reintroduce itself if the endpoint is refactored or the delete block is accidentally removed.

**Suggestion:**
```
Add a pytest integration test such as:

```python
def test_revert_clears_unapplied_logs(client, db_session, sample_project):
    # Arrange: create unapplied log entries
    log = ProjectChangeLog(project_id=sample_project.id, applied=False, ...)
    db_session.add(log)
    db_session.commit()

    # Act
    response = client.post(f"/projects/{sample_project.id}/revert", json={"checkpoint_id": None})

    # Assert: unapplied logs are gone
    remaining = db_session.query(ProjectChangeLog).filter_by(
        project_id=sample_project.id, applied=False
    ).count()
    assert remaining == 0
```
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 5]
```diff
@@ -188,6 +188,12 @@ async def revert_to_checkpoint(
         for log in logs:
             df = apply_logged_transformation(df, log.action_type, log.action_details)
 
+    # Clear unapplied logs — they represent transformations the user chose to discard
```
ivantha: **[LOW]** Comment scoping could be more precise

The comment `# Clear unapplied logs — they represent transformations the user chose to discard` is a good-faith explanation, but it does not mention:

- *When* this applies (all reverts? only None checkpoint?)
- *Why* this is placed before `save_csv_safe` rather than after
- The original bug that motivated this addition (issue #166)

For future maintainers, a more precise comment prevents misinterpretation of the fix's scope.

**Suggestion:**
```
```python
# Fix for #166: clear all unapplied logs so a subsequent save cannot
# re-apply stale transformations on top of the reverted file state.
# This runs for all revert paths (both checkpoint and original-state reverts).
```
```

File: dataloom-backend/app/api/endpoints/projects.py (hunk 4 above)
Comment (ivantha) [inline dataloom-backend/app/api/endpoints/projects.py]: **[INFO]** Positive: minimal, focused fix that correctly targets the root cause

The fix is well-scoped and correctly identifies the missing delete step as the root cause of the data corruption. Rather than patching the downstream `save` logic (which would be a workaround), the author addresses the source of truth — the stale log entries themselves. The placement before `save_csv_safe` and before `db.commit()` is correct: it groups the delete and the file write into the same logical unit of work. The use of `# noqa: E712` demonstrates awareness of tooling, even if `.is_(False)` would be cleaner.

**Suggestion:**
```
No action required — this is a positive observation.
```

**Comment by VishakBaddur** [comment]
Hi @ivantha, addressed all HIGH and MEDIUM findings:

- Added `if checkpoint_id is None:` guard so the delete only runs for full reverts, not checkpoint reverts
- Switched to `applied.is_(False)` and added `synchronize_session=False`
- Moved `db.commit()` before `save_csv_safe()` and wrapped in try/except with explicit rollback to close the atomicity gap
- Updated comment to clarify scope and reference issue #166

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR correctly identifies and addresses a real data-corruption bug (stale unapplied logs surviving a full revert), but introduces a **critical regression** in the process: by moving `db.commit()` before `save_csv_safe()` for all code paths, a file-write failure after a successful commit will leave the database and filesystem permanently inconsistent. The original code was safer in this regard. Additionally, the fix is incomplete — the same stale-log problem exists when reverting to a specific checkpoint (`checkpoint_id != None`), but that path is left unaddressed. No automated regression test is added, meaning this class of silent-corruption bugs could reappear undetected in future refactors.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes detected — patch-level change only: **Previous behavior:** `revert_to_checkpoint` with `checkpoint_id=None` left unapplied logs in the database after restoring the working copy.

**New behavior:** Unapplied logs are deleted for the project when `checkpoint_id=None`, and the DB commit now happens before the file write for all cases.

**Who is affected:** Internal callers of this endpoint only. No exported types, function signatures, API response shapes, or configuration keys were changed.

**Migration path:** No migration needed. The log-deletion is additive state cleanup. The revert endpoint's request/response contract is unchanged.
  > **Suggestion:** No action required. Tag this as a `patch` release.


### PR Description Issues

- **[MEDIUM]** PR description does not mention the transaction-ordering change: The description accurately explains the log-deletion fix, but it does not mention that `db.commit()` was moved to execute **before** `save_csv_safe()` rather than after. This is a meaningful behavioral change that affects all revert operations (not just `checkpoint_id=None`), and it introduces a new failure mode (DB committed, file write fails → inconsistent state). Reviewers reading only the description would not know to scrutinize this ordering.
  > **Suggestion:** Add a note to the **Fix** section:
> Additionally, the `db.commit()` call was wrapped in a try/except with rollback. **Note:** this reorders the commit to occur before the file write; if `save_csv_safe` fails after a successful commit, the database and filesystem will be in an inconsistent state. Consider whether `save_csv_safe` should remain before the commit as in the original code.
- **[LOW]** Testing section describes only the happy path — no failure scenarios documented: The manual testing steps in the PR description only verify that the fix works correctly when everything succeeds. There is no description of testing failure paths — e.g., what happens if the `save_csv_safe` call fails after the DB commit, or if the delete query fails mid-transaction.
  > **Suggestion:** Expand the Testing section:
```
## Testing
### Happy path
1. Upload a CSV project
2. Apply several transformations without saving
3. Call revert with no checkpoint_id
4. Call save — stale transformations are no longer re-applied ✓

### Failure path
5. Simulate a file-write failure after commit (e.g., make file_path read-only)
6. Verify the system error is surfaced and DB/file state is consistent
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 6]
```diff
@@ -193,8 +193,20 @@ async def revert_to_checkpoint(
...
+    # Only applies to full revert (checkpoint_id=None).
+    if checkpoint_id is None:
+        db.query(models.ProjectChangeLog).filter(
+            models.ProjectChangeLog.project_id == project_id,
+            models.ProjectChangeLog.applied.is_(False),
+        ).delete(synchronize_session=False)
+    try:
+        db.commit()
+    except Exception:
+        db.rollback()
```
ivantha, line 209: **[CRITICAL]** DB committed before file write — failure leaves state permanently inconsistent

The refactor moves `db.commit()` **before** `save_csv_safe(df, project.file_path)` for **all** code paths (not just the `checkpoint_id=None` branch). This is a regression from the original ordering.

**Before this PR:**
```python
save_csv_safe(df, project.file_path)  # file written first
db.commit()                            # DB committed after
```
If the file write failed, the DB was never committed — no data loss.

**After this PR:**
```python
db.commit()                            # DB committed first
save_csv_safe(df, project.file_path)  # file written after
```
If `save_csv_safe` raises (e.g., disk full, permission error, I/O failure), the database has already been committed with the log-deletion applied, but the file on disk still contains the **old, pre-revert state**. The system is now silently inconsistent: the DB says the revert happened and stale logs are gone, but the file still holds the stale data. The only fix would be manual intervention.

This affects every call to `revert_to_checkpoint`, not just the `checkpoint_id=None` case.

**Suggestion:**
```
Keep the file write **before** the commit, and scope the delete + commit together after a successful file write. One safe pattern:
```python
# 1. Write the file first (still safe to retry if this fails — DB unchanged)
save_csv_safe(df, project.file_path)

# 2. Now mutate DB state and commit atomically
if checkpoint_id is None:
    db.query(models.ProjectChangeLog).filter(
        models.ProjectChangeLog.project_id == project_id,
        models.ProjectChangeLog.applied.is_(False),
    ).delete(synchronize_session=False)
try:
    db.commit()
except Exception:
    db.rollback()
    raise
```
This preserves the original "file-first" guarantee: if the DB commit fails, the file has already been updated, which is acceptable since the file can be re-derived; but stale DB logs are far harder to recover from.
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 7]
```diff
@@ -193,8 +193,20 @@ async def revert_to_checkpoint(
         for log in logs:
             df = apply_logged_transformation(df, log.action_type, log.action_details)
 
+    # Fix for #166: clear unapplied logs so a subsequent save cannot
+    # re-apply stale transformations on top of the reverted file state.
+    # Only applies to full revert (checkpoint_id=None).
+    if checkpoint_id is None:
+        db.query(models.ProjectChangeLog).filter(
```
ivantha: **[HIGH]** Unapplied log cleanup is restricted to `checkpoint_id=None` — same bug exists for partial reverts

The guard `if checkpoint_id is None:` means that unapplied logs are **only** cleared when reverting to the original state. When reverting to a specific checkpoint (`checkpoint_id` is set), the function reads and applies all logs up to that checkpoint, then restores the file — but any unapplied logs that were queued **after** that checkpoint are left in the database unchanged.

A subsequent save would then re-apply those stale post-checkpoint transformations on top of the reverted state, producing exactly the same silent data corruption described in the PR's root-cause analysis — just for a different code path.

The PR description says *"Only applies to full revert (checkpoint_id=None)"* as if this were an intentional design choice, but no rationale is given for why partial reverts should behave differently. This appears to be an oversight rather than deliberate.

**Suggestion:**
```
Remove the `checkpoint_id is None` guard and always delete unapplied logs on any revert:
```python
# Clear all unapplied logs regardless of revert target,
# so a subsequent save cannot re-apply stale transformations.
db.query(models.ProjectChangeLog).filter(
    models.ProjectChangeLog.project_id == project_id,
    models.ProjectChangeLog.applied.is_(False),
).delete(synchronize_session=False)
```
If there is a genuine product reason to retain unapplied logs for partial reverts, that reasoning should be documented explicitly in the code comment and the PR description.
```

Comment (ivantha) [inline dataloom-backend/app/api/endpoints/projects.py]: **[MEDIUM]** Bulk delete with `synchronize_session=False` leaves in-memory session stale

`synchronize_session=False` tells SQLAlchemy to skip updating the identity map after the bulk delete. This is a valid performance choice, but it means that any `ProjectChangeLog` objects already loaded into the current session (e.g., from the `logs` query that runs earlier in the same request to apply transformations) will still appear valid/present in the session cache even after the delete.

If any code path between the delete and the commit (or even after, before the session is closed) queries those same objects via the session, it may silently receive stale, already-deleted rows. The current code appears safe because no re-query happens, but this is a fragile assumption — future code additions in this function or middleware could unknowingly read the deleted records.

**Suggestion:**
```
Either use `synchronize_session='evaluate'` (the default, safer option that SQLAlchemy will use when it can reason about the filter expressions) or explicitly expire / expunge the affected objects after the delete:
```python
# Safer: let SQLAlchemy synchronize the session (works for simple filters)
.delete(synchronize_session='evaluate')
```
Or, if `False` is intentional for performance, add a comment:
```python
# synchronize_session=False: bulk delete for performance.
# No in-session re-query of ProjectChangeLog occurs after this point.
.delete(synchronize_session=False)
```
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 8]
```diff
@@ -193,8 +193,20 @@ async def revert_to_checkpoint(
         for log in logs:
             df = apply_logged_transformation(df, log.action_type, log.action_details)
 
+    # Fix for #166: clear unapplied logs so a subsequent save cannot
```
ivantha: **[LOW]** GitHub issue reference in inline code comment is a maintenance liability

The comment `# Fix for #166: clear unapplied logs so a subsequent save cannot...` embeds a GitHub issue number directly in the source code. Issue trackers can be migrated, issues can be closed and become inaccessible, and the repository might move platforms. This makes the comment harder to decode without external context.

Moreover, `# Fix for #166` is commit-archaeology information that belongs in the commit message or PR body — not in the source code itself, where future readers should only need to understand *why* the logic exists, not *where* the bug was tracked.

**Suggestion:**
```
Remove the ticket reference from the code comment and keep only the rationale:
```python
# Clear unapplied logs so a subsequent save cannot re-apply stale
# transformations on top of the reverted file state.
```
The `Fixes #166` reference is already in the PR description, which is preserved in git history.
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 9]
```diff
@@ -193,8 +193,20 @@ async def revert_to_checkpoint(
...
+    # re-apply stale transformations on top of the reverted file state.
+    # Only applies to full revert (checkpoint_id=None).
+    if checkpoint_id is None:
+        db.query(models.ProjectChangeLog).filter(
+            models.ProjectChangeLog.project_id == project_id,
+            models.ProjectChangeLog.applied.is_(False),
+        ).delete(synchronize_session=False)
+    try:
+        db.commit()
```
ivantha, line 207: **[LOW]** `except Exception` with bare `raise` adds no value over no try/except

The pattern:
```python
try:
    db.commit()
except Exception:
    db.rollback()
    raise
```
is only useful if the rollback is meaningful. However, `db.rollback()` here undoes only what has not yet been committed — since this is the `commit()` call itself, a failure mid-commit leaves the session in an uncertain state. Most production SQLAlchemy setups (especially with connection pooling) will call `rollback()` automatically on session close or when an exception propagates to the framework's exception handler.

More importantly, if the commit fails *partially* (rare but possible), rolling back at this point may not be sufficient — the caller likely needs to know the session is tainted and should not be reused.

The bare `except Exception` also catches things like `KeyboardInterrupt` subclasses and `SystemExit` depending on the Python version, which is overly broad.

**Suggestion:**
```
If you want to keep the rollback-on-failure pattern for clarity or defensive coding, use the specific SQLAlchemy exception type and consider using a context manager:
```python
from sqlalchemy.exc import SQLAlchemyError

try:
    db.commit()
except SQLAlchemyError:
    db.rollback()
    raise
```
Or, if the framework already provides session-level rollback on exception propagation, simply:
```python
db.commit()
```
...and let the framework handle it, which is what the original code did.
```

File: dataloom-backend/app/api/endpoints/projects.py [hunk 10]
```diff
@@ -193,8 +193,20 @@ async def revert_to_checkpoint(
...
+    except Exception:
+        db.rollback()
+        raise
     save_csv_safe(df, project.file_path)
-    db.commit()
```
ivantha, line 211: **[INFO]** Positive: good use of `.is_(False)` for nullable boolean filtering

Using `.is_(False)` instead of `== False` is the correct SQLAlchemy idiom for filtering on boolean columns, as it generates `IS FALSE` SQL rather than `= FALSE`, which handles `NULL` correctly in databases where the column may be nullable. This is a subtle but correct detail.

**Suggestion:**
```
No change needed — this is a positive observation. Worth noting in code review culture as a good pattern to follow throughout the codebase.
```

VishakBaddur:
Thanks for the review @ivantha ! Addressed all findings:

[CRITICAL] DB commit ordering restored file-first ordering: save_csv_safe now runs before the DB commit, so a file-write failure leaves the DB unchanged
[HIGH] Partial revert stale logs: removed the checkpoint_id is None guard; unapplied logs are now cleared for all reverts (full and partial)
[MEDIUM] synchronize_session: changed from False to "evaluate" so SQLAlchemy keeps the session in sync after the bulk delete
[LOW] SQLAlchemyError: replaced bare except Exception with except SQLAlchemyError for a more specific catch
[LOW] Issue reference in comment: removed # Fix for #166 from inline comment, kept only the rationale

PR #169, opened 2026-03-04
fix: add missing project 404 check to user_logs endpoints

### LINKED ISSUE #168: [Bug] GET /logs/{project_id} returns empty list for non-existent project instead of 404
## Describe the bug
The `get_logs` endpoint in `user_logs.py` does not validate that the project exists before querying logs. Passing a non-existent project_id 
returns an empty list `[]` with HTTP 200, instead of a 404 error.

All other endpoints use `get_project_or_404` for this check, but `user_logs.py` skips it entirely.

## Steps to Reproduce
1. Call GET `/logs/<random-uuid>` with a project_id that doesn't exist
2. Observe: returns `[]` with HTTP 200 
3. Expected: returns HTTP 404 with "Project not found"

## Root Cause
`app/api/endpoints/user_logs.py` - both `get_logs` and `get_last_checkpoint` are missing the `get_project_or_404` guard used consistently across all other endpoints.

## Suggested Fix
Add `get_project_or_404(project_id, db)` at the start of both endpoint functions.

VishakBaddur:
## Summary
Fixes #168

Both endpoints in `user_logs.py` were missing the `get_project_or_404` guard used consistently across all other endpoints in the codebase.
- `get_logs` returned `[]` with HTTP 200 for non-existent projects
- `get_last_checkpoint` would query checkpoints for a project that doesn't exist

## Fix
Added `get_project_or_404(project_id, db)` at the start of both endpoint functions, consistent with `projects.py` and 
`transformations.py`.

## Related
This inconsistency was found by reviewing the codebase - all other endpoints import and use `get_project_or_404` from `dependencies.py` but `user_logs.py` never imported it.

### DIFF
--- a/dataloom-backend/app/api/endpoints/user_logs.py
+++ b/dataloom-backend/app/api/endpoints/user_logs.py
@@ -4,12 +4,31 @@ from fastapi import APIRouter, Depends, HTTPException
 from sqlmodel import Session
 
 from app import database, models, schemas
+from app.api.dependencies import get_project_or_404
 
 router = APIRouter()
 
 
+@router.get("/checkpoints/{project_id}", response_model=schemas.CheckpointResponse)
+def get_last_checkpoint(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
+    get_project_or_404(project_id, db)
+    last_checkpoint = (
+        db.query(models.Checkpoint)
+        .filter(models.Checkpoint.project_id == project_id)
+        .order_by(models.Checkpoint.created_at.desc())
+        .first()
+    )
+    if not last_checkpoint:
+        raise HTTPException(status_code=404, detail=f"No checkpoints found for project ID {project_id}")
+
+    return schemas.CheckpointResponse(
+        id=last_checkpoint.id, message=last_checkpoint.message, created_at=last_checkpoint.created_at
+    )
+
+
 @router.get("/{project_id}", response_model=list[schemas.LogResponse])
 def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
+    get_project_or_404(project_id, db)
     logs = (
         db.query(models.ProjectChangeLog)
         .filter(models.ProjectChangeLog.project_id == project_id)
@@ -28,20 +47,3 @@ def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
         )
         for log in logs
     ]
-
-
-@router.get("/checkpoints/{project_id}", response_model=schemas.CheckpointResponse)
-def get_last_checkpoint(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
-    last_checkpoint = (
-        db.query(models.Checkpoint)
-        .filter(models.Checkpoint.project_id == project_id)
-        .order_by(models.Checkpoint.created_at.desc())
-        .first()
-    )
-
-    if not last_checkpoint:
-        raise HTTPException(status_code=404, detail=f"No checkpoints found for project ID {project_id}")
-
-    return schemas.CheckpointResponse(
-        id=last_checkpoint.id, message=last_checkpoint.message, created_at=last_checkpoint.created_at
-    )


ivantha (commented):
## ◎ Argus Review

This PR correctly addresses a real consistency bug — both `user_logs.py` endpoints were missing the `get_project_or_404` guard used throughout the rest of the codebase, silently returning HTTP 200 for non-existent projects. The fix is minimal, accurate, and follows established patterns. The most significant concern is the absence of new tests to guard against regression of the original bug, which was itself silent and easy to overlook. Secondary concerns include a missing newline at EOF (a side-effect of the edit), the removal of readability-aiding blank lines, and a pre-existing authorization gap where no ownership check prevents one user from reading another's logs. The route ordering in the file is also fragile in a pre-existing way but functions correctly due to UUID type validation.

dataloom-backend/app/api/endpoints/user_logs.py
```diff
@@ -32,16 +33,15 @@ def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
...
         .first()
     )
-
     if not last_checkpoint:
         raise HTTPException(status_code=404, detail=f"No checkpoints found for project ID {project_id}")
-
     return schemas.CheckpointResponse(
         id=last_checkpoint.id, message=last_checkpoint.message, created_at=last_checkpoint.created_at
-    )
+    )
```
ivantha: **[LOW]** Missing newline at end of file

The diff ends with `\ No newline at end of file`. This violates the POSIX standard (which defines a text file as a sequence of lines each terminated by a newline) and is flagged by most linters, editors, and CI tools.

This appears to be an accidental omission introduced while editing the closing parenthesis of `schemas.CheckpointResponse(...)`.

**Suggestion:**
```
Add a trailing newline after the closing parenthesis on the last line:
```python
    return schemas.CheckpointResponse(
        id=last_checkpoint.id, message=last_checkpoint.message, created_at=last_checkpoint.created_at
    )
```
*(Ensure your editor or a pre-commit hook like `end-of-file-fixer` enforces this automatically.)*
```

File: dataloom-backend/app/api/endpoints/user_logs.py [hunk 2]
```diff
@@ -4,19 +4,20 @@
 from sqlmodel import Session
 
 from app import database, models, schemas
+from app.api.dependencies import get_project_or_404
 
 router = APIRouter()
 
 
 @router.get("/{project_id}", response_model=list[schemas.LogResponse])
 def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
```
ivantha: **[LOW]** Redundant database round-trip for project existence check

Both endpoints now issue **two sequential database queries per request**:
1. `get_project_or_404` fetches the project row to confirm existence.
2. The original query fetches logs or checkpoints filtered by `project_id`.

For `get_logs` in particular, if the project has no logs, the DB could return zero rows from the logs query alone — though this previously returned a silent 200, which is the bug being fixed. The extra query is the correct trade-off for correctness, but there is an opportunity to avoid the overhead.

This is a minor concern and acceptable as-is, but worth flagging for high-traffic endpoints.

**Suggestion:**
```
As an optional optimization (only if profiling shows this is a bottleneck), the existence check could be folded into the main query using a join or a subquery. Alternatively, the return value of `get_project_or_404` could be reused if any project fields are needed in the response later:
```python
project = get_project_or_404(project_id, db)  # capture if project data is needed downstream
```
```

File: dataloom-backend/app/api/endpoints/user_logs.py [hunk 3]
```diff
@@ -32,16 +33,15 @@ def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
 
 @router.get("/checkpoints/{project_id}", response_model=schemas.CheckpointResponse)
 def get_last_checkpoint(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
+    get_project_or_404(project_id, db)
```
ivantha: **[LOW]** Route ordering relies on UUID type validation — fragile if types change

This is a **pre-existing design concern** (not introduced by this PR). The router defines:
```python
@router.get("/{project_id}", ...)          # line ~11, catches any UUID-shaped segment
@router.get("/checkpoints/{project_id}", ...)  # line ~34
```
FastAPI matches routes in definition order. This works correctly today **only because** `project_id` is typed as `uuid.UUID` — FastAPI validates the type at route-matching time, so `/checkpoints/<uuid>` does not match `/{project_id}` (since `"checkpoints"` is not a valid UUID).

However, if someone removes the `uuid.UUID` type annotation (e.g., changes it to `str` for any reason), the `/checkpoints/{project_id}` route would become unreachable — all requests would be swallowed by the broader `/{project_id}` route with `project_id = "checkpoints"`, causing confusing failures.

The fix doesn't worsen this, but since files are being touched, it's a good opportunity to call it out.

**Suggestion:**
```
Reorder the routes so the more specific path (`/checkpoints/{project_id}`) is defined **before** the generic `/{project_id}` route. This is the standard FastAPI best practice and removes the implicit dependency on type-based route disambiguation:
```python
@router.get("/checkpoints/{project_id}", ...)  # specific first
def get_last_checkpoint(...):
    ...

@router.get("/{project_id}", ...)              # generic second
def get_logs(...):
    ...
```
```

File: dataloom-backend/app/api/endpoints/user_logs.py [hunk 4]
```diff
@@ -32,16 +33,15 @@ def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
...
     last_checkpoint = (
         db.query(models.Checkpoint)
         .filter(models.Checkpoint.project_id == project_id)
         .order_by(models.Checkpoint.created_at.desc())
         .first()
     )
-
     if not last_checkpoint:
```
ivantha: **[LOW]** Removal of blank lines reduces intra-function readability

The PR removes two blank lines as a side effect of the changes:
- The blank line after the `logs = (...)` block before `return [...]` in `get_logs`.
- The blank line after the `if not last_checkpoint: raise ...` before `return schemas.CheckpointResponse(...)` in `get_last_checkpoint`.

PEP 8 encourages using blank lines *within* functions sparingly but intentionally to separate logical sections. The removed blank lines previously served that purpose — separating the query phase from the return/serialization phase. Their removal is not harmful but slightly reduces scanability.

**Suggestion:**
```
Restore the blank lines to preserve the visual separation between the query/guard logic and the return statement:
```python
    logs = (
        db.query(models.ProjectChangeLog)
        ...
        .all()
    )

    return [
        schemas.LogResponse(...)
        ...
    ]
```
This is a minor point and can be addressed in a follow-up if the team's formatter handles it.
```

File: dataloom-backend/app/api/endpoints/user_logs.py [hunk 5]
```diff
@@ -32,16 +33,15 @@ def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
...
         raise HTTPException(status_code=404, detail=f"No checkpoints found for project ID {project_id}")
```
ivantha: **[LOW]** Two distinct 404 sources in get_last_checkpoint with potentially inconsistent error formats

`get_last_checkpoint` can now raise a 404 from two separate places:
1. `get_project_or_404(project_id, db)` — raised if the project doesn't exist (message format controlled by `dependencies.py`).
2. `raise HTTPException(status_code=404, detail=f"No checkpoints found for project ID {project_id}")` — raised if the project exists but has no checkpoints.

If `get_project_or_404` uses a different error detail format (e.g., `{"error": "Project not found"}` vs. a plain string), the two 404 responses will have inconsistent shapes, which can complicate client-side error handling.

The fix is logically correct — both cases *should* be 404s — but the inconsistency in error format is worth verifying.

**Suggestion:**
```
Check the error format returned by `get_project_or_404` in `dependencies.py` and ensure it matches the format of the manually raised HTTPException here. If the API has a standardized error envelope, both should use it:
```python
# Consistent format example:
raise HTTPException(status_code=404, detail={"code": "NOT_FOUND", "message": f"No checkpoints found for project {project_id}"})
```
```

**Comment by VishakBaddur** [comment]
Hi @ivantha , addressed the findings:

- Reordered routes so `/checkpoints/{project_id}` is defined before `/{project_id}`: removes the implicit dependency on UUID type validation for route disambiguation
- Restored blank lines between query and return phases in both functions
- Added trailing newline at EOF

**Comment by VishakBaddur** [comment]
Thanks @ivantha! Squashed into a single commit and rebased onto latest main. Previously addressed: route reordering, blank lines restored, trailing newline added. The authorization gap noted is pre-existing and tracked for a follow-up.

**Comment by ivantha** [comment]
The branch is now behind `main` (which just received a merge) and needs to be rebased. Please run:

```
git fetch origin
git rebase origin/main
git push --force-with-lease
```

The fix is clean and correct — adds the missing `get_project_or_404` guard to both `user_logs` endpoints and correctly reorders the routes so the `/checkpoints/{project_id}` path is registered before `/{project_id}` to avoid FastAPI route shadowing. Ready to merge once rebased.

ivantha:
Thank you for adding the 404 check to user_logs endpoints! This work has been superseded by #186 which provides the same fix with tests. Feel free to follow up on that PR.

PR #171, opened 2026-03-04
fix: reset modal form fields on close

### LINKED ISSUE #170: [Bug] New Project modal does not reset form fields on close
## Describe the bug
When the "New Project" modal is closed without submitting, the form fields (Project Name, Project Description, and file upload) retain 
their values. Reopening the modal shows the previously entered data instead of a blank form.

## Steps to Reproduce
1. Click "New Project" to open the modal
2. Enter a project name, description, and select a file
3. Click "Close" without submitting
4. Click "New Project" again
5. Observe: previous values are still present in all fields 

## Expected behavior
The modal should reset all fields to empty/null when closed.

## Root Cause
`handleCloseModal` in `Homescreen.jsx` only calls `setShowModal(false)` but does not reset the three form state values.

## Suggested Fix
```jsx
const handleCloseModal = () => {
  setShowModal(false);
  setProjectName("");
  setProjectDescription("");
  setFileUpload(null);
};
```

VishakBaddur:
## Summary
Fixes #170

When the "New Project" modal was closed without submitting, the form fields (Project Name, Project Description, file upload) retained their values. Reopening the modal would show previously entered data instead of a blank form.

## Root Cause
`handleCloseModal` only called `setShowModal(false)` without resetting the three form state values.

## Fix
Added state resets to `handleCloseModal`:
- `setProjectName("")`
- `setProjectDescription("")`  
- `setFileUpload(null)`

This ensures a clean form every time the modal is opened.

### DIFF
--- a/dataloom-frontend/src/Components/Homescreen.jsx
+++ b/dataloom-frontend/src/Components/Homescreen.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useRef } from "react";
 import { useNavigate } from "react-router-dom";
 import { uploadProject, getRecentProjects, deleteProject } from "../api";
 import { useToast } from "../context/ToastContext";
@@ -59,6 +59,7 @@ const NewProjectCard = ({ onClick }) => (
 const MAX_FILE_SIZE_BYTES = 10 * 1024 * 1024; // 10 MB
 
 const HomeScreen = () => {
+  const fileInputRef = useRef(null);
   const [showModal, setShowModal] = useState(false);
   const [fileUpload, setFileUpload] = useState(null);
   const [projectName, setProjectName] = useState("");
@@ -87,6 +88,12 @@ const HomeScreen = () => {
 
   const handleCloseModal = () => {
     setShowModal(false);
+    setProjectName("");
+    setProjectDescription("");
+    setFileUpload(null);
+    if (fileInputRef.current) {
+      fileInputRef.current.value = "";
+    }
   };
 
   const handleSubmitModal = async (event) => {
@@ -124,7 +131,7 @@ const HomeScreen = () => {
       showToast(message, "error");
     }
 
-    setShowModal(false);
+    handleCloseModal();
     fetchRecentProjects();
   };
 
@@ -210,6 +217,7 @@ const HomeScreen = () => {
             <h2 className="text-2xl font-semibold text-gray-900 mb-4">Upload Dataset</h2>
             <input
               type="file"
+              ref={fileInputRef}
               className="block w-full text-lg text-gray-900 border border-gray-300 rounded-md px-3 py-2 bg-white cursor-pointer focus:outline-none mb-4"
               onChange={handleFileUpload}
             />


ivantha (commented):
## ◎ Argus Review

This is a small, correct, and well-scoped fix that directly addresses the described stale-form-state bug. The three state resets in `handleCloseModal` are appropriately chosen and aligned with the PR description. However, a few adjacent concerns deserve attention: the file input's DOM element may not visually reset from state alone, `handleSubmitModal` may have a parallel gap if it doesn't call `handleCloseModal` on success, and any validation/error states are likely left unreset. No automated tests are included to guard against regressions.

VishakBaddur:
Hi @ivantha , addressed the key findings:

- Added `useRef` on the file input and clear `fileInputRef.current.value = ""` in `handleCloseModal` so the DOM element visually resets, not just React state
- Changed `setShowModal(false)` in the submit handler to call `handleCloseModal()` instead, so the form resets on successful submission too
- Both modal dismiss paths (close button and backdrop click) already delegate to `handleCloseModal` via `onClick={handleCloseModal}`

dataloom-frontend/src/Components/Homescreen.jsx
```diff
@@ -87,6 +88,12 @@ const HomeScreen = () => {
 
   const handleCloseModal = () => {
     setShowModal(false);
+    setProjectName("");
+    setProjectDescription("");
+    setFileUpload(null);
+    if (fileInputRef.current) {
+      fileInputRef.current.value = "";
```
ivantha, line 95: **[INFO]** Positive: Correct use of `useRef` for uncontrolled file input reset

File `<input type="file">` elements are **uncontrolled** in React — their value cannot be set via state alone. The author correctly uses `fileInputRef.current.value = ""` to imperatively reset the DOM node's value.

Without this, the visual file name displayed in the input would persist even after `setFileUpload(null)` cleared the React state, creating a confusing UI where the label shows a file but the state is empty.

**Suggestion:**
```
No change needed — this is the idiomatic React approach for resetting file inputs.
```

dataloom-frontend/src/Components/Homescreen.jsx
```diff
@@ -124,7 +131,7 @@ const HomeScreen = () => {
       showToast(message, "error");
     }
 
-    setShowModal(false);
+    handleCloseModal();
```
ivantha, line 134: **[INFO]** Positive: DRY improvement — `handleSubmitModal` delegates to `handleCloseModal`

The original code had `setShowModal(false)` duplicated directly in `handleSubmitModal`. The fix replaces that call with `handleCloseModal()`, which is the right approach. This:

1. Eliminates the duplication that originally caused the bug (submit didn't reset form fields).
2. Ensures any future additions to `handleCloseModal` (e.g., resetting a loading state) are automatically applied in both the cancel and submit paths.
3. Establishes a single authoritative place for "what happens when the modal closes".

**Suggestion:**
```
No change needed — this is exactly the right refactor.
```

dataloom-frontend/src/Components/Homescreen.jsx
ivantha, line 95: **[MEDIUM]** Potential missing state resets — other form-related states not cleared

Only the three states explicitly mentioned in the PR description are reset in `handleCloseModal`. Without seeing the full component, common patterns in upload forms suggest there are likely additional states that also need to be cleared:

- **Upload progress / loading state** (e.g., `isLoading`, `isUploading`) — if the modal is closed mid-flight (or after a failed attempt), a stale loading indicator could appear on reopen.
- **Validation error messages** (e.g., `nameError`, `fileError`) — any error banners/tooltips that appeared from a failed submit should clear when the modal is dismissed, not persist into the next open.
- **File size / type validation feedback** — often stored as local state to display inline errors.

If none of these exist, this finding does not apply; but the pattern of adding resets should be reviewed against the complete component state.

**Suggestion:**
```
Audit the full component for all `useState` declarations and consider whether each one represents transient form state that should be cleared. For example:

```js
const handleCloseModal = () => {
  setShowModal(false);
  setProjectName("");
  setProjectDescription("");
  setFileUpload(null);
  // Add any of these if they exist:
  // setIsLoading(false);
  // setUploadError(null);
  // setNameError("");
  if (fileInputRef.current) {
    fileInputRef.current.value = "";
  }
};
```
```

File: dataloom-frontend/src/Components/Homescreen.jsx (hunk 2 above)
Comment (ivantha) [inline dataloom-frontend/src/Components/Homescreen.jsx:134]: **[LOW]** Modal close during in-flight submission may cause a state race

If the modal has a close/cancel button that remains interactive while `handleSubmitModal` is awaiting an async response (the `uploadProject` API call), a user could trigger `handleCloseModal` while the async handler is still running. After the await resolves, `handleSubmitModal` will then call `handleCloseModal()` a second time and attempt to set state on potentially stale refs.

While React's state batching generally makes this safe in practice, the double-call could cause:
- Redundant re-renders.
- If `isLoading` state is ever added, it could be set to `true` after the first close has already hidden the modal.

This is a pre-existing structural concern that this PR makes slightly more visible by unifying the close path.

**Suggestion:**
```
Guard against concurrent submission by disabling the close button when a submission is in progress:

```jsx
// Disable the close/cancel button while submitting
<button
  onClick={handleCloseModal}
  disabled={isSubmitting}
>
  Cancel
</button>
```

And track in-flight state:
```js
const [isSubmitting, setIsSubmitting] = useState(false);

const handleSubmitModal = async (event) => {
  event.preventDefault();
  setIsSubmitting(true);
  try {
    // ...
  } finally {
    setIsSubmitting(false);
    handleCloseModal();
  }
};
```
```

File: dataloom-frontend/src/Components/Homescreen.jsx [hunk 3]
```diff
@@ -210,6 +217,7 @@ const HomeScreen = () => {
             <h2 className="text-2xl font-semibold text-gray-900 mb-4">Upload Dataset</h2>
             <input
               type="file"
+              ref={fileInputRef}
               className="block w-full text-lg text-gray-900 border border-gray-300 rounded-md px-3 py-2 bg-white cursor-pointer focus:outline-none mb-4"
               onChange={handleFileUpload}
             />
```
ivantha, line 223: **[LOW]** File input lacks `accept` attribute — pre-existing but worth flagging

The `<input type="file">` has no `accept` attribute. This is not introduced by this PR, but since the PR touches this element (adding `ref`), it is a good opportunity to fix it.

Without `accept`, users can select any file type. The `MAX_FILE_SIZE_BYTES` constant (10 MB) suggests this is a dataset upload, which likely has specific expected formats (e.g., `.csv`, `.json`, `.xlsx`). Allowing any file type means:

1. Users won't get OS-level filtering in the file picker.
2. Invalid file types only fail at the server, giving poor UX.
3. Security-sensitive types (`.exe`, `.sh`) could be selected, though server validation should catch these.

**Suggestion:**
```
Add an `accept` attribute matching your supported dataset formats:

```jsx
<input
  type="file"
  ref={fileInputRef}
  accept=".csv,.json,.xlsx,.xls,.tsv"
  className="..."
  onChange={handleFileUpload}
/>
```
```

File: dataloom-frontend/src/Components/Homescreen.jsx (hunk 1 above)
Comment (ivantha) [inline dataloom-frontend/src/Components/Homescreen.jsx:95]: **[LOW]** No regression tests added for the bug fix

This PR fixes a reproducible user-facing bug (#170), but no automated tests are included to prevent regression. Without a test, a future refactor of `handleCloseModal` or `handleSubmitModal` could silently re-introduce stale form values.

Ideal test coverage would verify:
1. After opening and typing into the form, closing without submitting leaves the state clean.
2. After a successful submission, reopening the modal shows an empty form.
3. The file input's DOM value is cleared (requires a jsdom-based check).

**Suggestion:**
```
Add a test in the component's test file (e.g., `Homescreen.test.jsx`):

```js
it('resets all form fields when modal is closed without submitting', async () => {
  render(<HomeScreen />);

  // Open modal
  fireEvent.click(screen.getByText('New Project'));

  // Fill in fields
  fireEvent.change(screen.getByLabelText(/project name/i), {
    target: { value: 'My Dataset' },
  });
  fireEvent.change(screen.getByLabelText(/description/i), {
    target: { value: 'Some description' },
  });

  // Close modal
  fireEvent.click(screen.getByRole('button', { name: /close/i }));

  // Reopen
  fireEvent.click(screen.getByText('New Project'));

  // Verify fields are empty
  expect(screen.getByLabelText(/project name/i)).toHaveValue('');
  expect(screen.getByLabelText(/description/i)).toHaveValue('');
});
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This is a clean, well-scoped bug fix that correctly addresses stale form state in the New Project modal. The author demonstrates good React knowledge by using `useRef` to reset the uncontrolled file input's DOM value — a subtlety that pure state resets cannot handle. The DRY refactor of `handleSubmitModal` calling `handleCloseModal()` is a meaningful improvement that prevents the same class of bug from re-occurring on the submit path. Minor concerns include the absence of regression tests, the possibility of other form-related states (loading, validation errors) that may also need clearing, and a PR description that omits the `useRef` implementation detail.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes — patch version bump appropriate: **Previous behavior**: `handleCloseModal` only called `setShowModal(false)`. The form state was preserved across modal open/close cycles. `handleSubmitModal` independently called `setShowModal(false)` and did not reset form fields.

**New behavior**: `handleCloseModal` now resets all form state and the file input DOM value. `handleSubmitModal` delegates to `handleCloseModal` instead of calling `setShowModal(false)` directly.

**Who is affected**: This is a purely internal change to a single component (`HomeScreen`). No exported APIs, hooks, types, context shapes, or route contracts are modified.

**Migration path**: N/A — no consumer code changes required.
  > **Suggestion:** No action needed. The `suggestedSemver` for this change is `patch`.


### PR Description Issues

- **[LOW]** PR description omits the `useRef` file input DOM reset — key implementation detail: The description lists the fix as three state resets:
- `setProjectName("")`
- `setProjectDescription("")`
- `setFileUpload(null)`

However, it does not mention the fourth and equally important change: using `useRef` and `fileInputRef.current.value = ""` to reset the file input's DOM value. Without this, the visual filename shown in the native file picker UI would persist even after state reset. This omission means reviewers reading only the description would have an incomplete picture of the fix.
  > **Suggestion:** Update the **Fix** section to include:

> Added state resets to `handleCloseModal`:
> - `setProjectName("")`
> - `setProjectDescription("")`  
> - `setFileUpload(null)`
> - `fileInputRef.current.value = ""` *(resets the native file picker display, since `<input type="file">` is uncontrolled in React)*
>
> A `useRef` was introduced to hold a reference to the file input DOM node for this purpose.
- **[LOW]** PR description missing a testing / verification section: The description explains the problem and the fix clearly, but has no **Testing** or **How to verify** section. Reviewers and QA cannot confirm whether the author manually verified the fix, and there's no guidance on how to reproduce the original bug or confirm the fix.

For a form bug, this is particularly easy to describe — a few step-by-step instructions would suffice.
  > **Suggestion:** Add a **Testing** section to the PR body:

```markdown
## Testing
- [ ] Open the New Project modal, enter a project name and description, and upload a file
- [ ] Close the modal using the X button without submitting
- [ ] Reopen the modal — verify all fields are empty and no file is shown
- [ ] Repeat the above, but this time submit the form — verify the modal reopens clean on next open
```

VishakBaddur:
Thanks @ivantha ! Squashed into a single commit and rebased onto latest main. The MEDIUM finding about other form states (loading, validation errors), checked the full component, no additional transient states exist beyond the three being reset. The accept attribute suggestion and race condition guard are good points tracked for follow-up.

PR #173, opened 2026-03-05
fix: raise TransformationError instead of silently returning unmodified DataFrame in apply_logged_transformation

### LINKED ISSUE #172: [Bug] apply_logged_transformation silently returns unmodified DataFrame for unknown action types
## Describe the bug
`apply_logged_transformation()` in `app/services/transformation_service.py` has a silent failure in its `else` clause:
```python
else:
    logger.warning("Unknown action type in log replay: %s", action_type)
    return df  # silently returns unmodified data
```

If an unknown or unhandled action type is encountered during checkpoint replay, the function returns the original DataFrame unchanged with no error raised. This causes silent data corruption, the user sees incorrect data with no indication that a transformation failed.

## Steps to Reproduce
1. Log a transformation that is not handled in `apply_logged_transformation`
2. Trigger a save/checkpoint replay
3. The transformation is silently skipped, returning wrong data

## Expected behavior
Raise a `TransformationError` instead of silently returning unmodified data, so callers can detect and report the failure clearly.

## Suggested Fix
```python
else:
    raise TransformationError(f"Unknown action type in log replay: {action_type}")
```

VishakBaddur:
## Summary
Fixes #172

## Change
`apply_logged_transformation()` previously logged a warning and silently returned the unmodified DataFrame for unknown action types, causing silent data corruption during checkpoint replay.

Now raises `TransformationError` so callers can detect and report the failure clearly to the user.

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -503,4 +503,4 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
 
     else:
         logger.warning("Unknown action type in log replay: %s", action_type)
-        return df
+        raise TransformationError(f"Unknown action type in log replay: {action_type}")


dataloom-backend/app/services/transformation_service.py
```diff
@@ -460,5 +460,4 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
         return trim_whitespace(df, column)
 
     else:
-        logger.warning("Unknown action type in log replay: %s", action_type)
-        return df
+        raise TransformationError(f"Unknown action type in log replay: {action_type}")
```
ivantha: **[LOW]** Function docstring (if present) should document the raised exception

If `apply_logged_transformation()` has a docstring, it should now document the `TransformationError` exception in its `Raises:` section. This is important for callers who rely on the docstring or generated API documentation to understand the function's contract.

Since this is a new exception condition that previously did not exist, omitting it from the docstring leaves the function's contract undocumented for this case.

**Suggestion:**
```
Update the function docstring to include:

```python
Raises:
    TransformationError: If `action_type` is not a recognized transformation type.
```

This makes the exception part of the documented interface rather than an implementation detail.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR correctly fixes a silent failure anti-pattern where `apply_logged_transformation()` would silently return an unmodified DataFrame for unknown action types, potentially causing undetected data corruption during checkpoint replay. The change is minimal, focused, and directionally correct — failing loudly is far better than failing silently in a data pipeline. However, the single-line change introduces a few concerns worth addressing: the warning log is lost entirely, callers must now be audited for exception-handling readiness, and there are minor style and security considerations around the error message construction.

**Comment by VishakBaddur** [comment]
Hi @ivantha , good catch. I've restored the logger.warning() call before the raise so the audit trail is preserved regardless of how callers handle the exception. All other findings noted, the tests and docstring updates are good suggestions I'll carry forward into future PRs.

File: dataloom-backend/app/services/transformation_service.py [hunk 2]
```diff
@@ -503,4 +503,4 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
 
     else:
         logger.warning("Unknown action type in log replay: %s", action_type)
-        return df
+        raise TransformationError(f"Unknown action type in log replay: {action_type}")
```
ivantha, line 569: **[MEDIUM]** TransformationError Import/Definition Not Verified in Diff

The diff introduces `TransformationError` as the raised exception type, but the diff does not show any import or definition of this class being added or confirmed present. If `TransformationError` is not already defined and imported in `transformation_service.py`, this change will introduce a `NameError` at runtime whenever the `else` branch is hit — replacing one silent failure with a different, potentially more confusing runtime crash.

This is especially concerning because the bug being fixed (#172) is about silent failures; introducing a `NameError` would be equally silent in some error-handling contexts.

**Suggestion:**
```
Confirm that `TransformationError` is already defined (e.g., in a `exceptions.py` module) and imported at the top of `transformation_service.py`. The diff should ideally show the import line to make this verifiable during review. If it doesn't exist yet, add it:
```python
# In app/exceptions.py
class TransformationError(Exception):
    """Raised when a DataFrame transformation fails or encounters an unknown action type."""
    pass

# In transformation_service.py
from app.exceptions import TransformationError
```
```

Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:569]: **[LOW]** Warning Log Removed — Diagnostic Visibility Reduced if Exception Is Caught Upstream

The previous code emitted a `logger.warning(...)` before returning. This has been replaced entirely by the `raise` statement, with no logging retained. While the exception message itself carries the same information, if the exception is caught upstream by a broad `except Exception` handler that swallows or only briefly logs it, there will be no structured warning-level log entry in the application logs for this failure path.

In observability-sensitive systems (e.g., those using log aggregation tools like Datadog, Splunk, or CloudWatch), warning-level logs are often monitored separately from exception traces. Removing the warning log in favor of a pure raise may reduce alertability for on-call engineers who rely on log-level filtering.

**Suggestion:**
```
Consider retaining the warning log before raising, so both the structured log entry and the exception are emitted:
```python
else:
    logger.warning("Unknown action type in log replay: %s", action_type)
    raise TransformationError(f"Unknown action type in log replay: {action_type}")
```
This ensures visibility in log aggregation systems regardless of how the exception is handled by callers.
```

Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:569]: **[MEDIUM]** No Unit Tests Added for New Exception-Raising Behavior

The diff introduces a meaningful behavioral change — raising an exception instead of returning silently — but does not include any new or updated unit tests to verify this behavior. Without a test, there is no automated guard against regressions (e.g., someone reverting the raise to a return in the future, or the exception type being changed).

Additionally, without a test, it is impossible to verify from the diff alone that `TransformationError` is importable and properly raised when an unknown action type is encountered.

**Suggestion:**
```
Add a dedicated test case, for example:
```python
# In tests/test_transformation_service.py
import pytest
from app.services.transformation_service import apply_logged_transformation
from app.exceptions import TransformationError

def test_apply_logged_transformation_raises_on_unknown_action_type():
    df = pd.DataFrame({"col": [1, 2, 3]})
    with pytest.raises(TransformationError, match="Unknown action type in log replay: nonexistent_action"):
        apply_logged_transformation(df, "nonexistent_action", {})
```
Also consider adding a regression test that verifies the old silent-return behavior no longer occurs (i.e., that the DataFrame is *not* returned unmodified for unknown types).
```

Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:569]: **[LOW]** Error Message Uses f-string While Logger Uses %-formatting — Inconsistent Style

The new `raise` statement uses an f-string: `f"Unknown action type in log replay: {action_type}"`, while the removed logger call used `%`-style formatting: `logger.warning("Unknown action type in log replay: %s", action_type)`. This is a minor inconsistency, but more importantly, exception messages that will be included in logs or surfaced to users should follow a consistent formatting convention across the codebase.

For exception messages specifically, f-strings are generally fine, but if the project enforces a style guide (e.g., all user-facing messages use a particular format), this should be verified.

**Suggestion:**
```
Verify the project's exception message formatting convention. If f-strings are acceptable (they usually are for exception messages), this is fine as-is. If the project standardizes on a different pattern, update accordingly:
```python
# Option A: f-string (current, idiomatic Python 3)
raise TransformationError(f"Unknown action type in log replay: {action_type}")

# Option B: .format() style
raise TransformationError("Unknown action type in log replay: {}".format(action_type))
```
```

Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:569]: **[INFO]** Positive: Correct Fix for Silent Data Corruption During Checkpoint Replay

The motivation for this change is sound and well-reasoned. Silently returning an unmodified DataFrame for an unknown action type during checkpoint replay is a particularly dangerous failure mode: it allows corrupted or incomplete replay operations to complete without any indication of failure, potentially propagating incorrect state downstream. Raising a `TransformationError` ensures the failure is surfaced immediately and cannot be ignored by callers, which aligns with the fail-fast principle and makes debugging significantly easier. The fix is minimal, targeted, and does not alter any other logic paths.

**Suggestion:**
```
No change needed. This is a commendable improvement in error-handling hygiene.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This one-line fix correctly addresses a silent data corruption risk in `apply_logged_transformation()` by converting an undetected failure path into an explicit `TransformationError`. The change is well-motivated and directionally correct. However, it introduces a behavioral breaking change that all call sites must be audited for, no unit tests are included to verify the new behavior, and the removed warning log reduces diagnostic visibility in log-aggregation systems. The PR description is also incomplete, lacking a testing section and any breaking change notice. The key pre-merge actions are: verify `TransformationError` is imported, audit all call sites, and add a regression test.

### ⚠️ Breaking Changes

- **[HIGH]** Behavioral Breaking Change: Exception Raised Where Silent Return Previously Occurred: **Previous behavior**: When `apply_logged_transformation()` received an unknown `action_type`, it logged a warning and returned the original DataFrame unmodified. Callers always received a `pd.DataFrame` back and could continue execution normally — even if silently receiving stale/incorrect data.

**New behavior**: The function now raises `TransformationError` for unknown action types. Any caller that does not explicitly catch `TransformationError` will now experience an unhandled exception and crash.

**Who is affected**: Every call site of `apply_logged_transformation()` — particularly checkpoint replay pipelines, batch transformation loops, and any code paths that invoke this function without a `try/except TransformationError` block. If this function is called in a loop or pipeline, a single unknown action type will now abort the entire operation rather than silently skipping it.

**Migration path**: Callers must be audited and updated to handle `TransformationError`. For example:
```python
try:
    df = apply_logged_transformation(df, action_type, action_details)
except TransformationError as e:
    logger.error("Transformation failed: %s", e)
    raise  # or handle gracefully
```
Any existing `except Exception` blocks will catch this implicitly, but explicit handling is strongly preferred for clarity.
  > **Suggestion:** Before merging, search the codebase for all call sites of `apply_logged_transformation` and verify each one handles `TransformationError`. A grep like `grep -rn 'apply_logged_transformation'` across the project is recommended. Consider adding a note in the PR description documenting which callers were audited.


### PR Description Issues

- **[MEDIUM]** PR Description Missing Testing Section: The PR description includes a summary and a change description, but contains no testing section — no mention of how the fix was tested, what existing tests cover the change, or whether new tests were added. Given that this is a behavioral change in error handling, the absence of a testing section makes it impossible to assess confidence in the fix during review.

This is especially important here because the diff itself contains no new tests, raising the question of whether the change was tested at all.
  > **Suggestion:** Add a **Testing** section to the PR description, for example:
```markdown
## Testing
- [ ] Added unit test `test_apply_logged_transformation_raises_on_unknown_action_type` in `tests/test_transformation_service.py`
- [ ] Verified existing transformation tests pass with `pytest tests/`
- [ ] Manually tested checkpoint replay with a valid action log to confirm no regression
```
- **[MEDIUM]** PR Description Does Not Call Out Breaking Change: This change alters the observable behavior of `apply_logged_transformation()` for all callers that pass an unknown action type: previously they received a DataFrame back; now they receive an exception. This is a breaking change for any caller that was (intentionally or accidentally) relying on the silent-return behavior.

The PR description does not include a **Breaking Changes** section or any migration guidance. Reviewers and downstream maintainers may not realize they need to audit call sites before merging.
  > **Suggestion:** Add a **Breaking Changes** section:
```markdown
## Breaking Changes
⚠️ `apply_logged_transformation()` now raises `TransformationError` for unknown action types instead of silently returning the input DataFrame.

**Migration**: Any caller that previously relied on the silent return (e.g., in a checkpoint replay loop) must now catch `TransformationError` explicitly. The following call sites were audited: [list audited files/functions here].
```

VishakBaddur:
Thanks @ivantha! Squashed and rebased. To address your findings: TransformationError is defined in the file itself (line 15), no separate import needed. The logger.warning was already restored before the raise in the previous commit. All call sites of apply_logged_transformation were audited, the only caller is the checkpoint replay loop in projects.py which is wrapped in a try/except that will surface the error correctly.

ivantha:
This branch is behind main (likely due to a recent merge). Please rebase onto main and re-push so it can be merged cleanly.

PR #178, opened 2026-03-10
fix: block duplicate target names in rename column

dhruv1955:
## Problem
Currently a column can be renamed to a name that already exists in the dataframe.  
This creates duplicate column names, which leads to ambiguous behaviour and can break downstream operations.

## Fix
Added a duplicate-name check in `rename_column()` inside `transformation_service.py`.

Before applying the rename, the function now checks whether `new_name` already exists in `df.columns` (unless it’s the same column being renamed). If it does, the operation is rejected with a clear error message.

**Error returned**

```
Column '{new_name}' already exists. Please choose a different name.
```

## Behaviour
- Renaming to an existing column name → blocked with a clear error
- Renaming to the same name (no-op) → allowed
- Case-sensitive rename (e.g. `age` → `Age`) → allowed

## Tests
Added tests to cover the new behaviour:

- `test_rename_column_to_existing_name`
- `test_rename_column_to_same_name`
- `test_rename_column_case_sensitive`

## Validation
- `pytest` → all tests passing  
- `ruff check .`  
- `ruff format --check .`   

## Files Changed
- `dataloom-backend/app/services/transformation_service.py`
- `dataloom-backend/tests/test_new_features.py`

### Original table
Screenshot 2026-03-10 161354

### Table after rename attempt
Screenshot 2026-03-10 161404

### Error message when duplicate name is used
Screenshot 2026-03-10 161247

## Notes
- Backend-only change
- No schema or frontend changes
- Reuses the existing `TransformationError` for consistency

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -232,13 +232,26 @@ def rename_column(df: pd.DataFrame, col_index: int, new_name: str) -> pd.DataFra
 
     Returns:
         DataFrame with the column renamed.
+
+    Raises:
+        TransformationError: If new_name is empty or whitespace.
+        TransformationError: If col_index is out of range.
+        TransformationError: If new_name already exists in df.columns
+                             (unless new_name equals the current column name).
     """
     if col_index < 0 or col_index >= len(df.columns):
-        raise TransformationError(f"Column index {col_index} out of range (0-{len(df.columns) - 1})")
+        raise TransformationError(
+            f"Column index {col_index} is out of range (DataFrame has {len(df.columns)} columns)."
+        )
     if not new_name or not new_name.strip():
-        raise TransformationError("New column name cannot be empty")
+        raise TransformationError("New column name cannot be empty or whitespace.")
 
     old_name = df.columns[col_index]
+
+    # Check if new_name already exists and is different from current column name
+    if new_name in df.columns and new_name != old_name:
+        raise TransformationError(f"Column '{new_name}' already exists. Please choose a different name.")
+
     return df.rename(columns={old_name: new_name})
 
 
--- a/dataloom-backend/tests/test_transformations.py
+++ b/dataloom-backend/tests/test_transformations.py
@@ -16,6 +16,7 @@ from app.services.transformation_service import (
     drop_duplicates,
     fill_empty,
     pivot_table,
+    rename_column,
 )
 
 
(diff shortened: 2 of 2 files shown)


ivantha (commented):
## ◎ Argus Review

This is a clean, well-scoped fix that correctly addresses a real data-integrity bug. The guard logic (`new_name in df.columns and new_name != old_name`) is placed at the right location, uses the existing `TransformationError` convention for consistency, and the three new tests map directly to the three behavioral guarantees stated in the PR. The main concerns are: a missing edge-case test for DataFrames that already contain duplicate column names (the behavior in this scenario is correct but undocumented), a pre-existing unhandled `IndexError` path for out-of-bounds `col_index` that the new code makes more conspicuous, and the long-term maintainability issue of accumulating tests in a generic `test_new_features.py` file rather than a service-specific test module.

dataloom-backend/tests/test_new_features.py
```diff
@@ -69,6 +69,22 @@ def test_rename_column_whitespace_name(self, sample_df):
...
+    def test_rename_column_to_existing_name(self, sample_df):
+        with pytest.raises(TransformationError, match="already exists"):
+            rename_column(sample_df, 1, "name")  # Try to rename "age" to "name"
+
+    def test_rename_column_to_same_name(self, sample_df):
+        # Renaming to the same name should succeed (no-op)
+        result = rename_column(sample_df, 0, "name")
+        assert list(result.columns) == ["name", "age", "city"]
+        assert result.iloc[0]["name"] == "Alice"
+
```
ivantha: **[LOW]** No-op rename test only checks first row — data integrity not fully verified

`test_rename_column_to_same_name` asserts `result.iloc[0]["name"] == "Alice"`. This verifies the column header survived the no-op, but only inspects one cell. A more robust test would also confirm that the DataFrame's shape, dtypes, and all row values are unchanged to guard against an inadvertent data corruption bug in `df.rename` edge cases.

**Suggestion:**
```
```python
def test_rename_column_to_same_name(self, sample_df):
    result = rename_column(sample_df, 0, "name")
    assert list(result.columns) == ["name", "age", "city"]
    # Verify shape and data integrity, not just one cell
    assert result.shape == sample_df.shape
    pd.testing.assert_frame_equal(result, sample_df)
```
```

File: dataloom-backend/tests/test_new_features.py [hunk 2]
```diff
@@ -69,6 +69,22 @@ def test_rename_column_whitespace_name(self, sample_df):
...
+    def test_rename_column_case_sensitive(self, sample_df):
+        # Renaming to different case should succeed (case-sensitive)
+        result = rename_column(sample_df, 1, "Age")  # Rename "age" to "Age"
+        assert list(result.columns) == ["name", "Age", "city"]
+        assert result.iloc[0]["Age"] == 30
+
```
ivantha: **[INFO]** Good: three core behaviors are explicitly and clearly tested

The three tests map directly and traceably to the three behavioral guarantees stated in the PR description:
- `test_rename_column_to_existing_name` → blocked with error
- `test_rename_column_to_same_name` → allowed (no-op)
- `test_rename_column_case_sensitive` → allowed (case-sensitive)

Each test has a comment explaining *why* the behavior is expected, making them easy to understand in isolation without needing to read the implementation. This is commendable.

**Suggestion:**
```
No change needed. This is a positive pattern worth preserving in future tests.
```

dataloom-backend/app/services/transformation_service.py
```diff
@@ -239,6 +239,11 @@ def rename_column(df: pd.DataFrame, col_index: int, new_name: str) -> pd.DataFra
         raise TransformationError("New column name cannot be empty")
 
     old_name = df.columns[col_index]
+
+    # Check if new_name already exists and is different from current column name
+    if new_name in df.columns and new_name != old_name:
+        raise TransformationError(f"Column '{new_name}' already exists. Please choose a different name.")
+
```
ivantha: **[INFO]** Discussion point: case-insensitive duplicate detection may be needed for some backends

The check `new_name in df.columns` is case-sensitive by Python's default string equality. The PR explicitly documents and tests this (`age` → `Age` is allowed). This is correct for Python/Pandas, but worth raising as a discussion point:

- **SQL databases** (PostgreSQL, MySQL, SQLite) are typically case-insensitive for identifiers. If this DataFrame is later written to a SQL table, renaming `age` to `Age` could silently collide with the existing `age` column at the database layer.
- **CSV/Excel exports** would preserve case, so `age` and `Age` would be two distinct columns.

The current behavior is correct for in-memory Pandas operations. If the app has a SQL persistence layer, a follow-up ticket to add an optional `case_sensitive` parameter (defaulting to `True`) might be worth filing.

**Suggestion:**
```
No immediate code change required, but consider filing a follow-up issue:
```
Title: Consider case-insensitive column name uniqueness check for SQL-backed DataFrames
Description: rename_column() currently blocks exact-match duplicates. 
For SQL backends, a case-insensitive check may be warranted.
```
Alternatively, document the case-sensitivity assumption in the function's docstring.
```

PR #179, opened 2026-03-10
fix: enforce max file size validation for CSV uploads

dhruv1955:
## Problem
The backend advertised file size validation for uploads, but in practice it only checked the file extension. This allowed CSV files larger than the configured `max_upload_size_bytes` limit in `config.py` to be uploaded, which could negatively affect server performance and resource usage.

## Fix
Added proper file size validation in `validate_upload_file()` (`app/utils/security.py`).

Key changes:
- Converted the function to `async` to work with FastAPI's async file handling.
- Read the uploaded file content to determine its size.
- Reject files exceeding `max_upload_size_bytes` with a `400` error.
- Reset the file pointer with `await file.seek(0)` so the upload flow continues normally.
- Updated the call site in `projects.py` to `await validate_upload_file()`.

**Error returned**
```
File size exceeds the maximum allowed size of 10MB.
```

## Tests Added
- `test_file_under_size_limit` – verifies a small file is accepted.
- `test_file_at_size_limit` – verifies a file exactly at the 10MB limit is accepted.
- `test_file_over_size_limit` – verifies files over the limit return a `400` error.

## Validation
- `pytest tests/test_upload.py` → 7/7 passing  
- `pytest` → 81/82 passing (1 unrelated pre-existing failure)  
- `ruff check .` → passed  
- `ruff format --check .` → passed  

## Files Changed
- `dataloom-backend/app/utils/security.py`
- `dataloom-backend/app/api/endpoints/projects.py`
- `dataloom-backend/tests/test_upload.py`

## Notes
- Backend-only change  
- No schema or frontend changes required  
- The failing test (`test_add_column_without_name_returns_422`) already fails on `main` and is unrelated to this change

### DIFF
--- a/dataloom-backend/app/utils/security.py
+++ b/dataloom-backend/app/utils/security.py
@@ -21,37 +21,13 @@ def sanitize_filename(filename: str) -> str:
     Returns:
         A safe, unique filename string.
     """
-    # Strip any directory path components (prevents ../../../etc/passwd)
     name = Path(filename).name
-    # Replace any non-alphanumeric chars (except dots, hyphens, underscores) with underscores
     name = re.sub(r"[^\w.\-]", "_", name)
-    # Prepend UUID for uniqueness
     return f"{uuid.uuid4().hex[:8]}_{name}"
 
 
-def _format_size(size_bytes: int) -> str:
-    """Format a byte count into a human-readable string (e.g. '10.0 MB').
-
-    Args:
-        size_bytes: Size in bytes.
-
-    Returns:
-        Human-readable size string.
-    """
-    for unit in ("B", "KB", "MB", "GB"):
-        if size_bytes < 1024:
-            return f"{size_bytes:.1f} {unit}"
-        size_bytes /= 1024
-    return f"{size_bytes:.1f} TB"
-
-
-def validate_upload_file(file: UploadFile) -> None:
-    """Validate an uploaded file's extension and size.
-
-    Checks the file extension against the allowed list and reads the file
-    content to verify it does not exceed the configured maximum size.
-    After validation the file cursor is reset to the beginning so
-    downstream consumers can read it normally.
+async def validate_upload_file(file: UploadFile) -> None:
+    """Validate an uploaded file extension and size.
 
     Args:
         file: The FastAPI UploadFile object.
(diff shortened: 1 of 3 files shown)


ivantha (commented):
## ◎ Argus Review

This PR correctly fixes the advertised-but-missing file size validation and properly converts the function to async. The core logic is sound, but the implementation reads the entire file into memory before checking its size, which is a meaningful performance and security concern in a web server context. Test coverage is thorough and includes good boundary-value testing, though the tests hardcode the size limit instead of reading from config. Overall the PR solves the stated problem but leaves a notable resource-consumption footgun.

dataloom-backend/app/utils/security.py
```diff
@@ -47,6 +47,16 @@ def validate_upload_file(file: UploadFile) -> None:
             status_code=400, detail=f"File type '{ext}' not allowed. Allowed: {settings.allowed_extensions}"
         )
 
+    # Check file size
+    await file.seek(0)  # Seek to beginning to ensure we're at start
+    content = await file.read()
+    file_size = len(content)
```
ivantha: **[INFO]** Positive: seek(0) reset after read ensures safe handoff to downstream consumers

The `await file.seek(0)` after reading correctly resets the file pointer so that `store_upload(file)`, called immediately after in `projects.py`, receives the file in a readable state from the beginning. This is the right contract for a validation function: consume what you need, leave the resource in a usable state for the caller. The intent is documented with an inline comment.

**Suggestion:**
```
No change needed. The pattern is correct. If the implementation is refactored to use `file.size` and avoid reading, this seek becomes unnecessary and can be removed.
```

ivantha:
@dhruv1955 Please resolve conflicts, rebase and squash all commits into one.

dhruv1955:
Hey @ivantha, rebased on upstream/main and squashed into a single commit. Conflicts resolved - our async + file.size fast-path implementation is preserved.

dhruv1955:
Hi @ivantha, I’ve rebased the branch onto the latest `main` and squashed everything into a single commit. Let me know if anything else is needed

ivantha:
This branch is behind main. Please rebase onto main and re-push so it can be merged cleanly.

dhruv1955:
@ivantha

PR #185, opened 2026-03-12
fix: move onClose to success path and add loading state in TrimWhitespaceForm

dhruv1955:
## Problem
`TrimWhitespaceForm` called `onClose()` unconditionally after the try/catch block, so the form dismissed itself even when the API returned an error. The error toast would flash and disappear along with the form, forcing the user to reopen the menu and reselect their column to retry.

Additionally, the submit button had no loading guard, allowing multiple rapid submissions on slow networks.

## Fix
- Moved `onClose()` inside the `try` block (success path only)
- Added `loading` state with `setLoading(true/false)` around the API call
- Disabled the submit button during the request
- Added `disabled:opacity-50 disabled:cursor-not-allowed` styling
- Changed button label to `"Applying..."` while loading

### Loading state - button disabled during API call
Screenshot 2026-03-13 012341

### DIFF
--- a/dataloom-frontend/src/Components/forms/TrimWhitespaceForm.jsx
+++ b/dataloom-frontend/src/Components/forms/TrimWhitespaceForm.jsx
@@ -10,9 +10,11 @@ const TrimWhitespaceForm = ({ projectId, onClose, onTransform }) => {
   const { showToast } = useToast();
 
   const [column, setColumn] = useState("");
+  const [loading, setLoading] = useState(false);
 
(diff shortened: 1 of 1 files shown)


dhruv1955:
Hi @ivantha, I’ve rebased the branch onto the latest `main` and squashed everything into a single commit. Let me know if anything else is needed

ivantha:

dhruv1955:
@ivantha

PR #186, opened 2026-03-12
fix: return 404 for logs endpoints when project does not exist

dhruv1955:
## Problem

`GET /logs/{project_id}` and `GET /logs/checkpoints/{project_id}` returned `HTTP 200` with an empty list `[]` for any random UUID that doesn't exist in the database.

Every other project endpoint (`get_project_details`, `save_project`, `revert_to_checkpoint`, `export_project`, `delete_project`) already uses `get_project_or_404` from `dependencies.py` to return a proper `404 Not Found`. The logs routes were the only ones skipping this check entirely - silently returning success for invalid project IDs instead of surfacing the error.

This inconsistency can mask frontend routing bugs where a stale or incorrect project ID is used - the UI would receive an empty logs list and show nothing instead of getting a clear error to handle.

## Fix

- Added `from app.api import dependencies` import to `user_logs.py`
- Added `dependencies.get_project_or_404(project_id, db)` as the first call in both `get_logs` and `get_last_checkpoint` handlers
- Added 2 regression tests verifying both endpoints return `404` with the correct error message for a non-existent project UUID

### After fix — `GET /logs/{project_id}` with non-existent project returns `404`
Screenshot 2026-03-13 013746

### DIFF
--- a/dataloom-backend/app/api/endpoints/user_logs.py
+++ b/dataloom-backend/app/api/endpoints/user_logs.py
@@ -4,12 +4,14 @@ from fastapi import APIRouter, Depends, HTTPException
 from sqlmodel import Session
 
 from app import database, models, schemas
+from app.api import dependencies
 
 router = APIRouter()
 
 
 @router.get("/{project_id}", response_model=list[schemas.LogResponse])
 def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
+    _project = dependencies.get_project_or_404(project_id, db)
     logs = (
         db.query(models.ProjectChangeLog)
         .filter(models.ProjectChangeLog.project_id == project_id)
@@ -32,6 +34,7 @@ def get_logs(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
 
 @router.get("/checkpoints/{project_id}", response_model=schemas.CheckpointResponse)
 def get_last_checkpoint(project_id: uuid.UUID, db: Session = Depends(database.get_db)):
+    _project = dependencies.get_project_or_404(project_id, db)
     last_checkpoint = (
         db.query(models.Checkpoint)
         .filter(models.Checkpoint.project_id == project_id)
--- a/dataloom-backend/tests/test_user_logs.py
+++ b/dataloom-backend/tests/test_user_logs.py
@@ -1,0 +1,44 @@
+import uuid
+
+
+class TestLogsEndpoint:
+    def test_get_logs_nonexistent_project_returns_404(self, client):
+        nonexistent_id = uuid.uuid4()
+        response = client.get(f"/logs/{nonexistent_id}")
+        assert response.status_code == 404
+        detail = response.json()["detail"]
+        assert isinstance(detail, str)
+        assert "not found" in detail.lower()
+
+    def test_get_checkpoints_nonexistent_project_returns_404(self, client):
+        nonexistent_id = uuid.uuid4()
+        response = client.get(f"/logs/checkpoints/{nonexistent_id}")
+        assert response.status_code == 404
+        detail = response.json()["detail"]
+        assert isinstance(detail, str)
+        assert "not found" in detail.lower()
+
+    def test_get_logs_existing_project_returns_200(self, client):
+        # Create a project first
+        create_response = client.post(
+            "/projects/",
+            json={"name": "logs-test-project"},
+        )
+        assert create_response.status_code == 200
+        project_id = create_response.json()["id"]
+
+        response = client.get(f"/logs/{project_id}")
+        assert response.status_code == 200
+        assert isinstance(response.json(), list)
+
+    def test_get_logs_existing_project_no_logs_returns_empty_list(self, client):
+        create_response = client.post(
+            "/projects/",
+            json={"name": "logs-empty-project"},
+        )
+        assert create_response.status_code == 200
+        project_id = create_response.json()["id"]
+
+        response = client.get(f"/logs/{project_id}")
+        assert response.status_code == 200
+        assert response.json() == []


dataloom-backend/app/api/endpoints/user_logs.py
```diff
@@ -4,12 +4,14 @@
 from sqlmodel import Session
 
 from app import database, models, schemas
+from app.api import dependencies
 
 router = APIRouter()
 
 
 @router.get("/{project_id}", response_model=list[schemas.LogResponse])
```
ivantha, line 12: **[LOW]** Extra Database Round-Trip on Every Request

Both `get_logs` and `get_last_checkpoint` now issue an extra `SELECT` query via `get_project_or_404` before the primary query. For `get_logs` this means two queries per request:
1. `SELECT * FROM projects WHERE id = ?` (existence check)
2. `SELECT * FROM project_change_logs WHERE project_id = ?` (actual data)

This pattern is consistent with the rest of the codebase (`get_project_details`, `save_project`, etc.) so it is architecturally coherent, but it is worth noting as a latency concern under load. At scale, the existence check could be folded into the primary query.

A more efficient alternative would be to do a single query and raise a 404 if no rows come back — but only when the semantic meaning of 'no rows' unambiguously implies 'project not found' (which is true for `get_last_checkpoint`, less clear for `get_logs` where a valid project with zero logs is a legitimate state).

**Suggestion:**
```
For `get_last_checkpoint`, the existence check and data fetch can be merged:
```python
# One query instead of two:
last_checkpoint = (
    db.query(models.Checkpoint)
    .filter(models.Checkpoint.project_id == project_id)
    .order_by(models.Checkpoint.created_at.desc())
    .first()
)
if last_checkpoint is None:
    # Differentiate 'project not found' from 'no checkpoints yet' by checking the project
    if not db.query(models.Project).filter(models.Project.id == project_id).first():
        raise HTTPException(status_code=404, detail=f"Project {project_id} not found")
```
For `get_logs`, the current two-query approach is the clearest option since an empty log list is valid.
```

Comment (ivantha) [inline dataloom-backend/app/api/endpoints/user_logs.py:12]: **[LOW]** Return Value of `get_project_or_404` Silently Discarded

`get_project_or_404` presumably returns the fetched `Project` object (its name and role in other endpoints implies this). Both call sites discard the return value entirely:
```python
dependencies.get_project_or_404(project_id, db)  # return value thrown away
```
This is not a bug, but it signals to future readers that the call is a pure side-effect check, which is non-obvious. It also means if future logic needs to use the project (e.g., to check permissions or enrich the response), the object will need to be fetched a second time.

Additionally, this uses a **direct function call** pattern rather than FastAPI's idiomatic `Depends()` injection system. The rest of the codebase likely uses `Depends(get_project_or_404)` as a path parameter dependency. Mixing both patterns in the same codebase reduces consistency.

**Suggestion:**
```
Capture the return value for clarity and future extensibility:
```python
project = dependencies.get_project_or_404(project_id, db)
```
Alternatively, align with the idiomatic FastAPI pattern used in other endpoints by injecting the dependency:
```python
from app.api.dependencies import get_project_or_404

@router.get("/{project_id}", response_model=list[schemas.LogResponse])
def get_logs(
    project_id: uuid.UUID,
    db: Session = Depends(database.get_db),
    _project: models.Project = Depends(get_project_or_404),  # validates existence
):
    ...
```
This makes the dependency graph explicit to FastAPI (useful for OpenAPI docs, testing overrides, etc.).
```

File: dataloom-backend/tests/test_new_features.py [hunk 2]
```diff
@@ -311,3 +311,18 @@ def test_delete_column_invalid_params(
...
+class TestLogsEndpoint:
+    def test_get_logs_nonexistent_project_returns_404(self, client):
+        response = client.get("/logs/00000000-0000-0000-0000-000000000000")
+        assert response.status_code == 404
+        assert "not found" in response.json()["detail"].lower()
+
+    def test_get_checkpoints_nonexistent_project_returns_404(self, client):
+        response = client.get("/logs/checkpoints/00000000-0000-0000-0000-000000000000")
+        assert response.status_code == 404
+        assert "not found" in response.json()["detail"].lower()
```
ivantha: **[LOW]** Tests Added to `test_new_features.py` — Wrong File for Regression Fixes

The new `TestLogsEndpoint` class is added to `test_new_features.py`. This filename implies tests for *new* features, while these tests are regression/bug-fix tests for corrected error-handling behavior. Over time, mixing bug-fix regressions into a 'new features' file makes it harder to find related tests and understand the test suite's structure.

If a dedicated `test_user_logs.py` exists (or should exist), these tests belong there alongside positive-path tests for the logs endpoints.

**Suggestion:**
```
Move `TestLogsEndpoint` to a dedicated file:
```
dataloom-backend/tests/test_user_logs.py
```
This file should include both the new 404 regression tests and existing/future positive-path tests (e.g., valid project with logs returns 200 with correct data shape).
```

ivantha: **[LOW]** Missing Happy-Path and Edge-Case Tests

The new test class only covers the 404 case (non-existent project). There are no tests verifying:
1. **Happy path**: A valid project with existing logs returns `200` with the correct response shape.
2. **Empty logs**: A valid project with *zero* logs returns `200` with `[]` (not 404).
3. **Checkpoint happy path**: A valid project with a checkpoint returns `200` with correct data.
4. **No checkpoints**: A valid project with no checkpoints returns the expected response (not 404).

Items 2 and 4 are particularly important to confirm that the `get_project_or_404` guard doesn't accidentally cause valid-but-empty states to return 404.

**Suggestion:**
```
Add tests for the valid-project scenarios:
```python
def test_get_logs_existing_project_returns_200(self, client, project_factory):
    project = project_factory()  # creates a valid project in test DB
    response = client.get(f"/logs/{project.id}")
    assert isinstance(response.json(), list)

def test_get_logs_existing_project_no_logs_returns_empty_list(self, client, project_factory):
    project = project_factory()
    response = client.get(f"/logs/{project.id}")
    assert response.json() == []
```
```

File: dataloom-backend/tests/test_new_features.py [hunk 3]
```diff
@@ -311,3 +311,18 @@ def test_delete_column_invalid_params(
...
+# --- Logs Endpoint Tests ---
+
+
+class TestLogsEndpoint:
+    def test_get_logs_nonexistent_project_returns_404(self, client):
+        response = client.get("/logs/00000000-0000-0000-0000-000000000000")
+        assert response.status_code == 404
+        assert "not found" in response.json()["detail"].lower()
+
+    def test_get_checkpoints_nonexistent_project_returns_404(self, client):
```
ivantha: **[INFO]** Hardcoded All-Zeros UUID Is Fragile If Sentinel Value Is Used Elsewhere

Both tests use `00000000-0000-0000-0000-000000000000` (the nil UUID) as the non-existent project ID. This is a common convention and works correctly here, but it carries a small risk: if another test or fixture happens to create a project with this exact ID (e.g., as a sentinel or default), the test would incorrectly pass a 200 instead of 404.

This is a low-probability concern but worth noting for test robustness.

**Suggestion:**
```
Generate a random UUID at test time to guarantee non-existence:
```python
import uuid

def test_get_logs_nonexistent_project_returns_404(self, client):
    nonexistent_id = uuid.uuid4()  # guaranteed to not exist
    response = client.get(f"/logs/{nonexistent_id}")
    assert response.status_code == 404
    assert "not found" in response.json()["detail"].lower()
```
```

File: dataloom-backend/tests/test_new_features.py [hunk 4]
```diff
@@ -311,3 +311,18 @@ def test_delete_column_invalid_params(
             },
         )
         assert response.status_code == expected_status
+
+
+# --- Logs Endpoint Tests ---
+
+
+class TestLogsEndpoint:
+    def test_get_logs_nonexistent_project_returns_404(self, client):
```
ivantha: **[INFO]** Error Message Assertion Uses Loose `lower()` String Match

```python
assert "not found" in response.json()["detail"].lower()
```
This assertion is intentionally flexible (case-insensitive substring match), which is a reasonable approach for avoiding brittle tests tied to exact error message wording. However, it provides weaker guarantees — it would pass for any 404 response from any middleware or route, not specifically from `get_project_or_404`. If the message ever changes structurally (e.g., becomes a list instead of a string), the test would raise an `AttributeError` rather than a clear assertion error.

**Suggestion:**
```
Add a minimal structural guard:
```python
detail = response.json()["detail"]
assert isinstance(detail, str), f"Expected detail to be a string, got: {type(detail)}"
assert "not found" in detail.lower()
```
Optionally, assert on the project ID being present in the detail for stronger specificity:
```python
assert str(project_id) in detail  # confirms it's about THIS project
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This is a clean, well-scoped bug fix that correctly addresses an inconsistency where two log endpoints silently returned HTTP 200 with empty data for non-existent project IDs, while every other project endpoint already enforced 404. The core change is minimal and correct — adding `get_project_or_404` calls at the top of both handlers is exactly the right pattern given the existing codebase conventions. The two regression tests are a good addition but could be expanded to cover happy-path and empty-but-valid states. The main concerns are architectural: the return value of `get_project_or_404` is discarded (making it a pure side-effect call), and direct function calls are used rather than FastAPI's idiomatic `Depends()` injection — both worth aligning for consistency. There is an implicit behavioral breaking change (200 → 404) that should be communicated to API consumers even though it corrects incorrect behavior.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral API Change: 200 → 404 for Non-Existent Projects on Logs Endpoints: **Previous behavior:** `GET /logs/{project_id}` returned `HTTP 200` with `[]` and `GET /logs/checkpoints/{project_id}` returned `HTTP 200` with an empty/null checkpoint response for any UUID that did not exist in the database.

**New behavior:** Both endpoints now return `HTTP 404 Not Found` with a structured error detail when the project does not exist.

**Who is affected:** Any API consumer (frontend, integration, or external client) that was silently handling the empty-list response instead of treating the missing project as an error. Code that does `if len(logs) == 0: show_empty_state()` will now receive a 404 exception instead and must handle it explicitly.

**Migration path:** Update all callers to handle 404 responses from these endpoints in addition to (or instead of) checking for empty lists. The fix correctly aligns behavior with all other project endpoints.

This is technically a breaking change but represents a **bug fix** — the old behavior was incorrect and inconsistent. No functional, well-written consumer should have been relying on the silent-200 behavior for non-existent resources.
  > **Suggestion:** No code change needed. Document this behavioral correction in a CHANGELOG entry and communicate it to frontend/consumer teams so they can add 404 error handling for these routes if not already present.


### PR Description Issues

- **[INFO]** PR Description Is High Quality — Minor Omissions Only: The PR description is well-structured with clearly separated **Problem** and **Fix** sections, explicit callout of the inconsistency pattern, and a screenshot demonstrating the after-fix behavior. The change count in the description ('Added 2 regression tests') matches the diff exactly.

Minor gaps:
1. **No 'Testing' section** describing how to run the new tests locally or what test command to execute.
2. **No mention of the extra DB query trade-off** introduced by adding `get_project_or_404` (each endpoint now does 2 DB queries instead of 1).
3. **No `CHANGELOG` / migration note** for API consumers who may have been relying on the silent-200 behavior, even though the description correctly frames it as a bug fix.
  > **Suggestion:** Consider adding:
```markdown
## How to Test
```bash
pytest dataloom-backend/tests/test_new_features.py::TestLogsEndpoint -v
```

## Notes for API Consumers
Clients relying on `GET /logs/{id}` or `GET /logs/checkpoints/{id}` returning `200 []` for non-existent projects must be updated to handle `404` responses. This aligns these endpoints with all other project routes.
```

dhruv1955:
@ivantha Addressed all feedback from the review:

- Captured return value as `_project = dependencies.get_project_or_404(...)` in both handlers - underscore prefix signals intentional side-effect guard while keeping the object available for future use
- Moved `TestLogsEndpoint` out of `test_new_features.py` into a dedicated `test_user_logs.py`
- Replaced hardcoded nil UUID with `uuid.uuid4()` in both 404 tests for guaranteed non-existence
- Added `assert isinstance(detail, str)` structural guard before the substring check
- Added happy-path tests: valid project returns `200` with a list, valid project with no logs returns `200 []`

ivantha:

dhruv1955:
@ivantha  

PR #187, opened 2026-03-12
test: add integration tests for recent, export, and delete project endpoints

dhruv1955:
## Problem
Three user-facing HTTP endpoints had zero integration test coverage:

| Endpoint | Handler |
|---|---|
| `GET /projects/recent` | `recent_projects` |
| `GET /projects/{id}/export` | `export_project` |
| `DELETE /projects/{id}` | `delete_project_endpoint` |

Error handling for non-existent project IDs on these endpoints was also unverified at the HTTP layer.

## Fix
Added `TestProjectEndpoints` class to `test_new_features.py` with 5 integration tests covering all 3 endpoints including both happy path and 404 error cases.

## Tests Added
- `test_recent_projects_returns_list` - uploads a project, asserts `GET /projects/recent` returns a non-empty list
- `test_export_project_returns_csv` - asserts export returns `200` with `text/csv` content-type
- `test_export_nonexistent_project_returns_404` - asserts export returns `404` for fake UUID
- `test_delete_project_returns_200` - asserts delete returns `200` and subsequent GET returns `404`
- `test_delete_nonexistent_project_returns_404` - asserts delete returns `404` for fake UUID

### DIFF
--- a/dataloom-backend/tests/test_new_features.py
+++ b/dataloom-backend/tests/test_new_features.py
@@ -1,6 +1,8 @@
 """Tests for new features: rename column, cast data type, export, and delete project."""
 
 import csv
+import io
+import uuid
 
 import pandas as pd
 import pytest
@@ -311,3 +313,60 @@ class TestTransformEndpoint:
             },
         )
         assert response.status_code == expected_status
+
+
+# --- Project Endpoint Integration Tests ---
+
+
+class TestProjectEndpoints:
+    def _upload_project(self, client, sample_csv, name="Test Project"):
+        with open(sample_csv, "rb") as f:
+            response = client.post(
+                "/projects/upload",
+                files={"file": ("test.csv", f, "text/csv")},
+                data={"projectName": name, "projectDescription": "Test"},
+            )
+        assert response.status_code == 200
+        return response.json()["project_id"]
+
+    def test_recent_projects_returns_list(self, client, sample_csv, db):
+        project_id = self._upload_project(client, sample_csv, name="Recent Test")
+        response = client.get("/projects/recent")
+        assert response.status_code == 200
+        projects = response.json()
+        assert isinstance(projects, list)
+        assert len(projects) >= 1
+        project_ids = [p["project_id"] for p in projects]
+        assert project_id in project_ids
+        client.delete(f"/projects/{project_id}")
+
+    def test_export_project_returns_csv(self, client, sample_csv, db):
+        import csv
+
+        project_id = self._upload_project(client, sample_csv, name="Export Test")
+        response = client.get(f"/projects/{project_id}/export")
+        assert response.status_code == 200
+        assert "text/csv" in response.headers["content-type"]
+        assert "attachment" in response.headers.get("content-disposition", "")
+        reader = csv.reader(io.StringIO(response.text))
+        rows = list(reader)
+        assert len(rows) >= 2
+        client.delete(f"/projects/{project_id}")
+
+    def test_export_nonexistent_project_returns_404(self, client, db):
+        fake_id = str(uuid.uuid4())
+        response = client.get(f"/projects/{fake_id}/export")
+        assert response.status_code == 404
+
+    def test_delete_project_returns_200(self, client, sample_csv, db):
+        project_id = self._upload_project(client, sample_csv, name="Delete Test")
+        response = client.delete(f"/projects/{project_id}")
+        assert response.status_code == 200
+        assert response.json() is not None
+        get_response = client.get(f"/projects/get/{project_id}")
+        assert get_response.status_code == 404
+
+    def test_delete_nonexistent_project_returns_404(self, client, db):
+        fake_id = str(uuid.uuid4())
+        response = client.delete(f"/projects/{fake_id}")
+        assert response.status_code == 404


ivantha (commented):
## ◎ Argus Review

This PR adds five well-structured integration tests covering three previously untested endpoints (`GET /projects/recent`, `GET /projects/{id}/export`, `DELETE /projects/{id}`). The overall approach is sound — using real uploads as prerequisites rather than mocked data gives genuine integration coverage, and the delete test's follow-up GET assertion is a particularly good pattern. The main issues are: (1) the two 404 tests inconsistently omit the `db` fixture used by every other test in the class, which could cause isolation failures when tests run standalone; (2) the upload setup block is copy-pasted three times and should be extracted into a shared fixture; and (3) the export test validates content-type but not the actual CSV body, leaving significant regression surface uncovered. No production code is changed, so there are no breaking changes.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes — pure additive test-only change: **Previous behavior**: The three endpoints (`GET /projects/recent`, `GET /projects/{id}/export`, `DELETE /projects/{id}`) had no integration test coverage.

**New behavior**: Five integration tests are added covering happy paths and 404 error paths for all three endpoints.

**Who is affected**: No production code, public APIs, exported symbols, or database schemas are modified. Only the test suite is affected.

**Migration path**: No migration required.
  > **Suggestion:** No action needed. `suggestedSemver` is `patch`.


### PR Description Issues

- **[LOW]** PR description is accurate and well-structured but omits fixture/infrastructure prerequisites: The PR description is genuinely good: it uses a clear problem/fix/tests-added structure, includes a table of affected endpoints, and accurately reflects what was implemented. However, it does not mention:
1. What fixtures (`client`, `sample_csv`, `db`) the tests depend on, or whether they already existed.
2. How to run only these tests locally (`pytest -k TestProjectEndpoints`).
3. Whether there is any prerequisite (e.g., a running database, environment variables) for the integration tests to pass.

This omission is minor but relevant for reviewers setting up the environment to validate the PR.
  > **Suggestion:** Add a small "Running the tests" section:
```markdown
## Running the Tests
These are integration tests and require the test database fixture to be available.

```bash
pytest dataloom-backend/tests/test_new_features.py::TestProjectEndpoints -v
```
All existing fixtures (`client`, `sample_csv`, `db`) are reused from `conftest.py` — no new infrastructure is required.
```

File: dataloom-backend/tests/test_new_features.py [hunk 1]
```diff
@@ -311,3 +311,67 @@ def test_delete_column_invalid_params(
...
+        project_id = response.json()["project_id"]
+
+        # Export the project
+        response = client.get(f"/projects/{project_id}/export")
+        assert response.status_code == 200
+        assert "text/csv" in response.headers["content-type"]
+
+    def test_export_nonexistent_project_returns_404(self, client):
+        response = client.get("/projects/00000000-0000-0000-0000-000000000000/export")
+        assert response.status_code == 404
```
ivantha, line 359: **[MEDIUM]** 404 tests omit `db` fixture — potential inconsistency with database state

Both `test_export_nonexistent_project_returns_404` and `test_delete_nonexistent_project_returns_404` do not accept the `db` fixture, while every other test in the class that interacts with the database does. If the `db` fixture is responsible for setting up or tearing down the database connection/schema (common in SQLAlchemy/FastAPI test setups), omitting it means these tests either rely on ambient state left by a previous test or could fail when run in isolation (e.g., `pytest tests/test_new_features.py::TestProjectEndpoints::test_export_nonexistent_project_returns_404`). Additionally, if the `db` fixture provides transactional rollback isolation, skipping it means any side-effects from the test session could leak into these tests.

**Suggestion:**
```
Add the `db` fixture to both 404 tests for consistency and safe isolation:
```python
def test_export_nonexistent_project_returns_404(self, client, db):
    response = client.get("/projects/00000000-0000-0000-0000-000000000000/export")
    assert response.status_code == 404

def test_delete_nonexistent_project_returns_404(self, client, db):
    response = client.delete("/projects/00000000-0000-0000-0000-000000000000")
    assert response.status_code == 404
```
```

File: dataloom-backend/tests/test_new_features.py [hunk 2]
```diff
@@ -311,3 +311,67 @@ def test_delete_column_invalid_params(
...
+        response = client.delete(f"/projects/{project_id}")
+        assert response.status_code == 200
+
+        # Verify project is gone
+        response = client.get(f"/projects/get/{project_id}")
+        assert response.status_code == 404
+
+    def test_delete_nonexistent_project_returns_404(self, client):
+        response = client.delete("/projects/00000000-0000-0000-0000-000000000000")
+        assert response.status_code == 404
```
ivantha, line 372: **[MEDIUM]** Duplicated project-upload setup block copied verbatim three times

The pattern of opening `sample_csv`, POSTing to `/projects/upload`, and asserting a `200` response is repeated identically in `test_recent_projects_returns_list`, `test_export_project_returns_csv`, and `test_delete_project_returns_200`. This violates DRY and creates a maintenance burden: if the upload endpoint changes (e.g., field names, URL, status code), three test methods must be updated in lockstep. It also obscures the intent of each test behind boilerplate.

**Suggestion:**
```
Extract the upload logic into a pytest fixture or a private helper method on the class:
```python
@pytest.fixture
def uploaded_project(self, client, sample_csv, db):
    """Uploads a CSV project and returns its project_id."""
        response = client.post(
            "/projects/upload",
            data={"projectName": "Test Project", "projectDescription": "Fixture project"},
        )
    return response.json()["project_id"]

def test_export_project_returns_csv(self, client, uploaded_project):
    response = client.get(f"/projects/{uploaded_project}/export")
    assert "text/csv" in response.headers["content-type"]
```
```

File: dataloom-backend/tests/test_new_features.py [hunk 3]
```diff
@@ -311,3 +311,67 @@ def test_delete_column_invalid_params(
...
+
+    def test_export_project_returns_csv(self, client, sample_csv, db):
+        # Upload a project first
+        with open(sample_csv, "rb") as f:
+            response = client.post(
+                "/projects/upload",
+                files={"file": ("test.csv", f, "text/csv")},
+                data={"projectName": "Test Export", "projectDescription": "Test export"},
+            )
+        assert response.status_code == 200
```
ivantha: **[MEDIUM]** Export test validates content-type but not CSV content or structure

The `test_export_project_returns_csv` test only checks that the HTTP status is 200 and that `text/csv` appears in the `Content-Type` header. It does not verify:
- That the response body is non-empty.
- That the CSV contains the expected headers/columns matching what was uploaded.
- That the row count matches the uploaded data.
- That the `Content-Disposition` header is set (important for browser download behavior).

A bug in the export handler that returns an empty CSV, a malformed CSV, or one with incorrect columns would not be caught by this test.

**Suggestion:**
```
Use Python's `csv` module to parse and validate the export body:
```python
import csv
import io

def test_export_project_returns_csv(self, client, sample_csv, db):
    # ... upload ...
    response = client.get(f"/projects/{project_id}/export")
    assert "text/csv" in response.headers["content-type"]
    # Validate non-empty CSV with headers
    reader = csv.reader(io.StringIO(response.text))
    rows = list(reader)
    assert len(rows) > 1, "Expected at least a header row and one data row"
    # Optionally check Content-Disposition
    assert "attachment" in response.headers.get("content-disposition", "")
```
```

File: dataloom-backend/tests/test_new_features.py (hunk 2 above)
Comment (ivantha) [inline dataloom-backend/tests/test_new_features.py:372]: **[LOW]** Hard-coded nil UUID for non-existent IDs — fragile and not semantically clear

Both 404 tests use the hard-coded UUID `00000000-0000-0000-0000-000000000000` (the nil UUID). While this is very unlikely to collide with a real project ID in a test database, it:
1. Could theoretically exist if any code path initializes IDs from zero bytes.
2. Does not communicate intent as clearly as generating a fresh random UUID.
3. Is repeated in two tests, creating another DRY issue.

Using `uuid.uuid4()` at test time is both more idiomatic and guaranteed to be random.

**Suggestion:**
```
```python
import uuid

def test_export_nonexistent_project_returns_404(self, client, db):
    fake_id = str(uuid.uuid4())
    response = client.get(f"/projects/{fake_id}/export")
    assert response.status_code == 404

def test_delete_nonexistent_project_returns_404(self, client, db):
    fake_id = str(uuid.uuid4())
    response = client.delete(f"/projects/{fake_id}")
    assert response.status_code == 404
```
```

Comment (ivantha) [inline dataloom-backend/tests/test_new_features.py:372]: **[LOW]** Delete test verifies project removal but not the response body content

`test_delete_project_returns_200` correctly performs a follow-up GET to confirm the project no longer exists — this is a good pattern. However, it does not validate the response body of the `DELETE` call itself. Many REST APIs return a confirmation message, a tombstone object, or metadata (e.g., `{"deleted": true, "project_id": "..."}`) on successful deletion. A regression where the handler returns `200` with an error payload embedded in the body would pass this test undetected.

**Suggestion:**
```
```python
# Delete the project
response = client.delete(f"/projects/{project_id}")
# Validate response body if the API returns structured data
body = response.json()
assert body.get("project_id") == project_id  # or whatever the API contract is
```
If the delete endpoint returns no body (204-style), consider whether `200` vs `204` is the correct status code — and add a test assertion accordingly.
```

File: dataloom-backend/tests/test_new_features.py [hunk 4]
```diff
@@ -311,3 +311,67 @@ def test_delete_column_invalid_params(
...
+        # Upload a project first
+        with open(sample_csv, "rb") as f:
+            response = client.post(
+                "/projects/upload",
+                files={"file": ("test.csv", f, "text/csv")},
+                data={"projectName": "Test Recent", "projectDescription": "Test recent projects"},
+            )
+        assert response.status_code == 200
+
+        # Get recent projects
```
ivantha: **[LOW]** `response.json()` called multiple times without caching the result

In `test_recent_projects_returns_list`, `response.json()` is called twice: once for `isinstance(response.json(), list)` and again for `len(response.json()) >= 1`. While the `TestClient` in FastAPI (backed by `httpx`) deserializes the JSON body each call, this is wasteful and slightly inconsistent — in some HTTP client implementations (or with streaming responses) this could fail on the second call.

**Suggestion:**
```
```python
projects = response.json()
assert isinstance(projects, list)
assert len(projects) >= 1
```
```

Comment (ivantha) [inline dataloom-backend/tests/test_new_features.py]: **[LOW]** Recent-projects test does not assert the uploaded project appears in the list

`test_recent_projects_returns_list` only verifies that the list is non-empty (`len >= 1`). It does not verify that the project just uploaded actually appears in the list. If other tests have already populated the database with older projects, the assertion would pass even if the `/projects/recent` endpoint is completely broken for new uploads. The test provides weak coverage of the actual recency semantics.

**Suggestion:**
```
Capture the `project_id` from the upload response and assert it appears in the recent list:
```python
project_id = response.json()["project_id"]

response = client.get("/projects/recent")
projects = response.json()
assert isinstance(projects, list)
project_ids = [p["project_id"] for p in projects]  # adjust key to actual schema
assert project_id in project_ids, "Newly uploaded project should appear in recent list"
```
```

File: dataloom-backend/tests/test_new_features.py (hunk 2 above)
Comment (ivantha) [inline dataloom-backend/tests/test_new_features.py:372]: **[LOW]** No test data cleanup — uploaded projects persist between test runs

Three tests upload projects to the database (`test_recent_projects_returns_list`, `test_export_project_returns_csv`, `test_delete_project_returns_200`), but only the delete test removes its project. The other two leave orphaned rows. If the `db` fixture does not use transactional rollback (i.e., it uses a real or persistent test database), repeated test runs will accumulate stale data. This can cause flakiness in count-sensitive assertions or hit storage limits in CI.

**Suggestion:**
```
If the `db` fixture does not already provide rollback isolation, add teardown to upload tests:
```python
def test_recent_projects_returns_list(self, client, sample_csv, db):
    # ... upload ...
    project_id = response.json()["project_id"]
    try:
        response = client.get("/projects/recent")
        # ... assertions ...
    finally:
        client.delete(f"/projects/{project_id}")
```
Alternatively, confirm that the `db` fixture already handles this via transaction rollback and document it explicitly.
```

File: dataloom-backend/tests/test_new_features.py (hunk 4 above)
Comment (ivantha) [inline dataloom-backend/tests/test_new_features.py]: **[LOW]** Missing tests for edge cases: pagination/limit on recent endpoint, invalid UUID format

The `GET /projects/recent` endpoint likely enforces a return limit (e.g., last 5 or 10 projects). There is no test verifying this contract. Similarly, passing a string that is not even a valid UUID (e.g., `/projects/not-a-uuid/export`) may return a `422 Unprocessable Entity` rather than a `404`. This difference in status codes is a real API contract worth testing, especially since it affects client error handling.

**Suggestion:**
```
Consider adding:
```python
def test_export_invalid_uuid_format_returns_422(self, client, db):
    response = client.get("/projects/not-a-valid-uuid/export")

def test_recent_projects_respects_limit(self, client, sample_csv, db):
    # Upload more than the expected limit
    for i in range(12):
            client.post("/projects/upload", ...)
    response = client.get("/projects/recent")
    assert len(response.json()) <= 10  # or whatever the documented limit is
```
```

File: dataloom-backend/tests/test_new_features.py (hunk 2 above)
Comment (ivantha) [inline dataloom-backend/tests/test_new_features.py:372]: **[INFO]** Good: Happy-path and error-path covered for each endpoint

Each of the three new endpoints has both a success case and a 404 error case tested. The delete test goes further by verifying the side-effect (subsequent GET returns 404), which is a strong integration assertion. The use of a real upload as a prerequisite rather than mocked IDs ensures the tests exercise the full request lifecycle. Test method names follow the `test_<action>_<outcome>` convention consistently, making the test suite easy to scan.

**Suggestion:**
```
No action needed — this is a positive observation. Consider this a baseline pattern to follow for future endpoint tests.
```

File: dataloom-backend/tests/test_new_features.py [hunk 5]
```diff
@@ -311,3 +311,67 @@ def test_delete_column_invalid_params(
             },
         )
         assert response.status_code == expected_status
```
ivantha, line 315: **[INFO]** Consider placing new test class in a dedicated file

`TestProjectEndpoints` is appended to `test_new_features.py`, which already has 311 lines of content. As the test suite grows, grouping tests by feature domain in separate files (e.g., `test_project_endpoints.py`) improves discoverability, reduces merge conflicts, and keeps individual files focused. This is not a blocker but worth discussing as the suite scales.

**Suggestion:**
```
Move `TestProjectEndpoints` to a new file `dataloom-backend/tests/test_project_endpoints.py`. Ensure shared fixtures (`client`, `sample_csv`, `db`) are in `conftest.py` so they remain accessible.
```

dhruv1955:
@ivantha Addressed all feedback from the review:

- Extracted repeated upload logic into `_upload_project()` helper method - eliminates the 3x duplicated setup block
- Added `db` fixture to both 404 tests (`test_export_nonexistent_project_returns_404`, `test_delete_nonexistent_project_returns_404`) for consistent isolation
- Replaced hardcoded nil UUID with `uuid.uuid4()` in both 404 tests
- Enhanced export test to validate CSV body (at least 2 rows) and assert `attachment` in `Content-Disposition` header
- Cached `response.json()` into `projects` variable in recent-projects test instead of calling it twice
- Assert uploaded `project_id` appears in the recent list - verifies recency semantics, not just non-empty
- Added inline cleanup (`client.delete`) after upload-dependent tests to avoid orphaned rows
- Assert `response.json() is not None` on delete response body

ivantha:

dhruv1955:
@ivantha

PR #189, opened 2026-03-13
add .db in git ignore

### LINKED ISSUE #190: [BUG] test.db file is getting commited
The test.db file in backend is getting updated which needs to mentioned in .gitignore file

schourasia750:
### while commiting test.db file is also getting commited which is not supposed to be

fixes: #190

### DIFF
--- a/dataloom-backend/.gitignore
+++ b/dataloom-backend/.gitignore
@@ -37,3 +37,4 @@ uploads/
 
 # SQLite (local dev)
 *.sqlite3
+*.db


ivantha (commented):
## ◎ Argus Review

This PR adds a `*.db` pattern to the `dataloom-backend/.gitignore` to prevent SQLite database files from being accidentally committed, addressing issue #190. The change itself is correct and minimal, but the PR is **incomplete**: adding to `.gitignore` does not untrack files already committed to git history — if `test.db` was previously committed, it will continue to be tracked until `git rm --cached` is run. This is the most important follow-up action. No breaking changes are introduced; this is a pure developer tooling fix.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes detected: **Previous behavior**: `.db` files were not gitignored and could be (and were being) committed.

**New behavior**: All files matching `*.db` in `dataloom-backend/` will be ignored by git going forward.

**Who is affected**: Only developers working on the backend locally. No runtime, API, or consumer-facing behavior is changed.

**Migration path**: N/A — this is a developer tooling improvement only. No consuming code changes are required.
  > **Suggestion:** No action needed for breaking changes. Semver impact: `patch`.


### PR Description Issues

- **[INFO]** PR description lacks testing/verification details: The description correctly identifies the problem (`test.db` being committed unintentionally) and references issue #190. However, it is missing key context:

- **No verification step**: Adding a pattern to `.gitignore` only prevents *future* commits of matching files. If `test.db` (or any other `.db` file) was **already tracked** by git, it will continue to appear in commits until explicitly removed with `git rm --cached <file>`. The description doesn't confirm whether this step was taken.
- **No testing approach**: No mention of how the author verified the fix works (e.g., running `git status` after the change, confirming `.db` files are no longer staged).
- **Scope not explained**: The fix is broader than stated — it silences *all* `*.db` files, not just `test.db`. This is the right call, but the description only calls out one file.
  > **Suggestion:** A stronger description would be:

```
### Problem
SQLite database files (e.g., `test.db`) were being accidentally committed to the repository during local development. These are runtime artifacts and should never be tracked.

### Changes
- Added `*.db` glob pattern to `dataloom-backend/.gitignore` to cover all SQLite `.db` files (complementing the existing `*.sqlite3` rule).

### Verification
- Ran `git rm --cached *.db` to untrack any previously committed `.db` files.
- Confirmed `git status` no longer shows `.db` files after the change.

Fixes: #190
```

dataloom-backend/.gitignore
```diff
@@ -37,3 +37,4 @@ uploads/
 
 # SQLite (local dev)
 *.sqlite3
+*.db
```
ivantha, line 40: **[HIGH]** Already-tracked `.db` files will not be ignored without manual untracking

Adding a pattern to `.gitignore` **only prevents untracked files from being added** in the future. If any `.db` file (such as `test.db`, which the PR description mentions) was **already committed to the repository**, git will continue to track it and include it in future diffs — the new `.gitignore` rule will have no effect on it.

This is a very common misconception and the fix is incomplete without the corresponding cache-removal step. Reviewers should confirm:
1. Was `test.db` (or any `*.db` file) previously committed to this repository?
2. If yes, was `git rm --cached` run to stop tracking it?

Without this, the stated goal of the PR (issue #190) may not actually be resolved.

**Suggestion:**
```
After merging this `.gitignore` change, run the following to stop tracking any currently-committed `.db` files:

```bash
# From the dataloom-backend directory
git rm --cached *.db
# Or recursively if they exist in subdirectories:
git rm --cached -r --ignore-unmatch '*.db'
git commit -m "chore: stop tracking .db files"
```

This should be included as part of this PR or as an immediate follow-up commit.
```

Comment (ivantha) [inline dataloom-backend/.gitignore:40]: **[LOW]** Consider consolidating SQLite ignore patterns under a single comment

The current block reads:
```
# SQLite (local dev)
*.sqlite3
*.db
```
This is functionally correct, but `.db` is a generic extension used by many database engines beyond SQLite (e.g., Berkeley DB, Realm, generic binary DBs). A more explicit comment helps future contributors understand what is being ignored and why.

**Suggestion:**
```
```gitignore
# SQLite / local database files (not to be committed)
*.sqlite3
*.db
```
This makes the intent clearer and reduces future confusion if new database types are introduced.
```

Comment (ivantha) [inline dataloom-backend/.gitignore:40]: **[INFO]** Pattern coverage is good and appropriately broad

Adding `*.db` is the correct and idiomatic approach here. SQLite databases can use either `.sqlite3` or `.db` extensions (and occasionally `.sqlite`), so having both patterns ensures comprehensive coverage. The placement beneath the existing `# SQLite (local dev)` comment is logical and consistent with the surrounding file structure. This is a well-scoped, minimal change.

**Suggestion:**
```
Optionally also add `*.sqlite` to cover all three common SQLite file extensions:
```gitignore
# SQLite / local database files (not to be committed)
*.sqlite
*.sqlite3
*.db
```
```

ivantha:

PR #192, opened 2026-03-13
fix(backend): correct transform HTTP status mapping and validate addCol name

### LINKED ISSUE #191: Fix transform endpoint status codes and validate addCol name
I found two related problems while working on the transform flow.

## What is wrong

1. **Wrong status code for internal failures**
- Endpoint: `POST /projects/{project_id}/transform`
- Sometimes unexpected server errors are being returned as `400`.
- `400` should mean bad user input. If the server crashes unexpectedly, it should be `500`.
- This makes debugging harder because client mistakes and server bugs look the same.

2. **`addCol` allows missing column name**
- `add_col_params` can come without `name`.
- That can create a `NaN` column name and fail later during response validation.
- This should fail early at request validation time with `422`.

## What I suggest

- Keep explicit `HTTPException` status codes as they are.
- Keep `TransformationError` mapped to `400`.
- Map unexpected exceptions to `500` with a safe message: `"Internal server error"`.
- Make `add_col_params.name` required and reject blank values.

## Done when

- Missing filter params -> `400`
- Invalid transform input -> `400`
- Unexpected error in transform execution -> `500` with generic detail
- Unexpected error in save/log step -> `500` with generic detail
- `addCol` without `name` -> `422`
- `addCol` with valid `name` -> `200`

## Tests to add/update

- HTTP semantics test for explicit `HTTPException` passthrough
- HTTP semantics test for `TransformationError -> 400`
- HTTP semantics test for unexpected error in transform step -> `500`
- HTTP semantics test for unexpected error in persistence step -> `500`
- Existing add/delete column tests should still pass with the name validation change

nodesagar:
Fixes #191

## What this PR fixes

I fixed two backend issues in the transform flow:

1. `POST /projects/{project_id}/transform` was returning `400` for unexpected internal failures in some paths.
2. `addCol` could be called without `add_col_params.name`, which later caused response validation errors (`NaN` column-name path) instead of failing early.

## Changes made

- Updated transform endpoint error handling in `app/api/endpoints/transformations.py`:
  - keeps explicit `HTTPException` status codes as-is
  - maps `TransformationError` to `400`
  - maps unexpected exceptions to `500` with safe detail: `"Internal server error"`
- Wrapped transform + persistence + response construction in one guarded flow so unexpected persistence failures are also handled as `500`.
- Enforced `add_col_params.name` as required and non-blank in `app/schemas.py`.
- Added regression tests in `tests/test_transform_http_semantics.py` for:
  - HTTPException passthrough
  - TransformationError -> 400
  - unexpected exception during transform execution -> 500
  - unexpected exception during persistence step -> 500

## Validation

I ran the full project checks:

- Backend: `uv run ruff check .` ✅
- Backend: `uv run pytest -q` ✅ (`97 passed`)
- Backend: `uv run ruff format .` ✅
- Frontend: `npm run lint` ✅
- Frontend: `npm run test` ✅ (`48 passed`)
- Frontend: `npm run format` ✅

## Impact

- Better API semantics for clients (input errors vs server errors are now clearly separated).
- `addCol` invalid payloads fail fast with `422` instead of failing later.
- Added focused tests to prevent regressions in this area.

### DIFF
--- a/dataloom-backend/app/api/endpoints/transformations.py
+++ b/dataloom-backend/app/api/endpoints/transformations.py
@@ -3,6 +3,7 @@
 All transformations are handled through a single unified /transform endpoint.
 """
 
+import re
 import uuid
 
 from fastapi import APIRouter, Depends, HTTPException
@@ -20,6 +21,61 @@ logger = get_logger(__name__)
 router = APIRouter()
 
 COMPLEX_OPERATIONS = {"dropDuplicate", "advQueryFilter", "pivotTables", "dropNa", "melt", "groupby"}
+SAFE_TRANSFORMATION_ERROR_DETAIL = "Invalid transformation request"
+
+_SENSITIVE_TOKEN_MARKERS = (
+    "traceback",
+    "sqlalchemy",
+    "psycopg",
+    "sqlite",
+    "postgres",
+    "password",
+    "secret",
+    "token",
+)
+_SQL_MARKERS = ("select ", "insert ", "update ", "delete ", "drop ", "alter ", "create table", " from ", " where ")
+
+
+def _safe_transformation_error_detail(error: Exception) -> str:
+    """Return a client-safe 400 detail for domain transformation failures."""
+    detail = str(error).strip()
+    if not detail:
+        return SAFE_TRANSFORMATION_ERROR_DETAIL
+
+    lowered = detail.lower()
+
+    if "\n" in detail or "\r" in detail:
+        return SAFE_TRANSFORMATION_ERROR_DETAIL
+
+    if any(token in lowered for token in _SENSITIVE_TOKEN_MARKERS):
+        return SAFE_TRANSFORMATION_ERROR_DETAIL
+
+    # Redact SQL-like payloads only when multiple SQL markers co-occur.
+    if sum(marker in lowered for marker in _SQL_MARKERS) >= 2:
+        return SAFE_TRANSFORMATION_ERROR_DETAIL
+
+    # Redact likely filesystem paths that should not be exposed to clients.
+    if re.search(r"[A-Za-z]:\\[^\\\n]+", detail) or re.search(r"/(?:[^/\n]+/)+[^/\n]+", detail):
+        return SAFE_TRANSFORMATION_ERROR_DETAIL
+    return detail
+
+
+def _safe_http_exception_detail(error: HTTPException) -> str | None:
+    """Return a redacted detail for sensitive HTTPException payloads."""
+    detail = error.detail if isinstance(error.detail, str) else ""
+    lowered = detail.lower()
+
+    if error.status_code >= 500:
+        return "Internal server error"
+
+    # Utility-layer CSV 404s may embed absolute paths.
+    if error.status_code == 404 and "csv file not found" in lowered:
+        return "CSV file not found"
+
+    if detail and (re.search(r"[A-Za-z]:\\[^\\\n]+", detail) or re.search(r"/(?:[^/\n]+/)+[^/\n]+", detail)):
+        return "Resource not found" if error.status_code == 404 else "Internal server error"
+
+    return None
 
 
 def _handle_basic_transform(df, transformation_input, project, db, project_id):
@@ -158,27 +214,58 @@ async def transform_project(
     Routes to the appropriate internal handler based on operation_type.
     """
     project = get_project_or_404(project_id, db)
-    df = read_csv_safe(project.file_path)
 
-    op = transformation_input.operation_type
+    # Keep an explicit local for consistency across dispatch, persistence and
+    # logging paths. Use a defensive fallback so exception logging never
+    # introduces a secondary NameError.
+    operation_type = getattr(transformation_input, "operation_type", "<unknown>")
 
     try:
-        if op in COMPLEX_OPERATIONS:
+        df = read_csv_safe(project.file_path)
+
+        if operation_type in COMPLEX_OPERATIONS:
             result_df, should_save = _handle_complex_transform(df, transformation_input, project, db, project_id)
         else:
             result_df, should_save = _handle_basic_transform(df, transformation_input, project, db, project_id)
+
+        if should_save:
+            save_csv_safe(result_df, project.file_path)
+            try:
+                log_transformation(db, project_id, operation_type, transformation_input.dict())
+            except Exception:
+                # Compensate disk mutation if audit logging fails to avoid a
+                # partially persisted state (file changed without log entry).
+                try:
+                    save_csv_safe(df, project.file_path)
+                except Exception:
+                    logger.exception(
+                        "Failed to restore project file after log_transformation failure for project_id=%s op=%s",
+                        project_id,
+                        operation_type,
+                    )
+                raise
+
+        resp = dataframe_to_response(result_df)
+        return {
+            "project_id": project_id,
+            "operation_type": operation_type,
+            **resp,
+        }
+    except HTTPException as e:
+        safe_detail = _safe_http_exception_detail(e)
+        if safe_detail is None:
+            # Preserve explicit HTTP errors (e.g., missing parameters) and their status codes.
+            raise
+
+        logger.warning(
+            "Redacted HTTPException detail during transform for project_id=%s op=%s status=%s",
+            project_id,
+            operation_type,
+            e.status_code,
+        )
+        raise HTTPException(status_code=e.status_code, detail=safe_detail) from e
     except ts.TransformationError as e:
-        raise HTTPException(status_code=400, detail=str(e)) from e
+        raise HTTPException(status_code=400, detail=_safe_transformation_error_detail(e)) from e
     except Exception as e:
-        raise HTTPException(status_code=400, detail=str(e)) from e
-
-    if should_save:
-        save_csv_safe(result_df, project.file_path)
-        log_transformation(db, project_id, transformation_input.operation_type, transformation_input.dict())
-
-    resp = dataframe_to_response(result_df)
-    return {
-        "project_id": project_id,
-        "operation_type": transformation_input.operation_type,
-        **resp,
-    }
+        logger.exception("Unexpected error during transform for project_id=%s op=%s", project_id, operation_type)
+        raise HTTPException(status_code=500, detail="Internal server error") from e
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -112,11 +112,18 @@ class AddColumn(BaseModel):
 
     Attributes:
         index: Zero-based column index where column will be inserted.
-        name: Column name; required for add operations.
+        name: Column name (required, non-blank).
     """
 
     index: int
     name: str
+
+    @field_validator("name")
+    @classmethod
+    def name_must_not_be_blank(cls, v: str) -> str:
+        if not v.strip():
+            raise ValueError("Column name cannot be empty or whitespace")
+        return v
 
 
 class DeleteColumn(BaseModel):
--- a/dataloom-backend/tests/test_transform_http_semantics.py
+++ b/dataloom-backend/tests/test_transform_http_semantics.py
@@ -1,0 +1,202 @@
+"""HTTP semantics tests for the transform endpoint."""
+
+import pytest
+from fastapi import HTTPException
+
+from app.utils.pandas_helpers import read_csv_safe
+
+
+@pytest.fixture
+def project(client, sample_csv):
+    with open(sample_csv, "rb") as f:
+        response = client.post(
+            "/projects/upload",
+            files={"file": ("test.csv", f, "text/csv")},
+            data={"projectName": "HTTP Semantics", "projectDescription": "fixture"},
+        )
+    assert response.status_code == 200, response.text
+    return response.json()
+
+
+@pytest.fixture
+def project_id(project):
+    return project["project_id"]
+
+
+def test_transform_preserves_http_exception_status(client, project_id):
+    # Missing filter params should return the explicit 400 from endpoint validation.
+    response = client.post(
+        f"/projects/{project_id}/transform",
+        json={"operation_type": "filter"},
+    )
+
+    assert response.status_code == 400
+    assert response.json()["detail"] == "Filter parameters required"
+
+
+def test_transform_maps_transformation_error_to_400(client, project_id):
+    response = client.post(
+        f"/projects/{project_id}/transform",
+        json={
+            "operation_type": "filter",
+            "parameters": {"column": "missing_column", "condition": "=", "value": "Alice"},
+        },
+    )
+
+    assert response.status_code == 400
+    assert "not found" in response.json()["detail"].lower()
+
+
+def test_transform_redacts_csv_not_found_path_detail(client, project_id, monkeypatch):
+    from app.api.endpoints import transformations as transformations_endpoint
+
+    def boom(*args, **kwargs):
+        raise HTTPException(status_code=404, detail="CSV file not found: /tmp/private/uploads/project.csv")
+
+    monkeypatch.setattr(transformations_endpoint, "read_csv_safe", boom)
+
+    response = client.post(
+        f"/projects/{project_id}/transform",
+        json={
+            "operation_type": "filter",
+            "parameters": {"column": "name", "condition": "=", "value": "Alice"},
+        },
+    )
+
+    assert response.status_code == 404
+    assert response.json()["detail"] == "CSV file not found"
+
+
+def test_transform_redacts_internal_http_exception_detail(client, project_id, monkeypatch):
+    from app.api.endpoints import transformations as transformations_endpoint
+
+    def boom(*args, **kwargs):
+        raise HTTPException(
+            status_code=500,
+            detail="Error reading CSV: [Errno 13] Permission denied: /tmp/private/uploads/project.csv",
+        )
+
+    monkeypatch.setattr(transformations_endpoint, "read_csv_safe", boom)
+
+    response = client.post(
+        f"/projects/{project_id}/transform",
+        json={
+            "operation_type": "filter",
+            "parameters": {"column": "name", "condition": "=", "value": "Alice"},
+        },
+    )
+
+    assert response.status_code == 500
+    assert response.json()["detail"] == "Internal server error"
+
+
+def test_transform_redacts_sensitive_transformation_error_message(client, project_id, monkeypatch):
+    from app.api.endpoints import transformations as transformations_endpoint
+
+    def boom(*args, **kwargs):
+        raise transformations_endpoint.ts.TransformationError(
+            "SQL failure: SELECT * FROM users WHERE password = 'secret'"
+        )
+
+    monkeypatch.setattr(transformations_endpoint.ts, "apply_filter", boom)
+
+    response = client.post(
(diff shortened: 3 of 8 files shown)


nodesagar:
@ivantha  this PR fixes #191 and closes it on merge, would really appreciate your review.

dataloom-backend/app/api/endpoints/transformations.py
```diff
@@ -156,18 +156,22 @@ async def transform_project(
...
+        resp = dataframe_to_response(result_df)
+        return {
+            "project_id": project_id,
+            "operation_type": transformation_input.operation_type,
+            **resp,
+        }
+    except HTTPException:
+        # Preserve explicit HTTP errors (e.g., missing parameters) and their status codes.
+        raise
     except ts.TransformationError as e:
```
ivantha, line 267: **[HIGH]** `op` variable referenced in logger may be undefined

The `except Exception` handler logs:
```python
logger.exception("Unexpected error during transform for project_id=%s op=%s", project_id, op)
```
The variable `op` is not defined anywhere in the visible diff. The success-path `return` dict uses `transformation_input.operation_type` directly — there is no visible `op = ...` assignment.

If `op` is not assigned earlier in the function body (outside the diff window), this will raise a `NameError` **inside the exception handler itself**, which will shadow the original exception and produce a confusing 500 with a misleading traceback. Even if it is defined, the inconsistency with `transformation_input.operation_type` used everywhere else in the diff is a code smell.

This is also testable: the existing 500 tests use a generic `RuntimeError`, so they would still pass even if `op` were undefined — the `NameError` would itself propagate as an unhandled exception at the ASGI layer, potentially producing a different response shape.

**Suggestion:**
```
Either confirm `op` is assigned earlier in the function, or use a safe fallback:
```python
except Exception as e:
    _op = getattr(transformation_input, "operation_type", "<unknown>")
    logger.exception(
        "Unexpected error during transform for project_id=%s op=%s",
        project_id,
        _op,
    )
    raise HTTPException(status_code=500, detail="Internal server error") from e
```
```

File: dataloom-backend/app/api/endpoints/transformations.py [hunk 2]
```diff
@@ -156,18 +156,22 @@ async def transform_project(
             result_df, should_save = _handle_complex_transform(df, transformation_input, project, db, project_id)
         else:
             result_df, should_save = _handle_basic_transform(df, transformation_input, project, db, project_id)
+
+        if should_save:
+            save_csv_safe(result_df, project.file_path)
+            log_transformation(db, project_id, transformation_input.operation_type, transformation_input.dict())
+
```
ivantha, line 247: **[MEDIUM]** Partial persistence on `log_transformation` failure leaves data in inconsistent state

The save and log steps are now inside the `try` block but executed sequentially without atomicity:
```python
if should_save:
    save_csv_safe(result_df, project.file_path)   # (1) file mutated on disk
    log_transformation(db, project_id, ...)        # (2) if this raises, file is already saved
```
If `save_csv_safe` succeeds but `log_transformation` raises (e.g. DB connection drops), the code falls to `except Exception` and returns HTTP 500. The file on disk has been mutated but the audit log has no record of the change. An API client receiving 500 may retry the operation, leading to a duplicate mutation.

Separating infallible concerns (disk write) from fallible ones (DB write) without compensating logic makes the system non-idempotent under partial failure.

**Suggestion:**
```
Wrap both in a DB transaction or, at minimum, log the error and decide whether to surface it as 500 or treat it as a non-fatal warning:
```python
if should_save:
    save_csv_safe(result_df, project.file_path)
    try:
        log_transformation(db, project_id, transformation_input.operation_type, transformation_input.dict())
    except Exception:
        logger.exception("Failed to log transformation for project_id=%s; file was saved.", project_id)
        # Either raise or swallow depending on audit requirements
```
Alternatively, record the log entry first (as a pending entry) and only write the file after committing.
```

File: dataloom-backend/tests/test_transform_http_semantics.py [hunk 3]
```diff
@@ -0,0 +1,86 @@
...
+    response = client.post(
+        f"/projects/{project_id}/transform",
+        json={
+            "operation_type": "addRow",
+            "row_params": {"index": 0},
+        },
+    )
+
+    assert response.status_code == 500
+    assert response.json()["detail"] == "Internal server error"
```
ivantha, line 174: **[MEDIUM]** No test for the `addCol` name validation change introduced in `schemas.py`

The PR adds a field validator that enforces `AddColumn.name` is required and non-blank, but the new test file contains **zero tests** for this change. The four tests cover only the HTTP semantics of the transform endpoint dispatcher.

This means:
1. The regression path the PR description specifically calls out (`addCol` with missing `name` causing downstream `NaN` path errors) is entirely untested.
2. Edge cases for the validator (empty string, whitespace-only string, `None` explicitly passed) are not covered.
3. Future refactors to `AddColumn` could silently regress without test feedback.

**Suggestion:**
```
Add tests in `tests/test_transform_http_semantics.py` or a dedicated schema/endpoint test file:
```python
def test_add_col_without_name_returns_422(client, sample_csv):
    project_id = _upload_project(client, sample_csv)
    response = client.post(
        f"/projects/{project_id}/transform",
        json={"operation_type": "addCol", "add_col_params": {"index": 0}},
    )

def test_add_col_with_blank_name_returns_422(client, sample_csv):
    project_id = _upload_project(client, sample_csv)
    response = client.post(
        f"/projects/{project_id}/transform",
        json={"operation_type": "addCol", "add_col_params": {"index": 0, "name": "   "}},
    )

def test_add_col_with_valid_name_succeeds(client, sample_csv):
    project_id = _upload_project(client, sample_csv)
    response = client.post(
        f"/projects/{project_id}/transform",
        json={"operation_type": "addCol", "add_col_params": {"index": 0, "name": "new_col"}},
    )
```
```

File: dataloom-backend/app/schemas.py [hunk 4]
```diff
@@ -113,7 +113,14 @@ class AddColumn(BaseModel):
...
 
     index: int
-    name: str | None = None
+    name: str
+
+    @field_validator("name")
+    @classmethod
+    def name_must_not_be_blank(cls, v):
+        if not v or not v.strip():
+            raise ValueError("name must not be empty")
```
ivantha: **[MEDIUM]** `AddColumn.name` changed from optional to required — breaking API change

**Previous behavior**: `name: str | None = None` — callers could omit `name` or pass `null`; the field defaulted to `None`.

**New behavior**: `name: str` (required, non-blank) — omitting `name` or passing `null` now causes Pydantic to reject the request with HTTP 422.

**Who is affected**: Any existing API client that calls `POST /projects/{project_id}/transform` with `operation_type: addCol` without supplying a non-blank `name` field. This includes any frontend code, scripts, or third-party integrations that relied on the old optional behaviour — even if that previous behaviour resulted in a downstream error, clients may have been handling it.

**Migration path**: All callers must update payloads to include a non-empty `name` in `add_col_params`. The PR description acknowledges this ("fail fast with 422 instead of failing later") but does not explicitly flag it as a breaking change for downstream consumers.

**Suggestion:**
```
If this is a public or client-consumed API, document the change in the changelog/release notes and bump the API version or add a deprecation notice. If the field genuinely never worked without a name, add a code comment making that intent clear:
```python
# Previously optional but functionally required; validated explicitly to fail fast.
name: str
```
```

File: dataloom-backend/app/schemas.py [hunk 5]
```diff
@@ -113,7 +113,14 @@ class AddColumn(BaseModel):
...
+        return v
```
ivantha, line 126: **[LOW]** Validator accepts names with leading/trailing whitespace; `not v` check is redundant

Two minor issues in the validator:

1. **Whitespace not stripped on return**: The validator correctly rejects a whitespace-only string (`"   "`), but it passes a string like `" my col "` through unchanged. Whether trailing/leading whitespace is intentional is not documented. If the downstream code uses the name in a column path or file path, having unexpected whitespace could still cause subtle issues.

2. **`not v` is redundant**: Since `name` is typed as `str` (not `str | None`), Pydantic v2 will raise a `ValidationError` before the validator is invoked if the value cannot be coerced to a string. An empty string `""` is falsy (`not v` is True), which is the only case this branch catches — the comment would be clearer if it only checked `not v.strip()`.

**Suggestion:**
```
```python
@field_validator("name")
@classmethod
def name_must_not_be_blank(cls, v: str) -> str:
    if not v.strip():
        raise ValueError("name must not be empty or blank")
    return v.strip()  # optionally normalise leading/trailing whitespace
```
```

File: dataloom-backend/app/api/endpoints/transformations.py [hunk 6]
```diff
@@ -156,18 +156,22 @@ async def transform_project(
...
+        resp = dataframe_to_response(result_df)
+        return {
+            "project_id": project_id,
+            "operation_type": transformation_input.operation_type,
+            **resp,
+        }
+    except HTTPException:
```
ivantha: **[LOW]** `detail=str(e)` for `TransformationError` may leak internal information

```python
except ts.TransformationError as e:
    raise HTTPException(status_code=400, detail=str(e)) from e
```
Unlike the `except Exception` branch (which deliberately uses the safe string `"Internal server error"`), `TransformationError` messages are surfaced verbatim to the API caller. While `TransformationError` is a domain error and its messages are likely user-friendly, the distinction is not enforced. If a future developer creates a `TransformationError` with a message that includes an internal path, SQL fragment, or credentials, it will be exposed.

This pre-existed the PR but the PR makes the contrast more visible by hardening the `Exception` branch.

**Suggestion:**
```
Either document that `TransformationError` messages are safe for external consumption (in the class definition), or use a sanitisation layer:
```python
except ts.TransformationError as e:
    # TransformationError messages are intentionally user-facing; do not include
    # internal state in subclass messages.
    raise HTTPException(status_code=400, detail=str(e)) from e
```
```

File: dataloom-backend/tests/test_transform_http_semantics.py [hunk 7]
```diff
@@ -0,0 +1,86 @@
...
+
+
+def _upload_project(client, sample_csv):
+    with open(sample_csv, "rb") as f:
+        response = client.post(
+            "/projects/upload",
+            files={"file": ("test.csv", f, "text/csv")},
+            data={"projectName": "HTTP Semantics", "projectDescription": "fixture"},
+        )
+    assert response.status_code == 200, response.text
```
ivantha, line 17: **[LOW]** `_upload_project` helper should be a pytest fixture

The `_upload_project` function is called in every test case and creates a project as a side effect. As a plain function rather than a fixture it:
1. Does not benefit from pytest's scope control (function/module/session), leading to redundant uploads.
2. Cannot perform teardown (projects created during tests are never cleaned up).
3. Cannot be composed with other fixtures automatically.
4. Is harder to mock or override in derived test classes.

The pattern of calling `_upload_project(client, sample_csv)` in every test body is repetitive.

**Suggestion:**
```
```python
import pytest

@pytest.fixture()
def project_id(client, sample_csv):
        response = client.post(
            "/projects/upload",
            data={"projectName": "HTTP Semantics", "projectDescription": "fixture"},
        )
    assert response.status_code == 200, response.text
    pid = response.json()["project_id"]
    yield pid
    # teardown: client.delete(f"/projects/{pid}")

def test_transform_preserves_http_exception_status(client, project_id):
    response = client.post(f"/projects/{project_id}/transform", json={"operation_type": "filter"})
    assert response.status_code == 400
    ...
```
```

File: dataloom-backend/tests/test_transform_http_semantics.py (hunk 3 above)
Comment (ivantha) [inline dataloom-backend/tests/test_transform_http_semantics.py:174]: **[LOW]** Persistence-failure test relies on opaque internal `should_save=True` assumption

```python
# This path runs only for mutating operations (should_save=True).
monkeypatch.setattr(transformations_endpoint, "save_csv_safe", boom)

response = client.post(
    f"/projects/{project_id}/transform",
    json={"operation_type": "addRow", "row_params": {"index": 0}},
)
```
The test implicitly relies on `addRow` setting `should_save=True` internally. If the `_handle_basic_transform` / `_handle_complex_transform` logic changes so that `addRow` stops triggering a save, this test will pass for the wrong reason — the monkeypatch will never fire, `save_csv_safe` won't be called, and the test will still see 200 rather than 500, causing a false negative.

The comment partially acknowledges this but doesn't guard against it.

**Suggestion:**
```
Assert that `save_csv_safe` was actually called using `monkeypatch` + a call counter, or add a guard assertion:
```python
called = []
def boom(*args, **kwargs):
    called.append(True)
    raise RuntimeError("disk error")

monkeypatch.setattr(transformations_endpoint, "save_csv_safe", boom)
...
assert response.status_code == 500
assert called, "save_csv_safe was never invoked — test may not be exercising the persistence path"
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR correctly addresses two real bugs: incorrect HTTP 500→400 status mapping for internal errors, and missing validation for `addCol.name`. The refactoring to move the persistence+response block inside the `try` scope is well-reasoned and the `HTTPException` passthrough pattern is correct. However, there are notable concerns: (1) the `op` variable referenced in the new logger call is not visible in the diff and may be undefined — if so, it would cause a `NameError` inside the exception handler, shadowing the original error; (2) the `save_csv_safe` + `log_transformation` split can leave the system in an inconsistent state under partial failure; (3) the new test file provides zero coverage for the `AddColumn.name` validation change that is a key motivation of this PR. The schema change making `name` required is a behavioural breaking change that is not flagged as such in the PR description.

### ⚠️ Breaking Changes

- **[MEDIUM]** `AddColumn.name` changed from optional to required — breaking API change: **Previous behavior**: `name: str | None = None` — callers could omit `name` or pass `null`; the field defaulted to `None`.

**New behavior**: `name: str` (required, non-blank) — omitting `name` or passing `null` now causes Pydantic to reject the request with HTTP 422.

**Who is affected**: Any existing API client that calls `POST /projects/{project_id}/transform` with `operation_type: addCol` without supplying a non-blank `name` field. This includes any frontend code, scripts, or third-party integrations that relied on the old optional behaviour — even if that previous behaviour resulted in a downstream error, clients may have been handling it.

**Migration path**: All callers must update payloads to include a non-empty `name` in `add_col_params`. The PR description acknowledges this ("fail fast with 422 instead of failing later") but does not explicitly flag it as a breaking change for downstream consumers.
  > **Suggestion:** If this is a public or client-consumed API, document the change in the changelog/release notes and bump the API version or add a deprecation notice. If the field genuinely never worked without a name, add a code comment making that intent clear:
```python
# Previously optional but functionally required; validated explicitly to fail fast.
name: str
```
- **[INFO]** HTTP 500 for unexpected errors is a behavioral change for existing API clients: **Previous behavior**: All unhandled exceptions (both domain errors and internal crashes) were mapped to HTTP 400 with `detail=str(e)`, leaking the exception message.

**New behavior**: Unhandled `Exception` (not `HTTPException`, not `TransformationError`) now maps to HTTP 500 with `detail="Internal server error"`.

**Who is affected**: Any API client that was matching on `status_code == 400` to detect all failures from the transform endpoint will now miss server errors (which will appear as 500). Clients that were parsing `detail` from 400 errors to display messages may no longer receive a message for server-side failures.

**Migration path**: Clients should handle both 4xx (client errors: bad params, validation failures) and 5xx (server errors: infrastructure failures, bugs) as distinct error categories. This is the correct HTTP semantics and the change is a bug fix, but it is a behavioral breaking change for clients that were not distinguishing error classes.

This is a net improvement in API correctness.
  > **Suggestion:** Document in the changelog that transform failures are now correctly categorised as 4xx (client error) vs 5xx (server error). Update any client-side error handling code that relied on catching a flat 400 for all transform failures.


### PR Description Issues

- **[INFO]** PR description does not flag `AddColumn.name` as a breaking API change: The description mentions the schema change as a fix: _"`addCol` could be called without `add_col_params.name`"_, framing it as a validation improvement. However, it does not explicitly flag this as a **breaking change** — existing API consumers that were calling `addCol` without a `name` field (even if it produced downstream errors) will now receive `422 Unprocessable Entity` instead.

For a public API, this omission could surprise downstream teams. The PR also does not include a "Breaking Changes" section or note a semantic version bump.
  > **Suggestion:** Add a **Breaking Changes** section to the PR description:
```
## Breaking Changes
- `addCol` requests without `add_col_params.name` now return `422 Unprocessable Entity`
  immediately instead of succeeding with an undefined column name. Callers must supply
  a non-empty `name` field.
- Unexpected server-side errors in the transform endpoint now return `500` instead of `400`.
  Clients that matched on `status_code == 400` for all failures must be updated to also
  handle `500`.
```
- **[INFO]** PR description is otherwise high quality — good motivation, validation, and impact sections: Positive observations:
- The description clearly references the issue number (`Fixes #191`) and provides a two-point summary of the root causes.
- The "Changes made" section maps closely to the actual diff with no significant gaps.
- The "Validation" section lists concrete commands and pass/fail outcomes — a good practice.
- The "Impact" section correctly identifies the improvement to API semantics.

The only gap is the missing "Breaking Changes" callout noted in the adjacent finding.
  > **Suggestion:** No changes needed beyond adding the breaking changes section. Consider also noting the affected version range if the project maintains a version history.

nodesagar:
@ivantha Addressed all review feedback on this PR and squashed the branch to a single commit: `4bfbfa4`.

### What was fixed
- Restored `AddColumn.name` schema compatibility and moved strict non-blank validation to the transform endpoint path.
- Corrected transform error semantics:
  - preserve explicit `HTTPException` status codes,
  - map `TransformationError` to safe 400 responses,
  - map unexpected exceptions to generic 500 (`"Internal server error"`).
- Fixed exception logging to use a safe local `operation_type` reference.
- Added persistence compensation: if post-save logging fails, file content is reverted.
- Added/updated regression coverage:
  - schema validation tests for addCol name contract,
  - transform HTTP semantics tests,
  - assertion that persistence path (`save_csv_safe`) is actually exercised.

### Verification
- Backend lint/checks passing.
- Targeted backend tests passing (`12 passed`).

ivantha:
This branch has conflicts with `main` that prevent a clean rebase-merge. Please rebase onto the latest `main` (e.g. `git fetch origin && git rebase origin/main`), resolve any conflicts, and force-push.

nodesagar:
@ivantha

 I've enforced the required/non-blank name at the schema level (fixed to return 422) and standardised error redaction for security. Tests are updated, linting is fixed, and I've squashed the PR. Ready to merge!

ivantha:
Hi @nodesagar, thank you for this contribution.

This PR has been inactive for 27 days since the last update, and we haven't been able to converge on getting it merged. I'm closing it to keep the review queue manageable.

This is not a rejection of the work. If you'd like to revive it, please feel free to rebase against `main` and reopen, or open a fresh PR. Thanks again for the effort.

nodesagar:
Closing the loop here: after fetching latest `main`, all of this PR's changes — plus every review fix requested (the `op`→`operation_type` safe reference, persistence compensation/rollback, schema-level `addCol.name` 422 validation, `TransformationError`/`HTTPException` 5xx detail redaction, moving the file read inside the guarded block, the `project_id` fixture, and the `save_table_safe`-was-called assertion) — are already present on `main`.

They landed via the later consolidation/hardening commits (`b617dfa` *refactor: consolidate transformation dispatch into a single registry* and `997ba31` *test+fix: …backend hardening*), so reviving/rebasing this branch would produce no meaningful diff. No further action needed on this PR.

One minor open item: issue #191 is still open even though the fix is fully merged — it can likely be closed.

PR #194, opened 2026-03-13
adding print when DB is not connected properly with the backend

### LINKED ISSUE #193: [Feature] When Database is not connected properly, even after running the app it doesn't show/print meassage
When Database is not connected properly, even after running the app it doesn't show/print meassage

schourasia750:
### Before
image

### After
image

Screenshot 2026-03-13 032118

fixes: #193

### DIFF
--- a/dataloom-backend/app/database.py
+++ b/dataloom-backend/app/database.py
@@ -1,12 +1,27 @@
+import sys
 from collections.abc import Generator
 
-from sqlmodel import Session, create_engine
+from sqlalchemy.exc import OperationalError
+from sqlmodel import Session, create_engine, text
 
 from app.config import get_settings
 
 settings = get_settings()
 
 engine = create_engine(settings.database_url, pool_pre_ping=True)
+
+
+def verify_database_connection():
+    """Attempt to connect to the database. If it fails, print a clean error and exit."""
+    try:
+        with engine.connect() as conn:
+            conn.execute(text("SELECT 1"))
+        print("✅ Successfully connected to the database!")
+    except OperationalError as e:
+        print("❌ FATAL ERROR: Could not connect to the Database!")
+        print("❌ Please check if your database is running and the URL in your .env file is correct.")
+        print(f"❌ Details: {e.orig}")
+        sys.exit(1)
 
 
 def get_db() -> Generator[Session, None, None]:
--- a/dataloom-backend/app/main.py
+++ b/dataloom-backend/app/main.py
@@ -12,6 +12,7 @@ from fastapi.responses import JSONResponse
 
 from app.api.endpoints import projects, transformations, user_logs
 from app.config import get_settings
+from app.database import verify_database_connection
 from app.exceptions import AppException, app_exception_handler
 from app.services.transformation_service import TransformationError
 from app.utils.logging import get_logger, setup_logging
@@ -22,6 +23,7 @@ logger = get_logger(__name__)
 @asynccontextmanager
 async def lifespan(app):
     """Application startup/shutdown lifecycle."""
+    verify_database_connection()
     from alembic.config import Config
 
     from alembic import command


dataloom-backend/app/database.py
```diff
@@ -9,6 +11,19 @@
...
+def verify_database_connection():
+    """Attempt to connect to the database. If it fails, print a clean error and exit."""
+    try:
+        with engine.connect() as conn:
+            conn.execute(text("SELECT 1"))
+        print("✅ Successfully connected to the database!")
+    except OperationalError as e:
+        print("❌ FATAL ERROR: Could not connect to the Database!")
+        print("❌ Please check if your database is running and the URL in your .env file is correct.")
+        print(f"❌ Details: {e.orig}")
```
ivantha, line 23: **[HIGH]** `sys.exit(1)` bypasses async lifespan cleanup

Calling `sys.exit(1)` inside `verify_database_connection()`, which is invoked at the top of the `lifespan` async context manager, raises `SystemExit` before the `yield` is ever reached. This means **all teardown code** placed after the `yield` in `lifespan` (e.g., flushing logs, closing connections, releasing resources) will be silently skipped.

Additionally, `SystemExit` propagates through the ASGI server's startup hook in an uncontrolled manner. Some ASGI servers (e.g., Uvicorn with process managers like Gunicorn) may not surface the exit code cleanly or may restart the process in a loop before giving up, depending on the supervisor configuration.

A better pattern is to raise a descriptive exception, letting the framework/server handle the failure gracefully:
```python
raise RuntimeError("Database connection failed — cannot start application.")
```
This will naturally abort startup, emit a traceback with context, and allow the server to exit with an appropriate code without skipping teardown hooks.

**Suggestion:**
```
Replace `sys.exit(1)` with a `raise RuntimeError("...")` so the ASGI lifecycle is respected:
```python
except OperationalError as e:
    logger.critical("Could not connect to the database!", exc_info=True)
    logger.critical("Check if your database is running and DATABASE_URL in .env is correct.")
    raise RuntimeError(f"Database unavailable at startup: {e.orig}") from e
```
This propagates cleanly through the lifespan context, allowing ASGI servers to exit with a non-zero code while still running shutdown hooks.
```

File: dataloom-backend/app/database.py [hunk 2]
```diff
@@ -9,6 +11,19 @@
...
 
+def verify_database_connection():
+    """Attempt to connect to the database. If it fails, print a clean error and exit."""
+    try:
+        with engine.connect() as conn:
+            conn.execute(text("SELECT 1"))
+        print("✅ Successfully connected to the database!")
+    except OperationalError as e:
+        print("❌ FATAL ERROR: Could not connect to the Database!")
+        print("❌ Please check if your database is running and the URL in your .env file is correct.")
```
ivantha, line 22: **[MEDIUM]** Potential credential exposure via `e.orig` in printed error

`e.orig` is the raw, underlying DBAPI exception (e.g., `psycopg2.OperationalError`). Depending on the driver and the nature of the error, this string can include the full database connection DSN — which may embed the **database password** in plaintext (e.g., `postgresql://user:password@host/db`).

Printing this to stdout means credentials could appear in:
- Container logs shipped to a central log aggregator (Datadog, CloudWatch, Splunk, etc.)
- CI/CD pipeline output
- Developer consoles shared in screenshots (as shown in the PR's own "After" screenshots)

This is especially risky in production environments.

**Suggestion:**
```
Sanitize the error detail before printing, or use the higher-level `str(e)` (which SQLAlchemy redacts) rather than `e.orig`:
```python
# SQLAlchemy's own str(e) is generally safer and still descriptive
print(f"❌ Details: {str(e)}")
```
Or better yet, rely on the logger (which can be configured to redact sensitive patterns) and avoid printing DSN-bearing strings entirely.
```

File: dataloom-backend/app/database.py (hunk 1 above)
Comment (ivantha) [inline dataloom-backend/app/database.py:23]: **[MEDIUM]** `print()` used instead of the established `get_logger` utility

The existing codebase already has a structured logging utility (`from app.utils.logging import get_logger, setup_logging`) imported in `main.py`. Using bare `print()` calls in `database.py` is inconsistent with this pattern and has several drawbacks:

1. **No log levels** — `print` output cannot be filtered by severity (DEBUG/INFO/WARNING/CRITICAL).
2. **No timestamps or structured metadata** — Log aggregators (e.g., Datadog, ELK) cannot parse or enrich plain `print` output.
3. **Not suppressible in tests** — Unit tests that trigger this code path will pollute test output; a logger can be mocked or captured.
4. **Inconsistency** — The rest of the application uses the logger; this module becomes a maintenance outlier.

The success message `"✅ Successfully connected to the database!"` is also noisy on every startup and emits an emoji, which may not render cleanly in all log backends.

**Suggestion:**
```
Use the established logger:
```python
from app.utils.logging import get_logger

logger = get_logger(__name__)

def verify_database_connection():
    try:
        with engine.connect() as conn:
            conn.execute(text("SELECT 1"))
        logger.info("Successfully connected to the database.")
    except OperationalError as e:
        logger.critical("Could not connect to the database. Check DATABASE_URL in .env.")
        raise RuntimeError(f"Database unavailable: {e}") from e
```
```

Comment (ivantha) [inline dataloom-backend/app/database.py:23]: **[MEDIUM]** Synchronous blocking call inside async `lifespan` blocks the event loop

`verify_database_connection()` is a **synchronous** function that performs I/O (a TCP connection + SQL round-trip to the database). It is called directly inside the `async def lifespan(app)` coroutine in `main.py` without being dispatched to a thread pool via `asyncio.to_thread` or `run_in_executor`.

This blocks the asyncio event loop for the duration of the connection attempt. While startup typically happens before request handling begins, this can cause issues with:
- Frameworks or tools that monitor the event loop for stalls during startup
- Connection timeouts that may take several seconds to fire (blocking the loop the entire time)
- Any other async startup tasks that should run concurrently

Note: the engine itself was created with a *synchronous* SQLAlchemy engine (`create_engine`), which is correct for a sync-style session dependency, but the call point is async.

**Suggestion:**
```
Wrap the synchronous call with `asyncio.to_thread` to avoid blocking the event loop:
```python
import asyncio

@asynccontextmanager
async def lifespan(app):
    await asyncio.to_thread(verify_database_connection)
    # ... rest of startup
    yield
    # ... teardown
```
Alternatively, consider an async SQLAlchemy engine for the health-check if the project moves toward fully async I/O.
```

Comment (ivantha) [inline dataloom-backend/app/database.py:23]: **[LOW]** Only `OperationalError` is caught — other startup failures go unhandled

The `except OperationalError` clause covers DBAPI-level connectivity errors (e.g., host unreachable, wrong port, authentication failure). However, several other SQLAlchemy exceptions can be raised at connection time and would **not** be caught, resulting in an unformatted traceback instead of the intended friendly error:

- `sqlalchemy.exc.ArgumentError` — raised when `DATABASE_URL` is malformed or the dialect is unrecognised.
- `sqlalchemy.exc.NoSuchModuleError` — raised when the DB driver (e.g., `psycopg2`) is not installed.
- `sqlalchemy.exc.DatabaseError` — base class for some driver errors not subclassed under `OperationalError`.
- Any `ImportError` from a missing optional driver dependency.

A user who mistyped the scheme in their `.env` (e.g., `postgress://` instead of `postgresql://`) would still see a raw traceback.

**Suggestion:**
```
Broaden the catch to cover the SQLAlchemy base exception, or add multiple clauses:
```python
from sqlalchemy.exc import SQLAlchemyError

except SQLAlchemyError as e:
    logger.critical("Database startup check failed: %s", e)
    raise RuntimeError("Cannot start: database error.") from e
except Exception as e:
    logger.critical("Unexpected error during database check: %s", e)
    raise
```
```

File: dataloom-backend/app/database.py (hunk 2 above)
Comment (ivantha) [inline dataloom-backend/app/database.py:22]: **[LOW]** `e.orig` may be `None` causing unhelpful output

The `OperationalError.orig` attribute is set from the underlying DBAPI exception. However, for certain SQLAlchemy-level `OperationalError` instances (e.g., those constructed directly by SQLAlchemy's connection pool or proxy layers rather than by a DBAPI driver), `orig` can be `None`.

The current code:
```python
print(f"❌ Details: {e.orig}")
```
would print `❌ Details: None`, which is confusing to end users and provides no actionable information.

**Suggestion:**
```
Prefer the top-level `str(e)` which always contains a human-readable message, and use `e.orig` only if it is not None:
```python
detail = str(e.orig) if e.orig is not None else str(e)
print(f"❌ Details: {detail}")
```
```

File: dataloom-backend/app/database.py (hunk 1 above)
Comment (ivantha) [inline dataloom-backend/app/database.py:23]: **[LOW]** Redundant DB connection check — engine already has `pool_pre_ping=True`

The engine is created with `pool_pre_ping=True` on line 11 (unchanged). This setting instructs SQLAlchemy to issue a lightweight `SELECT 1`-style ping before returning *any* connection from the pool, providing the same liveness guarantee on every request.

The new `verify_database_connection()` adds a duplicate ping at startup only. While an *eager* startup check is still valuable UX (fail fast with a clear message rather than on the first request), it is worth documenting that the two mechanisms serve overlapping purposes, so that future maintainers don't accidentally remove `pool_pre_ping` thinking the startup check makes it redundant.

**Suggestion:**
```
Add an inline comment noting the relationship:
```python
# pool_pre_ping ensures liveness on every acquired connection.
# verify_database_connection() provides a fast-fail check at startup for clearer error messages.
engine = create_engine(settings.database_url, pool_pre_ping=True)
```
```

Comment (ivantha) [inline dataloom-backend/app/database.py:23]: **[LOW]** No tests added for `verify_database_connection()`

The new `verify_database_connection()` function has two distinct code paths (success and failure) and a critical side effect (`sys.exit(1)`). Neither path is covered by tests in this PR.

In particular:
- The `sys.exit(1)` path is a common source of test-suite hangs or unexpected process termination if called from test code.
- The success path's `print()` output goes untested.

Without tests, regressions in this startup check (e.g., it suddenly always exits even with a live database) would only be caught in production.

**Suggestion:**
```
Add unit tests using `pytest` with `monkeypatch` and `unittest.mock`:
```python
from unittest.mock import MagicMock, patch
import pytest
from sqlalchemy.exc import OperationalError

def test_verify_db_success(monkeypatch):
    mock_conn = MagicMock()
    mock_engine = MagicMock()
    mock_engine.connect.return_value.__enter__ = lambda s: mock_conn
    mock_engine.connect.return_value.__exit__ = MagicMock(return_value=False)
    with patch("app.database.engine", mock_engine):
        verify_database_connection()  # should not raise

def test_verify_db_failure_exits(monkeypatch):
    with patch("app.database.engine") as mock_engine:
        mock_engine.connect.side_effect = OperationalError("conn", {}, Exception("refused"))
        with pytest.raises(SystemExit) as exc_info:
            verify_database_connection()
        assert exc_info.value.code == 1
```
```

File: dataloom-backend/app/main.py [hunk 3]
```diff
@@ -22,6 +23,7 @@
 @asynccontextmanager
 async def lifespan(app):
     """Application startup/shutdown lifecycle."""
+    verify_database_connection()
```
ivantha, line 26: **[INFO]** `verify_database_connection()` called before Alembic migrations run

In the `lifespan` function, `verify_database_connection()` is called *before* the Alembic migration block. This ordering is intentional and good (confirm connectivity before attempting migrations), but it should be documented with a comment to prevent a future developer from reordering these calls.

If someone moves the Alembic block above `verify_database_connection()`, Alembic would produce a cryptic error instead of the clear connectivity message.

**Suggestion:**
```
Add a brief comment:
```python
async def lifespan(app):
    # 1. Verify DB is reachable before attempting migrations.
    verify_database_connection()
    # 2. Run pending Alembic migrations.
    ...
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR adds a startup database connectivity check that provides a friendlier error message when the backend cannot reach the database. The intent is good and addresses a real developer-experience pain point (issue #193). However, the implementation has a significant correctness issue: calling `sys.exit(1)` inside an async ASGI lifespan hook skips all teardown logic and raises `SystemExit` (a `BaseException`), bypassing `except Exception` handlers. The function also uses bare `print()` calls instead of the project's established `get_logger` utility, risks exposing database credentials via `e.orig`, and only catches `OperationalError` — missing other failure modes like a malformed `DATABASE_URL`. These issues should be addressed before merging.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral change: startup failure now calls `sys.exit(1)` instead of raising an exception: **Previous behavior**: A missing or unreachable database caused a raw `OperationalError` traceback at startup.

**New behavior**: `sys.exit(1)` is called inside the ASGI lifespan hook, raising `SystemExit` (a `BaseException` subclass), skipping teardown and bypassing `except Exception` handlers.

**Who is affected**: Integration tests, deployment wrappers, and any code catching `Exception` around app startup.

**Migration path**: Switch to `raise RuntimeError(...) from e` inside `verify_database_connection()` so the failure propagates as a normal exception through the ASGI lifecycle.


### PR Description Issues

- **[LOW]** PR description lacks motivation and technical context: The PR description consists almost entirely of before/after screenshots with no prose explanation. While the screenshots effectively show the UX improvement, the description is missing:

1. **Motivation / problem statement**: Why was this change needed? What did the old behavior look like from a developer experience perspective (e.g., a silent crash, an unreadable stack trace)?
2. **Technical approach**: A sentence explaining that a startup health-check function was added to `database.py` and wired into the ASGI lifespan hook would help reviewers orient themselves quickly.
3. **Testing**: No mention of how this was tested (manually? unit tests? with a live database intentionally brought down?).
4. **Edge cases considered**: Does it handle a DB that's temporarily unavailable at startup but comes back? (Short answer: no — it exits, which is a reasonable choice, but worth stating.)
  > **Suggestion:** Suggested PR description:
```
## Problem
When the backend could not connect to the database at startup, it would
crash with a raw SQLAlchemy stack trace that was difficult to interpret,
especially for first-time contributors setting up their environment.

## Solution
Added a `verify_database_connection()` startup check in `database.py`
that:
- Performs a lightweight `SELECT 1` ping at application startup (before
  Alembic migrations run).
- Prints a clear, actionable error message if the DB is unreachable and
  exits with code 1.
- Prints a success confirmation on healthy startup.

The function is called from the FastAPI `lifespan` hook in `main.py`.

## Testing
- Tested manually by pointing `DATABASE_URL` at a non-existent host.
- Confirmed clean error message and exit code 1.
- Confirmed normal startup succeeds with a live database.

Fixes #193
```

ivantha:

PR #201, opened 2026-03-13
Fix dropNa replay crash when drop_na_params is None

### LINKED ISSUE #199: dropNa replay crashes during save/revert when `drop_na_params` is `None`
**Describe the bug**
Projects containing a `dropNa` transformation without parameters cannot be saved or reverted.

When `drop_na_params` is serialized as `None` by Pydantic (`transformation_input.dict()`), the replay handler assumes the value is always a dictionary and attempts to call `.get("columns")` on it. If the key exists with value `None`, Python returns `None` instead of the default `{}`, which causes an `AttributeError` during replay.

This causes save and revert operations to fail for projects that include a default `dropNa` transformation.
___
**To Reproduce**

1. Upload a CSV file containing rows with empty values (NaN).
2. Apply the `dropNa` transformation without specifying columns.
3. Confirm the transformation works correctly in the UI/API.
4. Attempt to save the project or revert to a checkpoint.
5. The server throws an error during replay.

Example error:

`AttributeError: 'NoneType' object has no attribute 'get'`
___
**Expected behavior**
The `dropNa` transformation should replay correctly during save and revert operations whether `drop_na_params` is provided or omitted. Projects containing default `dropNa` transformations should be able to create checkpoints and revert without errors.
___
**Screenshots**
N/A
___
**Environment:**

- OS: Ubuntu 22.04
- Browser: Chrome 120
- Node.js version: 18.x
- Python version: 3.11 / 3.12
___
**Additional context**
The issue occurs because the replay handler currently assumes `drop_na_params` is always a dictionary:

`columns = action_details.get("drop_na_params", {}).get("columns")`

However, when serialized with Pydantic, the transformation log may contain:

`{"drop_na_params": None}`

Since the key exists with value `None`, `.get("drop_na_params", {})` returns `None`, and calling `.get("columns")` on `None` raises an exception during replay.

anusha19murthy:
When `drop_na_params` is serialized as `None` by Pydantic, the replay handler assumes it is always a dictionary and calls `.get("columns")` on it. If the key exists with value `None`, Python returns `None` instead of the default `{}`, causing an `AttributeError` during save/revert replay.

## Description

This PR fixes a crash that occurs when replaying a `dropNa` transformation whose parameters were serialized as `None`.
The replay logic previously used:

`columns = action_details.get("drop_na_params", {}).get("columns")`

However, when the transformation log contains:

`{"drop_na_params": None}`

`dict.get()` returns `None` because the key exists, and calling `.get("columns")` on `None` raises:

`AttributeError: 'NoneType' object has no attribute 'get'`

This PR safely handles the optional parameter by normalizing `None` values before accessing `.get("columns")`:

`columns = (action_details.get("drop_na_params") or {}).get("columns")`

This ensures replay works correctly whether `drop_na_params` is missing, `None`, or a dictionary.

Fixes #199
___
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
___
## How Has This Been Tested?

- Reproduced the crash by applying a `dropNa` transformation without parameters and attempting to save/revert the project.
- Verified that replay succeeds after the fix.
- Confirmed that other transformation handlers are unaffected.
- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing
___
N/A
___
## Checklist

- [x] Tests pass locally

### Updates after review

- Added regression tests for dropNa replay:
- drop_na_params = None (crash scenario)
- missing drop_na_params key
- valid columns case
- Confirmed all transformation handlers use `_get_params`, so no similar issues exist
- Improved readability by extracting params and adding inline comment


anusha19murthy:
Hi @ivantha 
Please review this pr and let me know if any changes are needed.
Thanks!

dataloom-backend/app/services/transformation_service.py
```diff
@@ -498,7 +498,7 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
         return trim_whitespace(df, column)
 
     elif action_type == "dropNa":
-        columns = action_details.get("drop_na_params", {}).get("columns")
+        columns = (action_details.get("drop_na_params") or {}).get("columns")
```
ivantha: **[HIGH]** Test changes absent from diff despite PR claiming new tests were added

The PR description's checklist marks **"New tests added"** as completed, yet the diff contains only a single changed line in `transformation_service.py`. No test file is included in the diff.

This is a significant gap: without seeing the test code, it is impossible to verify:
- That a regression test reproduces the exact crash scenario (`drop_na_params` key present with value `None`)
- That edge cases are covered (key absent entirely, key present with empty dict `{}`, key present with valid columns list)
- That the fix is guarded against future regressions

If tests were added in a separate commit not included in the diff, the reviewer should explicitly request them. If they were not added, the checklist is misleading.

**Suggestion:**
```
Add a dedicated test module (or extend an existing one) covering at minimum:
```python
def test_apply_logged_transformation_dropna_params_none():
    """Regression test for #199: drop_na_params serialized as None must not crash."""
    df = pd.DataFrame({"a": [1, None, 3], "b": [None, 2, None]})
    action_details = {"drop_na_params": None}   # The crash-triggering case
    result = apply_logged_transformation(df, "dropNa", action_details)
    assert result is not None

def test_apply_logged_transformation_dropna_params_missing():
    df = pd.DataFrame({"a": [1, None]})
    result = apply_logged_transformation(df, "dropNa", {})
    assert result is not None

def test_apply_logged_transformation_dropna_params_with_columns():
    df = pd.DataFrame({"a": [1, None], "b": [None, 2]})
    action_details = {"drop_na_params": {"columns": ["a"]}}
    result = apply_logged_transformation(df, "dropNa", action_details)
    assert len(result) == 1
```
```

File: dataloom-backend/app/services/transformation_service.py [hunk 2]
```diff
@@ -498,7 +498,7 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
...
         return drop_na(df, columns)
```
ivantha, line 528: **[MEDIUM]** Consistency: other transformation handlers may share the same None-key bug pattern

The PR description states *"Confirmed that other transformation handlers are unaffected"*, but only one line is changed and no wider audit evidence is provided in the diff.

The pattern `action_details.get("<param_key>", {}).get("<sub_key>")` is almost certainly repeated for other `action_type` branches in `apply_logged_transformation`. Any of those branches is vulnerable to the identical crash if Pydantic serializes their respective parameter objects as `None`.

For example, if there are handlers for `renameColumn`, `fillNa`, `castType`, etc. that follow the same pattern, they will crash under the same conditions. A targeted search/audit of the full function body is warranted.

**Suggestion:**
```
Audit the entire `apply_logged_transformation` function (and any similar replay helpers) for occurrences of the pattern `.get("...", {}).get(...)` and replace them uniformly:
```python
# Vulnerable pattern (do not use):
params = action_details.get("some_params", {}).get("field")

# Safe pattern (use instead):
params = (action_details.get("some_params") or {}).get("field")
```
Alternatively, introduce a small helper to normalise params extraction:
```python
def _get_params(action_details: dict, key: str) -> dict:
    """Return params dict, safely handling None and missing keys."""
    return action_details.get(key) or {}
```
Then use `_get_params(action_details, "drop_na_params").get("columns")` everywhere.
```

File: dataloom-backend/app/services/transformation_service.py (hunk 1 above)
Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py]: **[LOW]** Implicit falsy-coercion via `or {}` swallows unexpected types silently

The expression `(action_details.get("drop_na_params") or {})` is idiomatic Python and correct for the stated `None`-vs-dict case, but it silently coerces **any** falsy value — including `0`, `False`, `[]`, and `""` — to `{}`. While these types are extremely unlikely in this context, the pattern provides no early warning if the serialisation logic produces something unexpected.

A more explicit check communicates intent more precisely and would surface type surprises as early errors rather than silent no-ops:
```python
drop_na_params = action_details.get("drop_na_params")
if drop_na_params is not None and not isinstance(drop_na_params, dict):
    raise ValueError(f"Expected dict or None for drop_na_params, got {type(drop_na_params)}")
columns = (drop_na_params or {}).get("columns")
```
For a production replay handler, this distinction can matter when debugging data corruption issues.

**Suggestion:**
```
If strict type safety is desired:
```python
drop_na_params = action_details.get("drop_na_params")  # may be None or dict
columns = drop_na_params.get("columns") if isinstance(drop_na_params, dict) else None
return drop_na(df, columns)
```
This is both explicit and safe, making the intent clear to future readers without relying on truthiness coercion.
```

Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py]: **[LOW]** Readability: inline one-liner obscures two-step logic; consider extracting to a variable

The fix is functionally correct, but the nested `(... or {}).get(...)` idiom, while concise, may be confusing to readers unfamiliar with the None-coercion pattern. The original bug and its fix are subtle enough that future maintainers may not immediately understand why `or {}` was preferred over a default argument.

Extracting into an intermediate variable with a brief comment would improve readability and make code-review of future changes to this block easier.

**Suggestion:**
```
```python
elif action_type == "dropNa":
    # drop_na_params may be None when serialised by Pydantic (see #199)
    drop_na_params = action_details.get("drop_na_params") or {}
    columns = drop_na_params.get("columns")
    return drop_na(df, columns)
```
```

Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py]: **[INFO]** Root cause (Pydantic serializing None) not addressed at the source

The fix is a defensive workaround at the **consumer** level. The underlying issue — that Pydantic serializes `drop_na_params` as `None` into the transformation log — is not addressed.

Depending on the Pydantic model, the field may be declared as `Optional[dict]` without a default of `{}`, causing it to serialize `null` into the log. If the serialisation were corrected upstream (e.g., by using `drop_na_params: dict = Field(default_factory=dict)`), the consumer would never receive `None` and the defensive `or {}` would be unnecessary.

This is not a blocker for merging the fix, but it is worth investigating to prevent analogous issues in other handlers.

**Suggestion:**
```
Locate the Pydantic model that owns `drop_na_params`. If it is declared as:
```python
drop_na_params: Optional[dict] = None
```
consider changing it to:
```python
drop_na_params: dict = Field(default_factory=dict)
```
so the field always serialises as `{}` rather than `null`. Keep the `or {}` guard in the handler as a belt-and-suspenders measure for backward compatibility with existing logs.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This is a minimal, well-targeted one-line bug fix that correctly resolves an `AttributeError` crash in the `dropNa` replay handler. The fix — replacing `.get("drop_na_params", {})` with `(... or {})` — properly handles the case where the key is present but holds `None`, which the original default-argument approach could not catch. The PR description is exemplary in its clarity and accurately describes the change. The main concerns are: (1) no test code is visible in the diff despite the checklist claiming tests were added, which should be verified; (2) the same vulnerable `.get(..., {}).get(...)` pattern likely exists in other transformation handlers and deserves a consistency audit; and (3) the root cause in the Pydantic model serialization is not addressed upstream. Overall the fix is correct and safe to merge after the test gap is resolved.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes detected — patch-level fix only: **Previous behavior**: `apply_logged_transformation` with `action_type == "dropNa"` and `action_details = {"drop_na_params": None}` raised `AttributeError: 'NoneType' object has no attribute 'get'`.

**New behavior**: The function safely treats `None` as an empty dict and proceeds to call `drop_na(df, None)` (i.e., drops NA across all columns), which is the correct fallback behavior.

**Who is affected**: Only consumers replaying a `dropNa` transformation where `drop_na_params` was serialized as `None`. All other callers are unaffected.

**Migration path**: None required. The change is fully backward-compatible and fixes a crash that made the feature unusable in the described scenario.
  > **Suggestion:** No action required. Confirm the `drop_na(df, None)` call (with `columns=None`) in the underlying function correctly defaults to dropping NA across all columns, to ensure the semantics match user expectations.


### PR Description Issues

- **[LOW]** Documentation checkbox left unchecked without justification: The PR checklist item *"I have added/updated documentation as needed"* is left unchecked (`- [ ]`). While a one-line bug fix rarely requires documentation changes, the PR description offers no explanation for why this item was skipped.

For a replay handler fix, it may be worth updating inline comments or docstrings in `apply_logged_transformation` to mention the `None`-safety requirement for parameter dictionaries.
  > **Suggestion:** Either:
1. Check the box with a note: `- [x] I have added/updated documentation as needed (added inline comment in handler)`
2. Or explicitly note: `- [ ] I have added/updated documentation as needed — no user-facing docs impacted by this internal fix`

Additionally, add an inline comment above the fixed line:
```python
# drop_na_params may be None when Pydantic serialises an absent parameter (see #199)
columns = (action_details.get("drop_na_params") or {}).get("columns")
```
- **[INFO]** PR description is high quality and accurately matches the diff: The PR description is well-written and thorough. It:
- Clearly explains the **root cause** with the exact key-present-with-None scenario
- Provides the **before/after code snippets** inline
- Explains the precise Python semantics (`dict.get()` returning `None` vs the default)
- Correctly identifies the fix and its rationale
- Links to issue #199
- Lists manual testing steps

This level of detail is commendable and makes the review significantly easier.
  > **Suggestion:** No changes needed. This is a positive observation.

anusha19murthy:
Hi @ivantha ! The pr is ready please review it once. 
Thanks!

ivantha:
The fix and tests look solid — the `_get_params` helper is a good defensive pattern applied consistently across all replay handlers, and the regression tests are comprehensive.

However, this PR includes a change to `dataloom-backend/test.db` (a binary SQLite test database file). This file should not be tracked in git — see the recently merged fix in #189 which added `*.db` to `.gitignore`. Please:
1. Remove `test.db` from version control: `git rm --cached dataloom-backend/test.db`
2. Rebase onto the latest `main` (which includes the .gitignore fix) and force-push

Once that's done the PR can be merged.

ivantha:
Thank you for fixing the dropNa replay crash! This work has been superseded by #227 which provides the same fix. Feel free to follow up on that PR.

PR #204, opened 2026-03-14
added multi sorting for pre-gsoc task #81

### LINKED ISSUE #81: Implement Multi-Column Sort
**Difficulty:** Hard
**Area:** full stack

## Description

DataLoom currently supports sorting data by a single column only, via the `apply_sort(df, column, ascending)` function. In real-world data wrangling, users frequently need to sort by multiple columns with independent sort directions. For example, sorting a sales dataset first by region (ascending) and then by revenue (descending) within each region.

This issue requires extending the sort functionality across the entire stack: the backend must accept and process a list of sort criteria, the schema must be updated to define the new input structure, and the frontend's SortForm must be redesigned to allow users to add, remove, and reorder multiple sort criteria before submitting.

The pandas `DataFrame.sort_values()` method natively supports multi-column sorting via the `by` and `ascending` parameters accepting lists, so the backend implementation is straightforward. The main complexity lies in the frontend UX for managing multiple sort criteria.

## Acceptance Criteria

- [ ] Sorting accepts a list of column + direction pairs and produces correct results (primary, secondary, etc.)
- [ ] At least one sort criterion is required; submitting with no criteria is rejected
- [ ] The sort form allows adding multiple sort criteria rows, each with a column dropdown and ascending/descending toggle
- [ ] Users can remove individual sort criteria rows
- [ ] Sort criteria order in the UI determines sort priority
- [ ] Single-column sort still works (backward compatibility)
- [ ] Sort operation is logged correctly in the project change log
- [ ] Form validation prevents submission with empty column selections

## Resources

- `dataloom-backend/app/services/transformation_service.py` — `apply_sort()` function
- `dataloom-backend/app/schemas.py` — schema definitions
- `dataloom-backend/app/api/endpoints/transformations.py` — endpoint handlers
- `dataloom-frontend/src/Components/forms/SortForm.jsx` — sort form UI
- [pandas `sort_values` documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sort_values.html)

schourasia750:
fixes: #81
image


dataloom-backend/app/services/transformation_service.py
```diff
@@ -91,20 +91,30 @@ def apply_filter(df: pd.DataFrame, column: str, condition: str, value: str) -> p
     return ops[condition_str]()
 
 
-def apply_sort(df: pd.DataFrame, column: str, ascending: bool) -> pd.DataFrame:
-    """Sort DataFrame by a column.
+def apply_sort(df: pd.DataFrame, criteria: list[dict]) -> pd.DataFrame:
```
ivantha, line 95: **[HIGH]** apply_sort function signature changed incompatibly

**Previous behavior**: `apply_sort(df, column: str, ascending: bool)` — simple positional parameters.

**New behavior**: `apply_sort(df, criteria: list[dict])` — completely different signature requiring a list of dicts.

**Who is affected**: Any internal or external code that calls `apply_sort` directly (e.g., other service modules, notebooks, custom transformation pipelines, or tests not included in this diff).

**Migration path**: All call sites must be updated. The old call `apply_sort(df, 'age', True)` must become `apply_sort(df, [{'column': 'age', 'ascending': True}])`.

dataloom-frontend/src/Components/forms/SortForm.jsx
```diff
@@ -39,47 +69,105 @@ const SortForm = ({ projectId, onClose }) => {
...
-            </select>
-          </div>
+        <h3 className="font-semibold text-gray-900 mb-1">Sort Dataset</h3>
+        <p className="text-xs text-gray-500 mb-3">
+          Add multiple sort criteria. Priority is determined by order (top = primary sort).
+        </p>
+
+        <div className="space-y-2 mb-3">
+          {criteria.map((criterion, index) => (
+            <div key={index} className="flex items-center gap-2">
```
ivantha, line 76: **[HIGH]** React list key uses array index on a reorderable list

The `criteria.map` renders each row using `key={index}`. React uses keys to track identity between renders. When `moveCriterion` swaps two elements, the array indices stay 0, 1, 2, … but the *content* shifts. React will therefore **not** unmount/remount the rows — it will re-use the same DOM nodes in place and only update their props. This means any uncontrolled state (e.g., browser-native select focus, internal component state in child components) will **not follow** the moved criterion and will appear to 'stick' to its original visual position.

This is a well-known React pitfall with reorderable lists.

**Suggestion:**
```
Attach a stable, unique ID to each criterion on creation:
```js
const emptyRow = () => ({ id: crypto.randomUUID(), column: "", ascending: true });

// In the map:
criteria.map((criterion) => (
  <div key={criterion.id} className="flex items-center gap-2">
    ...
  </div>
))
```
This ensures React correctly tracks each row's identity when the array is reordered.
```

File: dataloom-backend/app/api/endpoints/transformations.py [hunk 3]
```diff
@@ -43,8 +43,8 @@ def _handle_basic_transform(df, transformation_input, project, db, project_id):
     elif op == "sort":
         if not transformation_input.sort_params:
             raise HTTPException(status_code=400, detail="Sort parameters required")
-        p = transformation_input.sort_params
-        return ts.apply_sort(df, p.column, p.ascending), False
+        criteria = [c.dict() for c in transformation_input.sort_params.sort_criteria]
```
ivantha, line 46: **[MEDIUM]** Pydantic .dict() is deprecated in Pydantic v2

In `transformations.py` line 46, `c.dict()` is called to serialize each `SortCriterion`. In **Pydantic v2**, `.dict()` is deprecated in favor of `.model_dump()` and emits a `PydanticDeprecatedSince20` warning. Depending on the project's Pydantic version and warning filters, this may surface as runtime warnings in logs or, in future Pydantic versions, raise an error entirely.

**Suggestion:**
```
```python
# Replace:
criteria = [c.dict() for c in transformation_input.sort_params.sort_criteria]
# With:
criteria = [c.model_dump() for c in transformation_input.sort_params.sort_criteria]
```
Alternatively, pass the `SortCriterion` objects directly to `apply_sort` and type the parameter as `list[SortCriterion]` to avoid the dict serialization entirely.
```

File: dataloom-backend/app/services/transformation_service.py [hunk 4]
```diff
@@ -497,6 +507,14 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
...
+        sort_params = action_details.get("sort_params", {})
+        criteria = [
+            {"column": c["column"], "ascending": c["ascending"]}
+            for c in sort_params.get("sort_criteria", [])
+        ]
+        return apply_sort(df, criteria)
+
     elif action_type == "dropNa":
         columns = action_details.get("drop_na_params", {}).get("columns")
         return drop_na(df, columns)
```
ivantha, line 576: **[MEDIUM]** apply_logged_transformation sort path is unguarded against malformed stored data

In `apply_logged_transformation`, the sort action handler iterates over `sort_params.get('sort_criteria', [])` and accesses `c['column']` and `c['ascending']` via dict key access (not `.get()`). If `action_details` was persisted with a different schema (e.g., the old `{ 'column': ..., 'ascending': ... }` flat structure from before this PR), this will raise an unhandled `KeyError` at runtime when replaying historical logs.

This is a data migration concern: any sort actions recorded before this PR's deployment will fail to replay.

**Suggestion:**
```
Add a backward-compatibility shim for historical log entries:
```python
elif action_type == "sort":
    sort_params = action_details.get("sort_params", {})
    # Handle legacy flat format recorded before multi-sort migration
    if "column" in sort_params:
        criteria = [{"column": sort_params["column"], "ascending": sort_params.get("ascending", True)}]
    else:
        criteria = [
            {"column": c["column"], "ascending": c["ascending"]}
            for c in sort_params.get("sort_criteria", [])
        ]
    return apply_sort(df, criteria)
```
```

File: dataloom-frontend/src/Components/forms/SortForm.jsx [hunk 5]
```diff
@@ -5,31 +5,61 @@ import { SORT } from "../../constants/operationTypes";
...
+    if (emptyIndex !== -1) {
+      handleError({
+        message: `Row ${emptyIndex + 1}: Please select a column before submitting.`,
+      });
+      return;
+    }
+
     setLoading(true);
     clearError();
     try {
```
ivantha, line 51: **[MEDIUM]** No duplicate column validation in SortForm

The UI allows a user to add the same column twice (e.g., `age` ascending + `age` descending). `pandas.DataFrame.sort_values` silently accepts duplicate column names in the `by` list — the behavior is that the second occurrence is effectively redundant, not an error — but this is almost certainly a user mistake and yields confusing results. Neither the frontend nor the backend schema validates for uniqueness of column names across criteria.

**Suggestion:**
```
Add a uniqueness check in `handleSubmit`:
```js
const seenColumns = new Set();
for (const [i, c] of criteria.entries()) {
  if (seenColumns.has(c.column)) {
    handleError({ message: `Row ${i + 1}: Column "${c.column}" is already used in a previous criterion.` });
    return;
  }
  seenColumns.add(c.column);
}
```
Alternatively, filter already-selected columns from each row's dropdown options.
```

File: dataloom-backend/tests/test_transformations.py [hunk 6]
```diff
@@ -67,16 +67,35 @@ def test_filter_invalid_condition(self, sample_df):
...
+            {"column": "city", "ascending": True},
+            {"column": "age", "ascending": False},
+        ]
+        result = apply_sort(sample_df, criteria)
+        assert result.iloc[0]["city"] == "Chicago"
+
+    def test_multi_sort_single_criterion_compat(self, sample_df):
+        """Single-criterion list behaves like the old single-column sort."""
+        result = apply_sort(sample_df, [{"column": "age", "ascending": True}])
+        assert list(result["age"]) == [25, 30, 35]
```
ivantha, line 100: **[MEDIUM]** No test for apply_logged_transformation sort path

The `apply_logged_transformation` function was extended with a new `sort` branch (lines 507–519), but there are no tests covering this path. This is the function responsible for replaying transformation history, so correctness here is critical. The new branch involves non-trivial dict access and a call to `apply_sort` that could fail silently with empty criteria.

**Suggestion:**
```
Add tests for the logged transformation replay path:
```python
def test_logged_sort_multi_column(self, sample_df):
    action_details = {
        "sort_params": {
            "sort_criteria": [
                {"column": "age", "ascending": True},
                {"column": "city", "ascending": False},
            ]
        }
    }
    result = apply_logged_transformation(sample_df, "sort", action_details)
    assert result.iloc[0]["age"] == 25

def test_logged_sort_empty_criteria_raises(self, sample_df):
    action_details = {"sort_params": {"sort_criteria": []}}
        apply_logged_transformation(sample_df, "sort", action_details)
```
```

File: dataloom-backend/tests/test_transformations.py [hunk 7]
```diff
@@ -67,16 +67,35 @@ def test_filter_invalid_condition(self, sample_df):
...
 
 
 class TestAddRow:
```
ivantha, line 101: **[MEDIUM]** test_multi_sort_primary_secondary is too weak

The multi-sort test only asserts `result.iloc[0]['city'] == 'Chicago'`, which would pass even with a single-column sort on `city`. The entire point of the secondary sort (age descending within the same city) is never validated. If there are multiple rows with `city == 'Chicago'`, the test does not verify that those rows are ordered by descending age.

**Suggestion:**
```
Strengthen the test to verify the secondary sort key:
```python
def test_multi_sort_primary_secondary(self, sample_df):
    # Assumes sample_df has at least 2 rows with city == "Chicago"
    criteria = [
        {"column": "city", "ascending": True},
        {"column": "age", "ascending": False},
    ]
    result = apply_sort(sample_df, criteria)
    chicago_rows = result[result["city"] == "Chicago"].reset_index(drop=True)
    assert chicago_rows.iloc[0]["age"] >= chicago_rows.iloc[1]["age"], \
        "Rows with same city should be sorted by descending age"
```
Also consider adding a fixture with controlled data (two rows sharing a city) to make this test deterministic.
```

File: dataloom-backend/app/services/transformation_service.py [hunk 8]
```diff
@@ -91,20 +91,30 @@ def apply_filter(df: pd.DataFrame, column: str, condition: str, value: str) -> p
...
-        ascending: Sort direction.
+        criteria: List of dicts, each with 'column' (str) and 'ascending' (bool).
+                  Order determines sort priority (first = primary key).
 
     Returns:
         Sorted DataFrame.
+
+    Raises:
+        TransformationError: If criteria list is empty or any column is not found.
     """
```
ivantha, line 108: **[LOW]** Error message exposes Python list repr to end users

When one or more columns are not found, the error is formatted as:
```python
f"Column(s) not found: {missing}"
```
This produces messages like `Column(s) not found: ['age', 'city']` (Python list syntax with square brackets and quotes), which is not user-friendly and leaks implementation details.

**Suggestion:**
```
```python
missing_str = ", ".join(f"'{c}'" for c in missing)
raise TransformationError(f"Column(s) not found: {missing_str}")
# Produces: Column(s) not found: 'age', 'city'
```
```

File: dataloom-frontend/src/Components/forms/SortForm.jsx [hunk 9]
```diff
@@ -39,47 +69,105 @@ const SortForm = ({ projectId, onClose }) => {
...
+              >
+                <option value="">Column name</option>
+                {columns.map((col) => (
+                  <option key={col} value={col}>
+                    {col}
+                  </option>
+                ))}
+              </select>
+
+              {/* Direction select */}
```
ivantha, line 94: **[LOW]** Missing null-guard for columns from ProjectContext

The component now consumes `columns` from `useProjectContext()` and maps over it to render `<option>` elements. If `columns` is `undefined` or `null` (e.g., context not yet loaded, or project has no columns), calling `.map()` on it will throw a runtime error:
```
TypeError: Cannot read properties of undefined (reading 'map')
```

**Suggestion:**
```
Use a safe default:
```jsx
const { columns = [] } = useProjectContext();
// or inline:
{(columns ?? []).map((col) => (
  <option key={col} value={col}>{col}</option>
))}
```
```

File: dataloom-backend/app/services/transformation_service.py (hunk 1 above)
Comment (ivantha) [inline dataloom-backend/app/services/transformation_service.py:95]: **[LOW]** apply_sort accepts untyped list[dict] instead of typed models

The `apply_sort` service function now takes `criteria: list[dict]` — a plain list of untyped dicts. This bypasses Python's type system: if a caller passes `{'col': 'age', 'asc': True}` (wrong keys), there is no type error at the call site; the failure would only surface at runtime as a `KeyError`. Keeping Pydantic models in the service layer would provide validation and IDE completion.

**Suggestion:**
```
Import and use the `SortCriterion` model from schemas:
```python
from app.schemas import SortCriterion

def apply_sort(df: pd.DataFrame, criteria: list[SortCriterion]) -> pd.DataFrame:
    ...
    missing = [c.column for c in criteria if c.column not in df.columns]
    columns = [c.column for c in criteria]
    ascending = [c.ascending for c in criteria]
    return df.sort_values(by=columns, ascending=ascending)
```
Then in the endpoint, pass `transformation_input.sort_params.sort_criteria` directly instead of converting to dicts.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR successfully implements multi-column sorting end-to-end (schema, service, endpoint, UI), with clean code structure, good validator usage, and a meaningful UI upgrade. However, it contains a **breaking API change** (both the REST payload shape and the internal `apply_sort` signature) that is entirely undocumented in the PR description. The frontend has a classic React key-with-reorderable-list bug that will cause visual glitches when criteria are reordered, and the `.dict()` Pydantic v2 deprecation should be addressed. Test coverage is mostly good but the `apply_logged_transformation` sort path is untested and the multi-sort assertion is too weak to actually verify secondary sort behavior.

### ⚠️ Breaking Changes

- **[HIGH]** SortParameters schema is a breaking API change: **Previous behavior**: The `SortParameters` Pydantic model accepted `{ column: str, ascending: bool }` directly.

**New behavior**: `SortParameters` now wraps a `sort_criteria: list[SortCriterion]` field, entirely replacing the old flat structure.

**Who is affected**: Any API client (scripts, integrations, saved requests) that was posting `sort_params: { "column": "age", "ascending": true }` will now receive a `422 Unprocessable Entity` because the schema no longer accepts those fields at the top level.

**Migration path**: Clients must update their payloads from:
```json
{ "sort_params": { "column": "age", "ascending": true } }
```
to:
```json
{ "sort_params": { "sort_criteria": [{ "column": "age", "ascending": true }] } }
```
Consider adding a deprecation notice or a backward-compatible shim that accepts both shapes during a transition period.
- **[HIGH]** apply_sort function signature changed incompatibly: **Previous behavior**: `apply_sort(df, column: str, ascending: bool)` — simple positional parameters.

**New behavior**: `apply_sort(df, criteria: list[dict])` — completely different signature requiring a list of dicts.

**Who is affected**: Any internal or external code that calls `apply_sort` directly (e.g., other service modules, notebooks, custom transformation pipelines, or tests not included in this diff).

**Migration path**: All call sites must be updated. The old call `apply_sort(df, 'age', True)` must become `apply_sort(df, [{'column': 'age', 'ascending': True}])`.


### PR Description Issues

- **[HIGH]** PR description is insufficient — missing motivation, implementation details, and testing notes: The PR description contains only an issue reference (`fixes: #81`) and a screenshot. There is no explanation of:
- **What was changed**: multi-column sort support was added
- **Why**: what user problem this solves beyond the linked issue
- **How it works**: the new `sort_criteria` list schema, the updated `apply_sort` signature
- **Migration impact**: existing API clients using `{ column, ascending }` flat format will break
- **How to test**: manual testing steps or how to exercise the new UI
- **Breaking change notice**: the API contract changed and this is not called out

A screenshot alone does not communicate the scope or risk of the change.
  > **Suggestion:** Expand the description with at minimum:
```markdown
## What
Adds multi-column sort support (fixes #81). Users can now define an ordered list of sort criteria — each with a column and direction — enabling secondary/tertiary sorts when rows share the same primary sort key value.

## Changes
- **Backend**: Replaced flat `SortParameters(column, ascending)` with `SortParameters(sort_criteria: list[SortCriterion])`. Updated `apply_sort` service function and `apply_logged_transformation` replay path.
- **Frontend**: Rewrote `SortForm` to support a dynamic list of criteria with up/down reordering.

## ⚠️ Breaking Change
The `sort_params` API payload shape has changed. Existing clients sending `{ "column": "...", "ascending": true }` must migrate to `{ "sort_criteria": [{ "column": "...", "ascending": true }] }`.

## Testing
- [ ] Apply single-column sort — verify backward-compatible behavior
- [ ] Apply multi-column sort — verify secondary sort key is respected
- [ ] Submit with an empty column row — verify validation error shown
- [ ] Reorder criteria with up/down arrows — verify sort priority changes
```

dataloom-frontend/src/Components/forms/SortForm.jsx
```diff
@@ -5,31 +5,61 @@ import { SORT } from "../../constants/operationTypes";
...
         operation_type: SORT,
         sort_params: {
-          column,
-          ascending,
+          sort_criteria: criteria,
         },
       });
       setResult(response);
-      console.log("Sort API response:", response);
     } catch (err) {
```
ivantha, line 59: **[LOW]** criteria state not reset after successful sort application

After a successful sort, `result` is set but `criteria` remains in its current state. Depending on UX requirements, users might expect the form to reset to a single empty row after applying, or to preserve their criteria for iterative re-use. As written, criteria persist — which is fine, but it's an implicit UX choice that should be deliberate. More importantly, if `onClose` is called after viewing results, the stale criteria will persist if the form is ever reopened (depending on how the component is mounted/unmounted).

**Suggestion:**
```
Make it intentional by adding a reset after success if that is the desired UX, or add a comment explaining why criteria are preserved:
```js
setResult(response);
// Intentionally preserve criteria so user can refine and re-apply
```
```

File: dataloom-frontend/src/Components/forms/SortForm.jsx [hunk 11]
```diff
@@ -39,47 +69,105 @@ const SortForm = ({ projectId, onClose }) => {
...
+                onClick={() => moveCriterion(index, 1)}
+                disabled={index === criteria.length - 1}
+                className="p-1 text-gray-400 hover:text-gray-600 disabled:opacity-30 disabled:cursor-not-allowed"
+                title="Move down"
+              >
+                <LuChevronDown className="w-4 h-4" />
+              </button>
+
+              {/* Remove button */}
+              <button
```
ivantha, line 125: **[LOW]** Reorder buttons lack accessible aria-labels

The up/down chevron buttons use only the `title` attribute for their labels. The `title` attribute is not announced reliably by all screen readers and is invisible to keyboard users who don't hover. Buttons with only icon content must have an accessible name via `aria-label` or `aria-labelledby`.

**Suggestion:**
```
```jsx
<button
  type="button"
  onClick={() => moveCriterion(index, -1)}
  disabled={index === 0}
  aria-label={`Move criterion ${index + 1} up`}
  title="Move up"
  className="..."
>
  <LuChevronUp className="w-4 h-4" aria-hidden="true" />
</button>
```
Apply the same pattern to the down and remove buttons.
```

File: dataloom-backend/app/schemas.py [hunk 12]
```diff
@@ -91,13 +91,26 @@ class FilterParameters(BaseModel):
     value: str
 
 
-class SortParameters(BaseModel):
-    """Parameters for a column sort operation."""
+class SortCriterion(BaseModel):
+    """A single sort criterion: a column name and sort direction."""
 
     column: str
     ascending: bool
```
ivantha, line 100: **[INFO]** SortCriterion model could benefit from field-level validation for column

The `SortCriterion.column` field is typed as `str` with no constraints. An empty string `""` would pass schema validation but fail at the service layer with `TransformationError`. The frontend guards against this, but the backend API would accept `{ "column": "", "ascending": true }` and return a 500-level `TransformationError` rather than a clean 400 validation error.

**Suggestion:**
```
Use Pydantic's `field_validator` or `Field(min_length=1)` to reject empty column names early:
```python
from pydantic import Field

class SortCriterion(BaseModel):
    column: str = Field(min_length=1, description="Column name to sort by")
    ascending: bool
```
```

ivantha:
Holding this PR — it is linked to a GSoC-labeled issue (#81) and should go through GSoC evaluation rather than maintainer merge.

ivantha:
**Correction:** The previous comment incorrectly flagged this as a GSoC hold — issue #81 has a `pre-gsoc` label which does **not** block merges. The actual blockers are: (1) branch is behind main — please rebase, and (2) both Frontend and Backend Lint checks are failing. Please fix the lint errors and rebase before we can merge.

ivantha:

PR #183, opened 2026-03-12
feat: add multi-format file upload support (xlsx, json, parquet, tsv)

Pranjal0410:
Implements the first item in the 2026 scope: "Users can upload xlsx, json, parquet, and tsv files in addition to CSV."

## What changed and why

The upload pipeline was hardcoded to CSV — `validate_upload_file()` only accepted `.csv`, `store_upload()` copied the file as-is assuming CSV, and the upload endpoint read it with `read_csv_safe()`. This meant anyone with Excel, JSON, Parquet, or TSV data had to manually convert to CSV before using DataLoom.

**Dashboard showing successful uploads across multiple formats:**

Screenshot 2026-03-12 at 8 22 51 PM

## How it works now

**Format-aware reading:** Added a `read_datafile()` function in `pandas_helpers.py` that detects the file extension and routes to the correct pandas reader:
- `.csv` → `pd.read_csv()`
- `.xlsx` → `pd.read_excel(engine='openpyxl')`
- `.json` → `pd.read_json()`
- `.parquet` → `pd.read_parquet(engine='pyarrow')`
- `.tsv` → `pd.read_csv(sep='\t')`

**CSV normalization:** The key design decision — all formats get normalized to a CSV working copy after upload. This means the entire existing transformation pipeline (filter, sort, pivot, checkpoint/revert, etc.) works without any changes. The original file stays in its native format.

**Original path resolution:** Since the original can now be `.xlsx` but the working copy is always `_copy.csv`, updated `get_original_path()` to search across all supported extensions instead of blindly replacing `_copy.csv` with `.csv`.

## Files changed

- `app/config.py` — extended `allowed_extensions` to include all 5 formats
- `app/utils/pandas_helpers.py` — added `READER_MAP` and `read_datafile()`
- `app/services/file_service.py` — updated `store_upload()` to read any format and save CSV copy, updated `get_original_path()` to search across extensions
- `app/api/endpoints/projects.py` — upload endpoint now uses `read_datafile()` instead of `read_csv_safe()` for the initial read
- `pyproject.toml` / `uv.lock` — added `openpyxl` and `pyarrow` dependencies
- `tests/test_upload.py` — added acceptance tests for xlsx, json, parquet, tsv; kept rejection tests for exe, pdf, no-extension

## Also fixed

[...]

PR #215, opened 2026-03-22
Add multi-format ingestion and data quality engine PoC

anusha19murthy:
### **Description**
Traditional data wrangling tools like DataLoom only process data at the end of a pipeline — when a transform chain breaks or the final output looks wrong, users have no way to identify which step caused the problem. Additionally, there is no data quality assessment, no quality scoring, no channel distribution analysis, and no fix suggestions, making it impossible to understand the health of data at any operation. The existing pipeline executor is a black box: input goes in, final output comes out, everything in between is invisible.
___
### **Solution**
I introduced a run_quality_assessment() utility and extended the project pipeline with an opt-in intermediate capture loop. When include_intermediates: true is passed in the request, the executor captures a per-column statistical profile, duplicate report, outlier report (IQR and Z-score), composite quality score (0–100), and ranked fix suggestions at each transformation step. I return the intermediates as a structured list alongside the existing final output, completely bypassing any changes to the operator implementations themselves. I preserved the existing API contract fully for all callers that do not set the flag.
___
### **What This PoC Demonstrates**

- Non-intrusive quality engine: collect quality snapshots at each step in the executor loop without modifying any of the 15+ existing operator functions
- Opt-in by design: include_intermediates defaults to false so all existing API consumers are unaffected
- Multi-format support: I wrote integration tests verifying that all 5 formats (CSV, XLSX, JSON, Parquet, TSV) ingest and export correctly with full round-trip fidelity
- Partial results on failure: if an operator crashes mid-pipeline, I return all intermediates captured before the failure
- Zero new dependencies: I implemented everything within pandas, numpy, and openpyxl/pyarrow, which are already in pyproject.toml
- Strong foundation: this captures logic is the single shared infrastructure that all 5 GSoC features (quality engine, formula columns, visualization, merge/join, checkpoint DAG) will build on
___
### **Testing**
I wrote 132 tests that verify:

- run_quality_assessment() returns correct profiles, duplicate reports, outlier reports, fix suggestions, and composite scores for both numeric and string columns
- Quality score improves measurably after apply_fix() is called on a dataset with known issues
- All 5 format loaders produce correct DataFrame shape, types, and values from fixture files
- Round-trip fidelity: upload XLSX → export Parquet → reload Parquet matches original
- Pipeline with include_intermediates: false returns identical response to original API
- Pipeline with include_intermediates: true returns one quality snapshot per operator step
- Intermediate image values are pixel-exact matches to manual sequential execution
- Final results are returned when an operator raises mid-pipeline
- Response schema validates correctly via Pydantic

All 132 new tests pass. All previously-passing tests still pass. Ruff lint + format clean.
___
###  **Running It**
```
bashcd dataloom-backend
uv run pytest tests/test_format_handler.py tests/test_quality_engine.py -v
```
___
### **Next Steps**
This PoC proves the core quality and ingestion infrastructure works. In the full GSoC implementation I would:

- Add <CheckpointTimeline> and <QualityDashboard> React components to surface intermediates and quality results in the frontend
- Add pipeline-level snapshot caching (Parquet snapshots every N operations) to bound replay time for long transform histories
- Wrap the executor in a concurrent batch loop with progress tracking and ZIP download for multi-file sessions
- Implement macro expansion so composite operations expand to their constituent steps before execution

### DIFF
--- a/POC_README.md
+++ b/POC_README.md
@@ -1,0 +1,96 @@
+# DataLoom PoC -- Multi-Format Ingestion & Automated Quality Engine
+
+Proof-of-concept for the GSoC 2026 project:
+*"DataLoom -- A Web-Based Data Wrangling Tool with Git-Like Checkpoint/Revert, Multi-Format Support, Data Profiling, Formula Columns, Visualization, and Automated Quality Engine"*
+
+## What This Demonstrates
+
+This PoC implements the two foundational backend modules that the rest of the
+project builds on, following DataLoom's existing code conventions exactly.
+
+### 1. Multi-Format Ingestion & Export (`app/services/format_handler.py`)
+
+Loads **CSV, XLSX, JSON, Parquet, and TSV** files into pandas DataFrames and
+exports DataFrames back to any of those formats. Each loader handles
+format-specific edge cases (encoding fallback for CSV/TSV, orientation
+auto-detection for JSON, openpyxl engine for XLSX, pyarrow engine for Parquet).
+
+This is the foundation for the project -- DataLoom currently only accepts CSV.
+Without multi-format support, the tool cannot be the first step in a real data
+pipeline.
+
+### 2. Automated Data Quality Engine (`app/services/quality_engine.py`)
+
+The core mechanism of the project. Given a DataFrame, the engine:
+
+1. **Profiles** every column (null %, unique %, min/max/mean/std/median, top values)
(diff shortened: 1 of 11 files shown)


ivantha:
Thank you for the multi-format ingestion PoC! This work has been superseded by #183 which provides production-ready multi-format upload. Feel free to follow up on that PR with any data quality engine improvements.

PR #221, opened 2026-03-22
fix(backend): preserve response semantics and harden addCol validation

### LINKED ISSUE #220: Backend response serialization coerces missing values to empty strings, losing null semantics
## Summary

The shared backend response serializer currently collapses missing values into empty strings before returning API payloads. This happens in `dataframe_to_response()` in `dataloom-backend/app/utils/pandas_helpers.py`, where the DataFrame is normalized with `fillna("")`.

That means API consumers cannot distinguish a true missing value (`null`) from a real empty string (`""`).

## Why this matters

This affects every endpoint that uses the shared serializer, including project fetch/upload responses and transform responses. It also creates downstream ambiguity for:

- profiling and null-count correctness
- data-quality checks
- formula behavior
- export/report fidelity
- frontend rendering and edit flows that should preserve the difference between `null` and `""`

## Current behavior

A DataFrame like this:

- `None` in one cell
- `""` in another cell

is serialized so both values become `""` in the API response.

## Expected behavior

- Missing values should remain `null` in API responses
- Real empty strings should remain `""`
- Any display-layer blank rendering should happen in the frontend, not in backend serialization

## Suggested direction

Normalize response values at the serializer boundary instead of calling `fillna("")`, preserving `null` semantics while still keeping the payload JSON-safe.

This is a small backend-only change, but it improves correctness for future profiling, quality, formula, and export work.

nodesagar:
## Summary

This PR fixes two backend contract problems discovered while auditing the current transform and serialization flows:

1. shared API response serialization was collapsing missing values into empty strings, which destroyed the distinction between `null` and `""` across upload, project fetch, and transform responses
2. `addCol` requests without a `name` were slipping past schema validation and failing later as response-validation errors instead of being rejected at the request boundary

### Response serialization
- updated `dataframe_to_response()` to preserve real empty strings while normalizing missing and non-finite values to `null`
- added a focused regression test in `tests/test_dtypes.py`
- added an upload-endpoint regression test in `tests/test_response_serialization.py`

### addCol validation hardening
- made `AddColumn.name` required in `schemas.py`
- reject blank or whitespace-only names at schema validation time
- switched the transform logging path from deprecated `dict()` to `model_dump()` while touching that code

## Why this matters

Preserving null semantics improves correctness for downstream profiling, quality checks, formulas, and export/report work. Tightening `addCol` validation fixes a pre-existing backend failure and keeps the full test suite green.

Closes #220

### Backend
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pytest -q`

### Frontend
- `npm run lint`
- `npm run test`

All of the above pass locally.

### DIFF
--- a/dataloom-backend/app/api/endpoints/transformations.py
+++ b/dataloom-backend/app/api/endpoints/transformations.py
@@ -168,7 +168,7 @@ async def transform_project(
 
     if should_save:
         save_csv_safe(result_df, project.file_path)
-        log_transformation(db, project_id, transformation_input.operation_type, transformation_input.dict())
+        log_transformation(db, project_id, transformation_input.operation_type, transformation_input.model_dump())
 
     resp = dataframe_to_response(result_df)
     return {
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -5,7 +5,7 @@ import uuid
 from enum import StrEnum
 from typing import Any
 
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, ConfigDict, field_validator
 
 # --- Enums ---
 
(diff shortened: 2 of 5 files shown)


ivantha:
This PR's branch is behind main and needs to be rebased. Please run `git fetch origin && git rebase origin/main`, resolve any conflicts, and force-push to update this PR.

ivantha:

PR #225, opened 2026-03-23
fix(checkpoints): preserve cumulative transforms across repeated saves

### LINKED ISSUE #224: Repeated saves drop previously checkpointed transforms by rebuilding from only pending logs
## Summary

`POST /projects/{project_id}/save` currently rebuilds the project state from the original CSV plus only the unapplied change-log entries. After a project has already been saved once, that means a later save can silently discard earlier checkpointed transforms.

## Reproduction

1. Upload a CSV project.
2. Apply a transform such as `renameCol(age -> years)`.
3. Save the project to create the first checkpoint.
4. Apply a second transform such as `addCol(country)`.
5. Save again.

### Current behavior
The second save response (and persisted working copy) contains `age` again instead of `years`, because the save flow replays only the new unapplied `addCol` log onto the original CSV.

### Expected behavior
A later save should preserve the full current working dataset, including transforms that were already checkpointed earlier.

## Why this matters

This is a core checkpoint/history correctness bug. DataLoom's main differentiator is save/revert over a working dataset; if a second save silently drops an earlier transform, users can lose data shape without any explicit error.

## Suggested direction

Use the current working copy as the source of truth when creating a checkpoint, and let checkpoint creation only update metadata by marking pending logs as applied. A focused regression test can cover the two-save case to prevent future regressions.

nodesagar:
## Summary

This PR fixes a checkpoint/history regression where a later save could silently discard transforms that were already included in an earlier checkpoint.

Before this patch, `POST /projects/{project_id}/save` rebuilt project state from the original CSV plus only the unapplied change-log entries. In a flow like `renameCol -> save -> addCol -> save`, the second save dropped the earlier rename because only the pending `addCol` log was replayed.

This patch makes checkpoint creation preserve the current working copy as the source of truth and only use save to mark pending logs as applied.

Closes #224.

## What changed

- updated `save_project()` to read the current working copy instead of rebuilding from only the unapplied log tail
- added a focused regression test covering the two-save case where an earlier rename must survive a later save
- replaced the `assert` in the save path with an explicit 500 error if the working copy path is misconfigured

## Validation

- `ruff check .`
- `ruff format --check .`
- `pytest tests/test_save_revert.py -q`
- `pytest -q -k 'not test_add_column_without_name_returns_422'`
- `npm run lint`
- `npm run test -- --run`
- manual two-save repro confirmed the final saved columns remain `['name', 'years', 'city', 'country']`
- manual revert repro confirmed the latest checkpoint still restores `['name', 'years', 'city', 'country']`

## Note on baseline

`main` currently has one unrelated backend test failure in `tests/test_new_features.py::TestAddDeleteColumnEndpoint::test_add_column_without_name_returns_422`, which belongs to the separate addCol validation bug fixed in PR #221. Excluding that known baseline failure, the remaining backend suite passed locally on this branch.

### DIFF
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -94,38 +94,24 @@ async def save_project(
     commit_message: str,
     db: Session = Depends(database.get_db),
 ):
-    """Save project changes as a checkpoint.
+    """Save the current working dataset as a checkpoint.
 
-    Replays all pending transformations from the change log onto the original
-    file and creates a checkpoint record marking the save point.
+    The working copy already reflects the user's latest accepted transforms, so
+    checkpoint creation should preserve that current dataset and only update the
+    checkpoint/log metadata for pending actions.
     """
     project = get_project_or_404(project_id, db)
 
-    # Load original file for replaying transformations
     original_path = get_original_path(project.file_path)
-    df = read_csv_safe(original_path)
-
-    # Get all unapplied logs for this project
-    logs = (
-        db.query(models.ProjectChangeLog)
-        .filter(
-            models.ProjectChangeLog.project_id == project_id,
-            models.ProjectChangeLog.applied == False,  # noqa: E712
+    if project.file_path == str(original_path):
+        logger.error(
+            "Project working copy unexpectedly points at original file: id=%s path=%s",
+            project_id,
+            project.file_path,
         )
-        .order_by(models.ProjectChangeLog.timestamp)
-        .all()
-    )
+        raise HTTPException(status_code=500, detail="Project working copy is misconfigured")
 
-    # Replay each logged transformation on the original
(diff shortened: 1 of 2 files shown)


ivantha:
This PR's branch is behind main and needs to be rebased before it can be merged. Please run `git fetch origin && git rebase origin/main`, resolve any conflicts, and force-push to update this PR.

nodesagar:
@ivantha 

I’ve updated this branch to address the test-layer concern.

Changes made:
- Reworked the regression in `tests/test_save_revert.py` to call the real API route via `client.post("/projects/{id}/save", params={"commit_message": ...})` instead of invoking `save_project()` directly.
- Added explicit response status checks and kept the cumulative-columns assertions across both saves.

This now exercises FastAPI request parsing and response-model behaviour at the routing layer, as requested.

ivantha:

PR #226, opened 2026-03-25
fix: wrap alembic migration in try-except to prevent silent server crash

gitadi2:
## Problem
The Alembic auto-migration in the `lifespan` function causes the server to exit silently on Windows when the migration encounters issues. The server shows "Waiting for application startup" and then terminates without any error message, making it very difficult to debug.

## Solution
Wrapped the Alembic migration block in a try-except handler that catches the exception, logs the actual error message for debugging, and re-raises it so the server fails loudly instead of dying silently. This surfaces the root cause (missing `alembic.ini`, wrong working directory, DB connection issue) without masking schema errors.

## Testing
- Tested on Windows 11 with Python 3.12 and PostgreSQL
- If migration succeeds: server starts normally and serves API endpoints at /docs
- If migration fails: error is logged clearly and server exits with a visible error message

## Fixes 
Silent server crash on startup caused by Alembic auto-migration

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
The server was crashing silently on Windows after showing "Waiting for application startup" with no error message. After the fix, migration errors are logged and surfaced clearly.

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

### Steps to reproduce the bug:
1. Clone the repo on Windows with Python 3.12 + PostgreSQL
2. Run `uvicorn app.main:app --port 4200`
3. Server shows "Waiting for application startup" then exits silently — no error, no "Application startup complete"

### After the fix:
1. Run `uvicorn app.main:app --port 4200`
2. Server starts normally if migration succeeds
3. If migration fails, error is logged clearly instead of silently killing the server

## Screenshots 
[Server Crashes Silently](https://1drv.ms/i/c/87c7509185cc73aa/IQBxPbS0KfmjTJes0PDzDDEYAa5IUVxtGrUbE5iKpG945cA?e=RNXLp0)
[Server Starts Successfully 1](https://1drv.ms/i/c/87c7509185cc73aa/IQA0C970CV1uQqf7cYbPR2F6Adz1a9Ejn8aqtaIa1LP_5vw?e=9a3WI6)
[Server Starts Successfully 2](https://1drv.ms/i/c/87c7509185cc73aa/IQBLgKywOAgyTKjbPhMl5L1qAWCacBMwPCUVYbkxaUAjuQs?e=FmJrzX)
[FastAPI - Swagger UI](https://1drv.ms/i/c/87c7509185cc73aa/IQBJD_bA9MO7RIabcimjNhGsAXyrXumYcv3qigSy1c18edI?e=Uel2Ag)

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/main.py
+++ b/dataloom-backend/app/main.py
@@ -26,8 +26,12 @@ async def lifespan(app):
 
     from alembic import command
 
-    alembic_cfg = Config("alembic.ini")
-    command.upgrade(alembic_cfg, "head")
+    try:
+        alembic_cfg = Config("alembic.ini")
+        command.upgrade(alembic_cfg, "head")
+    except Exception as e:
+        logger.error("Alembic migration failed: %s", e)
+        raise
 
     settings = get_settings()
     setup_logging(settings.debug)


gitadi2:
Hi, I'm a first-time contributor and my workflow checks are awaiting approval. Could a maintainer please approve the CI runs? Thank you!

ivantha:
Thank you for investigating the Windows startup issue — but this approach of catching all exceptions in the migration block is too broad and dangerous for a production codebase.

Silently swallowing a migration failure means the server can start with a stale or incorrect schema, which can cause data corruption, unexpected query failures, or misleading errors for users. The right behavior is to surface migration errors loudly, not suppress them.

The underlying Windows issue is likely an environment problem (e.g., missing `alembic.ini`, wrong working directory when uvicorn is run, or a DB connection issue at startup). The fix should be to investigate and resolve the actual error rather than hide it.

If you're able to share the actual error message that appeared (perhaps by catching the exception and re-raising after logging, and checking logs), that would help identify the right fix. This PR as-is cannot be merged.

gitadi2:
Thanks for the detailed review @ivantha — you're absolutely right. Silently swallowing migration errors is dangerous and could mask schema issues.

I've updated the fix to:
1) Catch the exception and log the actual error message
2) Re-raise it so the server fails loudly with a clear error instead of starting with a broken schema

This surfaces the root cause (missing alembic.ini, wrong working directory, DB connection issue) while preventing the invisible silent crash.
The actual error on my Windows setup was the Alembic migration failing because of a fresh database with no prior migration history — the alembic.ini path resolution differed when running from the project root vs the backend directory. With this fix, that error is now clearly visible in logs instead of silently killing the server.

PR #227, opened 2026-03-25
fix: handle None drop_na_params during replay to prevent AttributeError

### LINKED ISSUE #199: dropNa replay crashes during save/revert when `drop_na_params` is `None`

gitadi2:
## Fixes 
#199

## Type of Change
- [x] Bug fix

## Problem
Projects containing a `dropNa` transformation without parameters cannot be saved or reverted. When `drop_na_params` is serialized as `None` by Pydantic, the replay handler calls `.get("columns")` on `None`, causing an `AttributeError`.

## Root Cause
The line:
```python
```
The default `{}` only applies when the key is **missing**. When the key exists with value `None`, `.get()` returns `None`, and calling `.get("columns")` on `None` crashes.

## Fix
```python
columns = (action_details.get("drop_na_params") or {}).get("columns")
```
Using `or {}` converts `None` to an empty dict, so `.get("columns")` always works.

## Testing
- [x] Manual testing
- [x] Existing tests pass

## Checklist

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -557,7 +557,7 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
         return trim_whitespace(df, column)
 
     elif action_type == "dropNa":
-        columns = action_details.get("drop_na_params", {}).get("columns")
+        columns = (action_details.get("drop_na_params") or {}).get("columns")
         return drop_na(df, columns)
 
     elif action_type == "melt":


ivantha:
The fix to `transformation_service.py` is correct — using `(action_details.get("drop_na_params") or {}).get("columns")` properly handles the `None` case and closes #199.

However, this PR also includes the same `main.py` change that wraps Alembic migration in a broad `except Exception` that silently swallows failures. That change is problematic (it lets the server start with a stale schema if migrations fail) and cannot be merged as-is.

Please revert the `main.py` change so this PR only contains the one-line `transformation_service.py` fix, then push an update — after that it should be ready to merge.

gitadi2:
Thanks for the review @ivantha — you're right, the main.py change shouldn't be in this PR. I've removed it and this PR now only contains the transformation_service.py fix for the None drop_na_params issue. The main.py migration fix is handled separately in #226.

PR #237, opened 2026-03-28
Fix delRow replay to use delete_row for RangeIndex consistency

### LINKED ISSUE #236: [Bug] delRow checkpoint replay uses df.drop without reset_index, diverging from live delete_row
## Summary
When a `delRow` action is **replayed** during save/checkpoint (`apply_logged_transformation`), the code used `df.drop(index)` only. The **live** transform path uses `delete_row()`, which calls `reset_index(drop=True)` after the drop.

That leaves replayed DataFrames with a non–`RangeIndex` in some sequences, while the UI and other operations (e.g. `change_cell_value` via `df.at[row_index, ...]`, which is **label-based**) assume consistent row indexing semantics.

## Expected behavior
Replay of `delRow` should match the live `delete_row()` behavior so row indices stay aligned (0 … n−1) after each delete.

## Actual behavior
Replay could produce a DataFrame whose index labels are not contiguous, diverging from the live path and risking wrong-row updates in later steps.

## Root cause
- Live: `delete_row` → `df.drop(index).reset_index(drop=True)`
- Replay (before fix): `df.drop(index)` only

## Proposed fix
In `apply_logged_transformation`, handle `delRow` by delegating to `delete_row(df, index)` (or equivalently always `reset_index(drop=True)` after drop).

## Tests
- Assert `apply_logged_transformation(..., "delRow", ...)` matches `delete_row(...)` (including `RangeIndex`).
- Optional: chained `delRow` replay ends with `RangeIndex(len(df))`.

## Environment
- Backend: `dataloom-backend`, `app/services/transformation_service.py`

RachanaB5:
Fixes #236

## Changes Made
- Modified apply_logged_transformation in 
  app/services/transformation_service.py
- Changed df.drop(index) to 
  delete_row(df, index) to match live behavior
- Added reset_index(drop=True) for consistency

## Tests
All 4 existing tests pass ✅

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -511,9 +511,7 @@ def apply_logged_transformation(df: pd.DataFrame, action_type: str, action_detai
 
     elif action_type == "delRow":
         index = action_details["row_params"]["index"]
-        if index < 0 or index >= len(df):
-            raise TransformationError(f"Row index {index} out of range")
-        return df.drop(index)
+        return delete_row(df, index)
 
     elif action_type == "addCol":
         params = action_details.get("add_col_params") or action_details.get("col_params")
--- a/dataloom-backend/tests/test_transformations.py
+++ b/dataloom-backend/tests/test_transformations.py
@@ -9,6 +9,7 @@ from app.services.transformation_service import (
     add_row,
     advanced_query,
     apply_filter,
+    apply_logged_transformation,
     apply_sort,
     change_cell_value,
     delete_column,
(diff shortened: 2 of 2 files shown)


ivantha:
Thanks for the PR! It currently has merge conflicts with main (or is behind). Could you rebase onto the latest main and push the updated branch? Happy to merge once it's auto-rebasable.

RachanaB5:
Rebased onto the latest main. The branch is now up to date with no conflicts.

This change adds replay support for `delRow` in apply_logged_transformation, and it coexists with the recently added replay support for other transformations.

PR #270, opened 2026-04-23
fix(backend+frontend): preserve null semantics, harden addCol, and align addCol payload

nodesagar:
## Why
This PR supersedes closed PR #221, which became unmergeable after base branch drift.

### Backend behavior
- Preserve response semantics in dataframe serialization:
  - Keep real empty strings as ""
  - Serialize missing and non-finite values as null
- Harden `addCol` validation:
  - Reject blank/whitespace-only column names (422)
- Use Pydantic v2 APIs:
  - `model_dump()` for transformation logging
  - `ConfigDict(from_attributes=True)` for `LastResponse`

### Frontend alignment
- Send `add_col_params` (instead of legacy `col_params`) for add-column requests

### Backend CI stabilization (SQLite test environment)
- Add explicit startup toggle `SKIP_MIGRATIONS` and skip Alembic in tests
- Set `SKIP_MIGRATIONS=1` in test bootstrap
- Add `pytest-asyncio` to dev dependencies and lockfile
- Align outdated test expectations with current behavior:
  - integer dtype assertion
  - delRow replay assertion
  - unknown action replay expectation
  - user logs setup path

## Validation
- Local backend (CI-equivalent env):
  - `DATABASE_URL=sqlite:///./test.db uv run pytest -q`
  - Result: `167 passed`
- Frontend lint for touched file passed earlier in this PR flow.

## Reference
- Supersedes: #221


ivantha:
Thanks for reviving this from #221, @nodesagar. The diagnosis is correct, the null-vs-empty-string distinction was the right call, and the per-cell normalization using `astype(object).where(pd.notna, None)` is in fact vectorized (Copilot's perf concern about a "Python-level nested loop" misreads the code).

Closing this PR because most of the bundle is now in main via different paths, but the valuable substance below should land. Asking for a focused follow-up.

### What main now has

- **Pydantic v2 migration**: `ConfigDict(from_attributes=True)` and `model_dump()` are both in main via #282 (commit `5b195e1`). Your changes to `schemas.py:339` and `transformations.py:177` are no-ops on rebase.
- **`pytest-asyncio`**: already in `pyproject.toml:20`. Your dep add is redundant.
- **Frontend `add_col_params` payload**: I merged this via #302. Your `Table.jsx` change will conflict on rebase. The `delCol` legacy `col_params` (which Copilot flagged on line 108 of your diff) had also already been fixed on main by an intermediate commit, so that hunk is also now a no-op.
- **Alembic-on-pytest**: main solved this in `a5eb4c5` via `conftest.py` monkeypatching `verify_database_connection` and `command.upgrade`. That avoids the production-code `SKIP_MIGRATIONS` hook entirely. Your `main.py` and `conftest.py` changes would conflict.
- **Stale-test cleanups**: `test_del_row` has a stronger assertion in main than the version this PR rewinds to; `test_unknown_action_type_*` already raises (only the name is stale, which is fine to rename in a tiny follow-up).

### What's still genuinely valuable

1. **`pandas_helpers.py` null preservation**. `df.fillna("")` is still in main at `pandas_helpers.py:74-75`. The change to `astype(object).where(pd.notna, None)` is correct and Copilot's perf concern is incorrect: `where` over an `astype(object)` frame is vectorized C-level dispatch in pandas, not a Python loop. The new behavior is the right one for the API contract.

2. **`AddColumn.name` field_validator**. The 422-on-blank check is a clean schema-layer fix. Pair it with the regression test in `test_new_features.py`.

3. **`test_dtypes.py` additions** (`test_preserves_nulls_without_coercing_them_to_empty_strings` and `test_preserves_non_scalar_cells_without_isna_truthiness_errors`) and **`test_response_serialization.py`** (new file) are good regression coverage.

### What needs coordination

The null-preservation change is a user-visible API contract change. The frontend's `Table.jsx:233` currently does `setEditValue(cellValue)` and will now receive `null` instead of `""` for missing cells. That triggers a React controlled-input warning and may display the literal string "null" in the edit input. Please add `setEditValue(cellValue ?? "")` (or equivalent) to the follow-up PR.

### Asking for a slim follow-up

A clean PR with just:
- `app/utils/pandas_helpers.py`: null preservation change.
- `app/schemas.py`: `name_must_not_be_blank` validator on `AddColumn`.
- `dataloom-frontend/src/Components/Table.jsx`: `setEditValue(cellValue ?? "")` defensive update on line 233.
- New tests: `test_dtypes.py` additions, `test_response_serialization.py`, `test_new_features.py::test_add_column_with_blank_name_returns_422`.

That would be 5-6 files, focused, mergeable in one pass.

### On Copilot's review

- Concern #1 (delCol payload also broken on line 108) was valid at the time but has since been fixed on main by an intermediate commit. Both addCol and delCol are correct in main now.
- Concern #2 (performance regression on per-cell normalization) is a misread. `astype(object).where(pd.notna, None)` is vectorized; it does not loop in Python. The concern can be set aside.

Closing as superseded. Thank you for the careful work, especially the regression tests.

PR #235, opened 2026-03-27
fix(frontend): prevent invalid file uploads from triggering API calls

sanchit955:
### Fix: Prevent unnecessary API calls for invalid file uploads

Previously, selecting an unsupported or invalid file still triggered an API request to `/projects/upload`, which resulted in a backend 400 error. This caused unnecessary network calls and a poor user experience.

### Changes

* Added client-side file validation (extension + empty file check)
* Prevents API calls for invalid files
* Displays errors using toast notifications instead of relying on backend responses
* Added `accept` attribute to file input for better UX
* Added defensive validation before form submission

### Supported formats

.csv, .xlsx, .json, .parquet, .tsv

### Impact

* Improves UX by validating inputs early
* Reduces unnecessary backend load
* Keeps implementation minimal and consistent with existing codebase
* Aligns frontend behavior with supported dataset formats

Issue #76, opened 2026-02-25
GSoC 2026 — DataLoom: Multi-Source Data Profiling, Transformation Pipeline, and Visualization Engine


ivantha:
## Title

Multi-Source Data Profiling, Transformation Pipeline, Visualization Engine, Data Quality Analysis, and Multi-Format Export

## Abstract

DataLoom is a web-based data wrangling tool with a unique git-like checkpoint/revert system. Users upload a CSV, apply pandas-powered transformations (filter, sort, pivot, cast, etc.), and can save/revert to named checkpoints. However, the tool is currently limited to single CSV files, offers no data profiling, has no visualization, and requires users to type column names manually in every form.

This project expands DataLoom into a complete data preparation platform across seven pillars:

1. **Multi-format upload + data profiling** — support xlsx/json/parquet/tsv files; auto-profile on upload (dtype inference, null counts, unique values, value distributions)
2. **Profiling frontend + column selector refactor** — a DataProfilePanel showing column statistics, and a refactor of all 6 transform forms to use column dropdown selectors instead of free-text input
3. **Merge/join operations** — inner/left/right/outer/cross joins and concat between projects
4. **Formula columns + transformation pipelines** — safe expression engine via `pandas` DataFrame operations, plus reusable named pipelines (save a sequence of transforms, replay on new data)
5. **Data visualization panel** — histogram, bar chart, scatter plot, and time series via recharts
6. **Data quality engine** — automated quality assessment with duplicate detection (exact + fuzzy), outlier flagging (IQR/z-score/Isolation Forest), pattern validation, composite quality scores, and one-click fixes
7. **Multi-format export + quality reports** — export data as xlsx/parquet/json/tsv with format-specific options, and generate downloadable HTML/PDF quality reports combining profiling, quality scores, and auto-generated charts

## Goals

- Support multiple file formats beyond CSV
- Give users instant statistical insight into their data on upload
- Eliminate manual column-name typing in all forms
- Enable combining data from multiple projects via join/merge
- Allow users to define computed columns and save reusable transform sequences
- Provide basic but useful data visualization
- Automatically detect data quality issues (duplicates, outliers, pattern violations) and suggest fixes
- Export data in multiple industry-standard formats with downloadable quality reports

## Expected Outcomes

- Users can upload xlsx, json, parquet, and tsv files in addition to CSV
- Automatic data profiling gives instant insight into column statistics on upload
- All transform forms use dropdown column selectors instead of free-text input
[...]

PR #273, opened 2026-04-24
fix(frontend): prevent invalid file uploads from triggering API calls

sanchit955:
Fix: Prevent invalid file uploads from triggering API calls

Previously, invalid or unsupported files still triggered API requests to `/projects/upload`, resulting in backend errors and poor UX.

### Changes
- Added client-side file validation (empty file, size limit, supported formats)
- Prevented API calls for invalid inputs
- Added loading state to disable submit button during upload
- Improved UX with clear error messages and file input restrictions

### Testing
- Verified invalid files do NOT trigger API calls
- Verified valid files upload successfully with backend running via uv

### Note
Re-submitting this fix as a fresh PR after previous PR (#235) was closed due to outdated base.

### DIFF
--- a/dataloom-frontend/src/Components/Homescreen.jsx
+++ b/dataloom-frontend/src/Components/Homescreen.jsx
@@ -56,8 +56,6 @@ const NewProjectCard = ({ onClick }) => (
   </button>
 );
 
-const MAX_FILE_SIZE_BYTES = 10 * 1024 * 1024; // 10 MB
-
 const EmptyState = ({ onClick }) => (
   <div className="flex flex-col items-center justify-center py-16 px-6 rounded-xl border-2 border-dashed border-gray-200 bg-gray-50 text-center">
     <div className="mb-4 flex items-center justify-center w-16 h-16 rounded-full bg-blue-50">
@@ -80,7 +78,7 @@ const EmptyState = ({ onClick }) => (
     </div>
     <h3 className="text-lg font-semibold text-gray-800 mb-1">No projects yet</h3>
     <p className="text-sm text-gray-500 mb-6 max-w-xs">
-      Upload a CSV file to get started. Your recent projects will appear here.
+      Upload a dataset file to get started. Your recent projects will appear here.
     </p>
     <button
       type="button"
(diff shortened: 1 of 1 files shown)


sanchit955:
Hi maintainers,
All checks are passing and this PR fixes the issue where invalid file uploads were triggering unnecessary API calls.

Kindly review and approve the pending workflow so it can be merged. Thanks!

ivantha:
Thanks for picking this up, @sanchit955. The intent (skip the API round-trip for obviously invalid files) is sound, and the loading-state separation (`isUploading`) is a nice clarification.

I'm going to close this one, for two reasons.

### 1. The format expansion overlaps with GSoC #76 scope

The PR moves `ALLOWED_EXTENSIONS` from `[".csv"]` to `[".csv", ".xlsx", ".json", ".parquet", ".tsv"]` and updates the `<input accept>`, the help text, and the empty-state copy to match. That is pillar 1 ("Multi-format upload + data profiling") of the DataLoom GSoC 2026 scope reserved in #76, and it is currently reserved for our selected student. Shipping it now (especially while the backend still only handles CSV) would both front-run the GSoC scope and create misleading UX where users can pick an xlsx file and get a backend 400.

Per our workspace convention on `gsoc`-labeled issues, any PR touching the reserved scope gets redirected regardless of author.

### 2. Several non-scope parts of the PR are already in `main` or are regressions

- The `handleCloseModal` TDZ reorder is already in `main` via #281 (commit 50422d8), so the base of this PR is stale on that axis.
- Reducing `MAX_FILE_SIZE_BYTES` from 10 MB to 5 MB is an unannounced regression. No rationale was given, and the help-text "Maximum file size: 10 MB" elsewhere in the file would silently become a lie.
- The `isSubmitting` -> `isUploading` rename is pure churn and adds review noise around the actual functional change.
- The PR's base predates the drag-and-drop upload UX in #289 (commit 1196347), so the input element and surrounding markup would conflict significantly on rebase.

### If you want a scoped, mergeable version

- Keep the `validateFile` helper and the "don't fire the API call on a bad file" guard.
- Keep `[".csv"]` as the only allowed extension until pillar 1 lands.
- Do not touch `MAX_FILE_SIZE_BYTES`. If you have a reason to lower it, open a separate issue first.
- Drop the rename and the reorders that are already in `main`.
- Rebase on current `main` and preserve the drag-and-drop input shape.

That would be a small, focused PR. Happy to look at a reopen along those lines.

For other ways to contribute before GSoC selection, the `pre-gsoc` label on the issues board lists self-contained tasks that are outside the #76 scope.

Closing.

sanchit955:
Addressed the feedback in a smaller focused follow-up PR: #317

PR #275, opened 2026-04-25
fix(checkpoints): preserve cumulative transforms across repeated saves

nodesagar:
## Summary
This PR revives and updates the checkpoint cumulative-save fix from previously closed PR #225, rebased onto the latest `main` so it is mergeable and isolated.

## Why This Change Is Needed
On repeated saves, previously checkpointed transformations could be lost or not reflected correctly in subsequent saved states. This can produce inconsistent checkpoint history and incorrect project state during save/revert workflows.

## Root Cause
The save flow did not consistently treat the current working copy as the source of truth across repeated checkpoint operations, which allowed cumulative transformation state to drift.

## What This PR Changes
- Preserves cumulative transforms across repeated saves.
- Keeps checkpoint/log metadata updates aligned with the current working dataset.
- Adds regression coverage for repeated-save behavior to prevent future regressions.

## Validation
- Ran targeted save/revert regression tests.
- Result: relevant tests passed for repeated-save checkpoint flow.

## Impact and Risk
- Scope is focused on checkpoint save behavior.
- No API contract expansion.
- Low risk outside save/revert paths, with regression tests included.

## References
- Supersedes closed PR #225.
- Context: #225 was closed during backlog cleanup after becoming non-mergeable as `main` advanced; this PR reintroduces the same intent on the latest base.

### DIFF
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -4,6 +4,7 @@ Handles upload, retrieval, save (checkpoint), and revert operations.
 """
 
 import uuid
+from pathlib import Path
 
 from fastapi import APIRouter, Depends, File, Form, HTTPException, UploadFile
 from fastapi.responses import FileResponse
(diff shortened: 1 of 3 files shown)


ivantha:
Thanks @nodesagar. The diagnosis here is right: `apply_logged_transformation` at `transformation_service.py:595-597` raises on any action_type it does not recognize, and four `OperationType` values (`filter`, `sort`, `advQueryFilter`, `pivotTables`) are not in the dispatcher, so the old save flow would have crashed for any project that used those operations. Switching to "the working copy is the source of truth" sidesteps that completely and is the right call.

Engaging Copilot's two inline notes:
- **Path comparison**: addressed by using `Path(...).resolve()` on both sides before equality. Good.
- **Error detail and logging**: the log line now includes `project_id`, `project.file_path`, and `original_path`; the client-side detail references `project_id` and suggests retry / contact support. Sufficient for an unexpected misconfiguration that should never happen in normal use.

One thing this PR does not address: the revert endpoint still uses `apply_logged_transformation` (`projects.py:219`), so any project that has used `filter` / `sort` / `advQueryFilter` / `pivotTables` will fail to revert. Worth a follow-up issue but not a blocker for this fix.

Rebased on current `main` and resolved the two stale-test conflicts in `tests/test_transformations.py` (kept main's stronger `test_del_row` assertion since it already covers the `RangeIndex` shape; took the new name `test_unknown_action_type_raises_transformation_error` to match the body). Lint and 167 tests pass locally. Merging.

PR #278, opened 2026-04-29
fix(upload): single-source size limit from config, chunked validation in store_upload

### LINKED ISSUE #217: File validation loads entire file into memory causing potential memory issues
**Describe the bug**
The current file validation logic reads the entire uploaded file into memory to determine its size. This can lead to high memory usage and potential crashes when handling large files.

**To Reproduce**
1.Upload a large CSV file (e.g., >50MB)
2.Trigger the /upload API
3.Observe memory usage increasing significantly

**Expected behavior**
File size validation should not load the entire file into memory. Instead, it should process the file in chunks and reject oversized files early.

**Environment:**
- OS:  Microsoft Windows 11 10.0.26200 N/A Build 26200
- Browser: 146.0.7680.81
- Node.js version:  v20.19.5
- Python version: [e.g. 3.13.12]

**Additional context**
This issue affects scalability and can lead to performance degradation or crashes under large file uploads.

nodesagar:
### Problem

The upload path has two separate validation points with inconsistent behavior:

1. **`validate_upload_file()` in `security.py`** — enforces the configured 10 MB limit from `settings.max_upload_size_bytes`, but reads the **entire file into memory** (`file.file.read()`) before checking. This creates a memory spike on large uploads.
2. **`store_upload()` in `file_service.py`** — has **no validation at all**. If called directly (bypassing `validate_upload_file`), any file of any type and any size is accepted and written to disk.

#### `app/utils/security.py`
- Replaced `contents = file.file.read()` (full memory load) with a **64 KB chunked streaming loop**
- Rejects mid-stream as soon as cumulative bytes exceed the config limit — no need to read the entire file
- Cursor reset preserved after validation

#### `app/services/file_service.py`
- Added **extension validation** (`.csv` only) and **chunked size enforcement** as defense-in-depth
- Uses `get_settings().max_upload_size_bytes` — same single source of truth
- Raises `ValueError` (not `HTTPException`) to match service-layer conventions
- Pointer reset preserved so `shutil.copyfileobj` writes complete content

#### `tests/test_file_service.py` [NEW]
- Extension validation tests (csv, uppercase, non-csv, exe, no extension)
- Size limit tests (under, exact, over) using config value
- Chunked validation tests (multi-chunk reject, exact-limit accept)
- Pointer reset tests (small file, multi-chunk file written completely)

#### `tests/test_upload.py`
- Added `test_chunked_streaming_rejects_oversized_file` — verifies HTTP 413 via chunked path
- Added `test_chunked_streaming_cursor_reset_after_validation` — verifies cursor at 0 after multi-chunk validation

### Test Results

```
28 passed, 1 warning in 0.87s
```

- ✅ Exact-limit accept
- ✅ Over-limit reject
- ✅ Chunked validation path
- ✅ Pointer reset and full write after validation
- ✅ ruff check — all passed
- ✅ ruff format — all formatted

### DIFF
--- a/dataloom-backend/app/services/file_service.py
+++ b/dataloom-backend/app/services/file_service.py
@@ -3,22 +3,20 @@
 import shutil
 from pathlib import Path
 
+from app.config import get_settings
 from app.utils.logging import get_logger
 from app.utils.security import resolve_upload_path, sanitize_filename
 
 logger = get_logger(__name__)
 
-# Maximum allowed upload size: 50 MB
-MAX_FILE_SIZE = 50 * 1024 * 1024
-
 
 def store_upload(file) -> tuple[Path, Path]:
     """Store an uploaded file and create a working copy.
 
-    Validates the file extension (must be .csv) and size (max 50 MB)
-    before writing anything to disk. Saves the file with a sanitized name
-    and creates a _copy.csv for transformation operations, keeping the
-    original pristine.
+    Validates the file extension (must be .csv) and enforces the configured
+    ``max_upload_size_bytes`` limit via chunked streaming before writing
+    anything to disk.  The file pointer is reset to 0 after validation so
+    ``shutil.copyfileobj`` writes a complete file.
 
     Args:
         file: The FastAPI UploadFile object.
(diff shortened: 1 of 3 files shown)


ivantha:
Thanks @nodesagar. The change is correct and overdue: `store_upload()` was reading the full 10 MB into memory just to check size, and `MAX_FILE_SIZE` was hardcoded to 50 MB while the config-driven `validate_upload_file()` in `security.py` already enforced 10 MB. Moving everything to `settings.max_upload_size_bytes` removes the inconsistency, and the 64 KB chunked loop avoids the memory spike. The pointer-reset and `shutil.copyfileobj` path is preserved correctly.

Test coverage looks solid:
- Multi-chunk reject covers the accumulating size path.
- Exact-limit accept covers the off-by-one boundary.
- `test_full_write_after_chunked_validation` verifies the seek(0) reset.
- `test_chunked_path_*` in `test_upload.py` covers the `validate_upload_file()` chunked branch via `MockUploadFileNoSize`.

Rebased on current `main` and resolved the same `test_transformations.py` conflicts as #275 (kept main's stronger `test_del_row` assertion, used main's existing `test_unknown_action_type_raises_transformation_error` name). Lint and 172 tests pass locally. Merging.

PR #280, opened 2026-05-02
feat: add search/filter bar to Projects section on Home Screen

### LINKED ISSUE #252: [Feature] Add search/filter bar to Projects section on Home Screen
## Description

As users create more projects in DataLoom, finding a specific project 
becomes increasingly difficult. Currently there is no way to search or 
filter projects by name — users must visually scan through all project 
cards to find the one they need. This becomes a real pain point once 
a user has 10+ projects.

## Current Behavior

- All projects are displayed in a grid with no filtering capability
- Users must scroll and visually scan to find a specific project
- There is no search input anywhere on the Home Screen
- No keyboard shortcut to quickly jump to a project

Image

## Expected Behavior

- A search bar appears below the "Projects" heading
- As the user types, project cards filter in real time by project name
- If no projects match the search query, a "No projects found" message 
  is shown
- Search is case-insensitive
- Clearing the search input restores all project cards

## Design Spec

- Placement: between the "Projects" `<h2>` heading and the project grid
- Placeholder text with a search icon inside : `Search projects...`
- Subtle border with focus ring matching existing input styles
- "No results" empty state when search yields no matches:
  - Text: `No projects match "<query>"`
  - Subtext: `Try a different name`

## Implementation Approach

- Add a `searchQuery` state in `HomeScreen.jsx`
- Filter `recentProjects` array using `project.name.toLowerCase()
  .includes(searchQuery.toLowerCase())`
- Render filtered list instead of full `recentProjects` array
- Keep `+ New Project` card always visible regardless of search query
- Show "No results" empty state when filtered list is empty but 
  `searchQuery` is not empty

## Files to Modify

- `src/components/HomeScreen.jsx` — add search input and filter logic

## Why This Matters

DataLoom is designed to help users manage multiple datasets. As usage 
grows, navigating a large list of projects without search becomes 
frustrating. This is a zero-dependency, pure frontend improvement that 
directly improves the core workflow of the application — and is 
especially important for non-technical users who may not remember 
exact project names.

## Additional Context

- No backend changes required — filtering happens client-side
- No new dependencies required
- Minimal diff — only `HomeScreen.jsx` needs to be modified
- Pairs well with the recently merged project card improvements

Pranavvv3004:
Closes #252

## Description

Fixes #252 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/Homescreen.jsx
+++ b/dataloom-frontend/src/Components/Homescreen.jsx
@@ -95,6 +95,7 @@ const EmptyState = ({ onClick }) => (
 const HomeScreen = () => {
   const fileInputRef = useRef(null);
   const [showModal, setShowModal] = useState(false);
+  const [searchQuery, setSearchQuery] = useState("");
   const [fileUpload, setFileUpload] = useState(null);
   const [projectName, setProjectName] = useState("");
   const [projectDescription, setProjectDescription] = useState("");
@@ -106,19 +107,15 @@ const HomeScreen = () => {
 
   const isFormValid =
     projectName.trim().length > 0 && projectDescription.trim().length > 0 && fileUpload !== null;
+    const filteredProjects = recentProjects.filter((project) =>
+  project.name.toLowerCase().includes(searchQuery.toLowerCase())
+);
 
   useEffect(() => {
     fetchRecentProjects();
   }, []);
 
-  useEffect(() => {
-    if (!showModal) return;
-    const handleKeyDown = (e) => {
-      if (e.key === "Escape" && !isSubmitting) handleCloseModal();
-    };
-    window.addEventListener("keydown", handleKeyDown);
-    return () => window.removeEventListener("keydown", handleKeyDown);
-  }, [showModal, isSubmitting, handleCloseModal]);
+  
 
   const fetchRecentProjects = async () => {
     try {
@@ -142,6 +139,14 @@ const HomeScreen = () => {
       fileInputRef.current.value = "";
     }
   }, []);
+  useEffect(() => {
+    if (!showModal) return;
+    const handleKeyDown = (e) => {
+      if (e.key === "Escape" && !isSubmitting) handleCloseModal();
+    };
+    window.addEventListener("keydown", handleKeyDown);
+    return () => window.removeEventListener("keydown", handleKeyDown);
+  }, [showModal, isSubmitting, handleCloseModal]);
 
   const handleSubmitModal = async (event) => {
     event.preventDefault();
@@ -263,12 +268,22 @@ const HomeScreen = () => {
           )}
         </div>
 
-        {recentProjects.length === 0 ? (
+        <div className="mb-4">
+  <input
+    type="text"
+    placeholder="Search projects..."
+    value={searchQuery}
+    onChange={(e) => setSearchQuery(e.target.value)}
+    className="w-full text-sm text-gray-900 border border-gray-300 rounded-md px-3 py-2 bg-white focus:outline-none focus:ring-2 focus:ring-blue-500 focus:border-blue-500"
+    aria-label="Search projects"
+  />
+</div>
+{recentProjects.length === 0 ? (
           <EmptyState onClick={handleNewProjectClick} />
         ) : (
           <div className="grid grid-cols-2 md:grid-cols-4 gap-4">
             <NewProjectCard onClick={handleNewProjectClick} />
-            {recentProjects.map((project) => (
+            {filteredProjects.map((project) => (
               <ProjectCard
                 key={project.project_id}
                 project={project}
--- a/dataloom-frontend/src/Components/Homescreen.jsx
+++ b/dataloom-frontend/src/Components/Homescreen.jsx
@@ -107,15 +107,13 @@ const HomeScreen = () => {
 
   const isFormValid =
     projectName.trim().length > 0 && projectDescription.trim().length > 0 && fileUpload !== null;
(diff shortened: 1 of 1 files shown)


Pranavvv3004:
Here's a screenshot of the search/filter bar in action:

Screenshot 2026-05-02 213840

ivantha (changes requested):
Thanks for the simple, scoped search bar, @Pranavvv3004. The filter logic is clean and the input styling matches the rest of the form controls.

A few things before this can land.

### 1. Search bar is shown even when there are no projects

The conditional rendering at lines 270-281 of the new diff puts the search input BEFORE the `recentProjects.length === 0 ? <EmptyState> : <grid>` check. That means a brand-new user with zero projects sees both a "Search projects..." input with nothing to search AND the empty state. Hide the search bar when `recentProjects.length === 0`.

### 2. No feedback when the search filters everything out

If the user types a query that matches no projects, the grid renders only the "New Project" tile with no message explaining that the search excluded all results. Please add a small "No projects match your search" message inside the grid when `recentProjects.length > 0 && filteredProjects.length === 0`.

### 3. Search is only over `/projects/recent` (capped at 10)

`/projects/recent` returns at most 10 projects (`projects.py:100` calls `get_recent_projects(db, limit=10)`). A user with more than 10 projects will not be able to find older ones via this search. Either:
- Document this in the PR description (e.g. "searches the 10 most recent projects") so reviewers know the intent is to search the visible list, or
- Wire up a separate endpoint or query parameter to fetch all projects when searching.

Option (a) is fine for an MVP. Option (b) is the eventual right answer. Pick one and call it out.

### 4. Stale base for Homescreen.jsx

The `handleCloseModal` reorder at lines 109-117 / 140-148 of the diff is already in `main` via #281 (commit `50422d8`). Drop that hunk on rebase; only the search input, state, and filter belong in this PR.

### 5. PR description is template-only

The Description, Type of Change, Testing, Screenshots, Checklist sections are unfilled. Please fill in a short description (one paragraph), add a screenshot of the search bar in action, and tick the relevant checklist items so reviewers can quickly confirm scope.

### 6. No tests

A Vitest test in `src/Components/__tests__/Homescreen.test.jsx` covering:
- Typing in the search bar updates the input value.
- The filtered grid only renders projects whose name (case-insensitively) contains the query.
- An empty query renders all projects.

### 7. (Nice-to-have) Also search description

Currently only `project.name.toLowerCase().includes(...)` is matched. Users often remember a project by its description text rather than its name. Consider:
```js
const q = searchQuery.toLowerCase();
const filteredProjects = recentProjects.filter(
  (p) => p.name.toLowerCase().includes(q) || (p.description ?? "").toLowerCase().includes(q),
);
```
Optional. Not a blocker.

---

Happy to re-review once items 1, 2, and 4 are addressed.

ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `dataloom-frontend/src/Components/Homescreen.jsx:140` Unrelated scope creep: the PR relocates the Escape-key useEffect (the block that adds/removes the keydown listener) from above fetchRecentProjects to just after handleCloseModal's useCallback definition. This has nothing to do with the search/filter feature, is behavior-neutral (handleCloseModal is a const-bound useCallback already in the dep array), and pollutes the diff. The feature itself only needs the searchQuery state, the filteredProjects computation, the input element, and swapping recentProjects.map -> filteredProjects.map. Recommend dropping the hook move to keep the diff focused.
- **[minor/quality]** `dataloom-frontend/src/Components/Homescreen.jsx:279` No empty-results state. The render branches only on recentProjects.length === 0. When recentProjects is non-empty but the search query matches nothing, filteredProjects is empty and the grid renders just the NewProjectCard with no 'no matching projects' feedback, so the user cannot tell whether the filter worked or the data failed to load. Add a 'no results' message when filteredProjects.length === 0 && searchQuery is non-empty.
- **[nit/bug]** `dataloom-frontend/src/Components/Homescreen.jsx:111` filteredProjects calls project.name.toLowerCase() with no guard. This is currently safe because the backend model defines name: str = Field(index=True) (non-nullable, app/models.py:24) and matches the existing unguarded {project.name} render in ProjectCard, so it will not throw in practice. Flagging only as defensive hardening (e.g. (project.name || '').toLowerCase()) should the schema ever allow null/empty names.
- **[nit/test]** `dataloom-frontend/src/Components/Homescreen.jsx:110` No test added for the new filter behavior, and the PR description's 'New tests added' box is unchecked (only manual testing claimed). There is no existing Homescreen test file, so this is a low-priority gap rather than a regression, but a small test asserting that typing in the search box narrows the rendered ProjectCards would lock in the behavior.

PR #280 adds a client-side, case-insensitive name filter over the recent-projects list on the Home Screen (searchQuery state + filteredProjects + a search input, swapping recentProjects.map for filteredProjects.map in Homescreen.jsx). It does not touch the reserved GSoC scope in issue #76, so there is no fencing concern and it can be reviewed on its own merits as a small enhancement. Two things to address before merge: it relocates the unrelated Escape-key useEffect, which should be reverted to keep the diff focused, and it shows no empty-results message when the query matches nothing (only the New Project card renders). Minor extras: the project.name filter is unguarded (safe today since the backend enforces a non-null name) and no test covers the new filtering.

ivantha:
Closing due to inactivity.

PR #297, opened 2026-05-08
Fix: update failing tests to match current implementation behaviour

SHIVAM-KUMAR-59:
## Description
Fixes two pre-existing test failures in the test suite that were unrelated 
to recent changes.

1. `test_del_row` — updated assertion to expect reset indexes `[0, 1]` 
   after row deletion instead of original indexes `[0, 2]`. The 
   `delete_row()` function correctly resets indexes after deletion — 
   the test expectation was outdated.

2. `test_unknown_action_type_returns_df_unchanged` — updated test to expect 
   a `TransformationError` for unknown action types instead of the dataframe 
   unchanged. The `apply_logged_transformation()` function correctly raises 
   the error — the test was outdated.

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- [x] Existing tests pass

All tests now pass locally:
```
uv run pytest
```

## Screenshots
**Before**

Screenshot 2026-05-08 at 17 59 51

**After**

Screenshot 2026-05-08 at 22 55 41

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review
- [x] I have added/updated documentation as needed
- [x] My changes generate no new warnings
- [x] Tests pass locally

Issue #177, opened 2026-03-09
[Feature] Authentication and user-scoped projects


hanzalahwaheed:
**Is your feature request related to a problem?**
The project has no auth. I think it was intended to add auth to the project as I can see a `Profile` button.

**Describe the solution you'd like**
Signin and signup cards where user can authenticate themselves via email and a password.

**Describe alternatives you've considered**
OAuth support can be added or we can use some other popular library as well.

hanzalahwaheed:
going with JWT-based auth

Wrt routing flow:
- default app view is already /projects
- unauthenticated users trying to access it get redirected to /signin
- After signin/signup, users can access /projects
- Project creation/read actions would require an authenticated project owner.


Wrt seeding the db,
- lets create a default dev/demo user during seeding
- attach all seeded projects to that user

Wrt schema/tables,
- new user table
- projects table gets an owner_id

hanzalahwaheed:
auth added successfully in this PR: https://github.com/c2siorg/dataloom/pull/299

PR #299, opened 2026-05-18
Feat: Add Authentication and Authorization

### LINKED ISSUE #76: GSoC 2026 — DataLoom: Multi-Source Data Profiling, Transformation Pipeline, and Visualization Engine

hanzalahwaheed:
## Description

Fixes #76 

## Type of Change

- [ ] Bug fix
- [X] New feature
- [X] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [X] New tests added
- [X] Manual testing

### Sign In Page
image

### Sign Up Page
image

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-backend/.env.example
+++ b/dataloom-backend/.env.example
@@ -12,3 +12,15 @@ MAX_UPLOAD_SIZE_BYTES=10485760
 
 # Debug mode
 DEBUG=false
+
+# Secret key for signing JWT auth tokens (REQUIRED, generate with: openssl rand -hex 32)
+JWT_SECRET=
+
+# JWT signing algorithm
+JWT_ALGORITHM=HS256
+
+# Auth token lifetime in hours
+JWT_EXPIRY_HOURS=24
+
+# Send the auth cookie only over HTTPS. Set to false for local http development.
+COOKIE_SECURE=true
--- a/dataloom-backend/alembic/versions/c3d4e5f6a7b8_add_users_and_project_owner.py
+++ b/dataloom-backend/alembic/versions/c3d4e5f6a7b8_add_users_and_project_owner.py
@@ -1,0 +1,78 @@
+"""add users and project owner
+
+Revision ID: c3d4e5f6a7b8
+Revises: b2c3d4e5f6a7
+Create Date: 2026-05-13 00:00:00.000000
+
+"""
+
+import secrets
+from collections.abc import Sequence
+
+import bcrypt
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "c3d4e5f6a7b8"
+down_revision: str | Sequence[str] | None = "b2c3d4e5f6a7"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+# System user owns legacy projects that predate authentication. It uses a bcrypt
+# hash of a random secret that is discarded inside _locked_password_hash(), so
+# no password can ever match and the account is unloginable. Dev/test users
+# belong in scripts/seed.py, never in the schema migration.
+SYSTEM_USER_ID = "00000000-0000-0000-0000-000000000000"
+SYSTEM_USER_EMAIL = "system@dataloom.local"
+
+
+def _locked_password_hash() -> str:
+    random_secret = secrets.token_urlsafe(32)
+    return bcrypt.hashpw(random_secret.encode("utf-8"), bcrypt.gensalt()).decode("utf-8")
+
+
+def upgrade() -> None:
+    op.create_table(
+        "users",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("email", sa.String(length=320), nullable=False),
+        sa.Column("password_hash", sa.String(length=1024), nullable=False),
+        sa.Column("created_at", sa.DateTime(), server_default=sa.func.now(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(op.f("ix_users_email"), "users", ["email"], unique=True)
+
+    bind = op.get_bind()
+    bind.execute(
+        sa.text("INSERT INTO users (id, email, password_hash, created_at) VALUES (:id, :email, :password_hash, now())"),
+        {
+            "id": SYSTEM_USER_ID,
+            "email": SYSTEM_USER_EMAIL,
+            "password_hash": _locked_password_hash(),
+        },
+    )
+
+    op.add_column("projects", sa.Column("owner_id", sa.Uuid(), nullable=True))
+    bind.execute(
+        sa.text("UPDATE projects SET owner_id = :owner_id WHERE owner_id IS NULL"),
+        {"owner_id": SYSTEM_USER_ID},
+    )
+    op.alter_column("projects", "owner_id", nullable=False)
+    op.create_index(op.f("ix_projects_owner_id"), "projects", ["owner_id"], unique=False)
+    op.create_foreign_key(
+        "projects_owner_id_fkey",
+        "projects",
+        "users",
+        ["owner_id"],
+        ["id"],
+        ondelete="CASCADE",
+    )
+
+
+def downgrade() -> None:
+    op.drop_constraint("projects_owner_id_fkey", "projects", type_="foreignkey")
+    op.drop_index(op.f("ix_projects_owner_id"), table_name="projects")
+    op.drop_column("projects", "owner_id")
+    op.drop_index(op.f("ix_users_email"), table_name="users")
+    op.drop_table("users")
--- a/dataloom-backend/app/api/dependencies.py
+++ b/dataloom-backend/app/api/dependencies.py
@@ -2,29 +2,67 @@
 
 import uuid
 
-from fastapi import Depends, HTTPException
+from fastapi import Cookie, Depends, HTTPException
 from sqlmodel import Session
 
 from app import database, models
+from app.services import auth_service
 from app.utils.logging import get_logger
 
 logger = get_logger(__name__)
 
 
-def get_project_or_404(project_id: uuid.UUID, db: Session = Depends(database.get_db)) -> models.Project:
-    """FastAPI dependency that fetches a project or raises 404.
+def get_current_user(
+    access_token: str | None = Cookie(default=None),
+    db: Session = Depends(database.get_db),
+) -> models.User:
+    """FastAPI dependency that resolves the authenticated user from the auth cookie.
+
+    Args:
+        access_token: The JWT auth token, read from the httpOnly cookie.
+        db: Injected database session.
+
+    Returns:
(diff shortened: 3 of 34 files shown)


hanzalahwaheed:
## Notes:

  This PR adds email/password authentication to DataLoom
  - backend (JWT + bcrypt)
  - frontend (sign-in/up pages, protected routing)
  - Projects are now owner-scoped: a user only sees and edits their own

  ### What changed
**Backend:**
  1. new `users` table
  2. `owner_id` column on `projects` table for owner-scoped project access
  3. `/auth/signup`, `/auth/signin`, `/auth/logout`, `/auth/me`, a `get_current_user` dependency
  4. Alembic migration with a seed test user
  
**Frontend:**
1. `SignInPage` / `SignUpPage`, `AuthContext` + `AuthProvider`, `ProtectedRoute`
2. Navbar sign-out, 
3. restructured routes.

**Tests:** 
1. auth tests + existing project tests updated for ownership.

  ### Key decisions & why

  **Auth mechanism**
  - **JWT**
  - **Token in an `httpOnly` cookie** (not localStorage) — JS can't read it, so it's
    not exposed to XSS
  - **Cookie auth over an `Authorization` header** — since the cookie is sent
    automatically, the frontend needs no token-handling code.
  - **24h expiry** — bounds the damage window if a token leaks; logout just clears
    the cookie (no server-side revocation needed at this scale).

  **Security**
  - **`SameSite=Lax`** — cookie isn't sent on cross-site POSTs (basic CSRF
    defense) but still works on normal navigation into the app. OK for now.
  - **`Secure` flag is env-driven (`COOKIE_SECURE`)** — HTTPS-only in prod, off in
    local dev so it works over http.
  - **bcrypt for password hashing**
  - **Password length validated in bytes (8–72)**

  **Data model**
  - **`uuid7` for user IDs** — time-ordered, so better index locality than random
    `uuid4`; IDs are app-generated (known before insert).
  - **`uuid4` kept for project IDs**: left unchanged for now to keep the change scoped but should be updated to uuid7 as well. better in long run.
  - **`owner_id` FK on projects + owner-scoped queries** — every project belongs
    to a user.
  - **Non-owned project returns 404, not 403** — doesn't leak that another user's
    project exists.
  - **`get_current_user` dependency** — auth logic lives in one place, reused
    across endpoints.

  **Frontend**
  - **`ProtectedRoute`** — guests are redirected to `/signin` with a `next` param
    so they land back where they were headed; mirrors the backend gate.
  - **`withCredentials: true` on the axios client** for the browser to
    send/receive the httpOnly cookie across the dev origins.
  - **New files written in TypeScript**: incremental TS adoption as mentioned in my GSoC proposal.

  **Testing**
  - Integration tests (FastAPI `TestClient` + SQLite)
 - existing project tests were updated to thread `owner_id`

hanzalahwaheed:
The failing CI tests will be resolved by merging this PR: https://github.com/c2siorg/dataloom/pull/297

ivantha (changes requested):
Thanks for the thorough auth implementation, Hanzalah. The overall shape (bcrypt+JWT, httpOnly Lax cookies, owner-scoped 404 on cross-user access, comprehensive `test_auth.py`) is solid, and `jwt_secret: str` having no default in `app/config.py` is exactly right (fail-fast on startup if unset).

Before this can merge, a few things need attention. Items 1–3 are blockers; the rest are smaller fixes.

### 1. Scope linkage to GSoC issue #76
The PR currently references issue #177. Authentication wasn't named in the original phasing of #76, but we'll be amending #76 to add it as an explicit phase. Once that's done, please update the PR description to reference #76 so the work is traceable against the GSoC scope of record.

### 2. Backend Tests CI is failing
Needs to be green before re-review. (Note: I just merged #297 which fixes two unrelated drift failures on `main` — once you rebase, that part should clear up. Anything remaining is auth-specific.)

### 3. Hardcoded test user baked into a versioned migration
\`dataloom-backend/alembic/versions/c3d4e5f6a7b8_add_users_and_project_owner.py\` lines 21–23 hardcode \`test@test.com / testpassword\` (with a fixed UUID and bcrypt hash) into the migration's \`upgrade()\`. Any environment that runs migrations to head — including production — will get this user. That's a real security hazard.

DataLoom is moving seed logic *out* of Alembic anyway — see Shivam's open #306 (\"move database seeding out of alembic migrations into standalone script\") for the project-seed equivalent. Please follow the same pattern for the user seed: keep the migration purely structural (create \`users\` table, add \`owner_id\` to \`projects\`, backfill via env-provided owner or NULL→error gate, FK, index), and move \`test@test.com\` into the dev-only seeding script that #306 introduces.

The backfill itself needs a story too: \`UPDATE projects SET owner_id = :owner_id\` currently points all pre-existing projects at the hardcoded test user. If we ship this to a real deployment, every legacy project gets reassigned to a fake user. The dev seed should own this, not the migration.

### 4. No rate limiting on `/auth/signup` and `/auth/signin`
Both endpoints are open to credential stuffing and enumeration. Please add \`slowapi\` (FastAPI-friendly) with per-IP limits on these two routes specifically (something like 5/minute on signin, 3/hour on signup). If you'd rather defer the implementation, open a follow-up issue and label it as a release blocker.

### 5. `.env.example` should mark `JWT_SECRET` as required
The config-side is correct (no default, will crash on startup). But \`.env.example\` should make it explicit — e.g., \`JWT_SECRET=  # REQUIRED — generate with: openssl rand -hex 32\`. Currently a reader may not realize the absence of a default is intentional.

### 6. Migration reversibility
The structure of \`downgrade()\` is correct (FK → index → column → table-index → table, in the right order). Please smoke-test it once locally — \`alembic downgrade -1\` followed by \`alembic upgrade head\` — and confirm it round-trips cleanly. Not a code change, just verification.

### 7. CORS with credentials
The frontend sets \`withCredentials=true\` for the auth cookie. Please confirm \`CORSMiddleware\` in \`app/main.py\` is initialized with \`allow_credentials=True\` and that \`cors_origins\` does not include \`\"*\"\` (it currently lists \`http://localhost:3200\` — that's fine for dev, but the browser silently rejects credentialed CORS requests if origin is \`*\`).

### 8. (Nice-to-have) Cross-user access test
The \"return 404 on cross-user project access (hide existence)\" behavior is the right call. \`test_auth.py\` looks like it covers this — just double-check there's an explicit assertion for the case where User B tries to GET a project owned by User A and gets 404, not 403.

---

Happy to re-review as soon as items 1–4 are addressed and CI is green. Thanks again for the careful work on the test coverage and the SameSite/httpOnly cookie setup.

ivantha:
Hi @hanzalahwaheed, I rebased onto main and addressed the migration security blocker plus the two metadata nits from the review:

1. **Rebased onto current main** and squashed the 6 commits into a single `feat:` commit. Conflicts in `schemas.py` (Pydantic v2 from #282), `projects.py` (checkpoint refactor from #275 + the redundant `get_project_or_404` call removed) and `api/index.js` (kept `undoLastTransformation` from #298 alongside the new auth exports).
2. **Migration security fix**: replaced the hardcoded `test@test.com` / `testpassword` user with a non-loginable `system@dataloom.local` user whose `password_hash` is a bcrypt hash of a discarded random secret. Existing projects backfill to this system user, satisfying the NOT NULL FK with no exploitable creds in the schema.
3. **Dev seed user** moved into `scripts/seed.py` as `seed_dev_user()` (creates `test@test.com` / `testpassword` for local dev). Also updated `scripts/seed.py` to pass `owner_id` to `create_project`, which was otherwise broken by the new required arg.
4. **PR description** now links #76 (was #177).
5. **.env.example**: `JWT_SECRET` is now marked REQUIRED with no default, so pydantic-settings fails fast if unset.
6. **Stale tests** that the rebase exposed (`test_undo.py` x3 and one regression test in `test_save_revert.py`) updated to set `owner_id=test_user.id` on `models.Project` and `create_project`.

**Deferred to follow-ups** (per scope agreement): rate limiting on `/auth/signup` and `/auth/signin`, the explicit 404-vs-403 cross-user assertion, and the manual `alembic downgrade -1; upgrade head` round-trip check.

Force-pushed with `--force-with-lease`. Will merge once CI is green.

PR #279, opened 2026-05-01
feat: add dark mode toggle to navbar

Pranavvv3004:
- Add ThemeContext with localStorage persistence
- Add dark mode toggle button to Navbar
- Wrap app in ThemeProvider in main.jsx
- Enable darkMode: 'class' in tailwind.config.js
- Fix handleCloseModal initialization order in Homescreen

Closes #251

## Description

Summary of the changes. Reference any related issues.

Fixes #251 

## Type of Change

- [ ] Bug fix
- [ x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Describe the tests you ran to verify your changes.

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist

- [ ] My code follows the project's style guidelines
- [x] I have performed a self-review
- [x] I have added/updated documentation as needed
- [ ] My changes generate no new warnings
- [ ] Tests pass locally

PR #300, opened 2026-05-19
Feature: add column reorder via drag and drop

### LINKED ISSUE #80: Column Reorder via Drag-and-Drop
**Difficulty:** Medium
**Area:** frontend

## Description

DataLoom's table view currently displays columns in the order they appear in the CSV file. Users have no way to rearrange columns to better suit their workflow — for example, moving the target column next to related feature columns, or grouping similar columns together. Being able to visually reorder columns makes data exploration significantly easier.

This issue adds drag-and-drop reordering of table columns by dragging column headers. The reorder is purely visual — it changes the display order but does not modify the underlying data. The column order is persisted in the project context so it is maintained while the user navigates within the application.

## Acceptance Criteria

- [ ] Column headers (excluding "S.No.") are draggable
- [ ] Dragging a column header and dropping it on another column reorders the display
- [ ] Both the header row and all data rows reflect the new column order
- [ ] Visual feedback is provided during drag: the dragged column is semi-transparent and the drop target is highlighted
- [ ] Column headers show a grab cursor to indicate they are draggable
- [ ] The column order is persisted during the session (maintained while navigating within the app)
- [ ] The "S.No." column is always first and is not draggable
- [ ] The reorder does not modify the underlying data or affect backend operations
- [ ] Existing right-click context menu on headers still works after adding drag support
- [ ] Loading a new project resets the column order to the default CSV order

## Resources

- `dataloom-frontend/src/Components/Table.jsx` — main data table with `<thead>` and column rendering
- `dataloom-frontend/src/context/ProjectContext.jsx` — project context with `columns` and `rows` state
- [HTML Drag and Drop API (MDN)](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API)
- [Drag events (MDN)](https://developer.mozilla.org/en-US/docs/Web/API/DragEvent)

SHIVAM-KUMAR-59:
## Description
Adds column reorder via drag-and-drop on table column headers. Users can drag any column header (excluding S.No.) and drop it onto another column to reorder the display. The reorder is purely visual — it does not modify the underlying data or affect backend operations.

Changes made:
- Added `columnOrders` state to `ProjectContext.jsx` to persist column order per project
- Added `columnOrder` and `setColumnOrder` to project context value
- Updated `updateData` to accept `options` object with `resetColumnOrder` flag
- Added `useEffect` to initialize column order when columns load
- Updated `Table.jsx` to render columns and rows in the current display order
- Added drag state (`draggedColIndex`, `hoveredTargetIndex`) to `Table.jsx`
- Added drag-and-drop event handlers to column headers
- Visual feedback during drag: dragged column is semi-transparent, drop target highlighted with blue ring
- Column headers show grab cursor to indicate they are draggable
- Updated `handleEditCell` to translate display index → backend index using `columnOrder`
- Updated `handleRenameColumn` to use backend index from `columnOrder`
- Updated `handleDeleteColumn` to use backend index and `del_col_params`
- Updated `handleAddColumn` to use backend index from `columnOrder`
- Updated `handleAddRow` and `handleDeleteRow` to normalize rows correctly
- Loading a new project resets column order to default CSV order

Closes #80

## Type of Change
- [x] New feature

## How Has This Been Tested?
- [x] Existing tests pass
- [x] Manual testing

1. Drag column header and drop on another → columns reorder correctly
2. Edit cell after reordering → correct cell edited
3. Rename column after reordering → correct column renamed
4. Delete column after reordering → correct column deleted
5. Add column after reordering → column added at correct position
6. Right-click context menu after reordering → still works correctly
7. Load a new project → column order resets to default
8. S.No. column is not draggable and always stays first
9. All existing frontend tests pass

## Screen Recordings

https://github.com/user-attachments/assets/d9af0520-eff2-487d-aa7f-fdd60435abef

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/Table.jsx
+++ b/dataloom-frontend/src/Components/Table.jsx
@@ -37,6 +37,8 @@ const Table = ({ projectId, data: externalData }) => {
     rows: ctxRows,
     dtypes,
     updateData,
+    columnOrder,
+    setColumnOrder,
     totalRows,
     totalPages,
     page,
@@ -53,12 +55,24 @@ const Table = ({ projectId, data: externalData }) => {
   const [inputConfig, setInputConfig] = useState(null);
   const [toast, setToast] = useState(null);
 
+  const [draggedColIndex, setDraggedColIndex] = useState(null);
+  const [hoveredTargetIndex, setHoveredTargetIndex] = useState(null);
+
   useEffect(() => {
     if (ctxColumns.length > 0 && ctxRows.length > 0) {
-      setColumns(["S.No.", ...ctxColumns]);
-      setData(ctxRows.map((row, index) => [(page - 1) * pageSize + index + 1, ...row]));
+      const orderedColumns =
+        columnOrder.length === ctxColumns.length
+          ? columnOrder.map((i) => ctxColumns[i])
+          : ctxColumns;
+      setColumns(["S.No.", ...orderedColumns]);
+      setData(
+        ctxRows.map((row, index) => [
+          (page - 1) * pageSize + index + 1,
+          ...(columnOrder.length === ctxColumns.length ? columnOrder.map((i) => row[i]) : row),
+        ]),
+      );
     }
-  }, [ctxColumns, ctxRows, page, pageSize]);
+  }, [ctxColumns, ctxRows, page, pageSize, columnOrder]);
 
   useEffect(() => {
     if (externalData) {
@@ -82,6 +96,10 @@ const Table = ({ projectId, data: externalData }) => {
         row_params: { index },
       });
       updateTableData(response);
+      const normalizedRows = response.rows.map((row) =>
+        Array.isArray(row) ? row : Object.values(row),
+      );
+      updateData(response.columns, normalizedRows, { resetColumnOrder: false });
       refreshProject(projectId, 1, pageSize);
     } catch {
       setToast({
@@ -102,11 +120,23 @@ const Table = ({ projectId, data: externalData }) => {
         }
 
         try {
+          let backendIndex; // Normalized index
+          if (index === 0) {
+            backendIndex = 0;
+          } else {
+            const displayDataIndex = index - 1;
+            const baseIndex = columnOrder[displayDataIndex];
+            backendIndex = baseIndex + 1;
+          }
           const response = await transformProject(projectId, {
             operation_type: ADD_COLUMN,
-            add_col_params: { index, name: newColumnName },
+            add_col_params: { index: backendIndex, name: newColumnName },
           });
           updateTableData(response);
+          const normalizedRows = response.rows.map((row) =>
+            Array.isArray(row) ? row : Object.values(row),
+          );
+          updateData(response.columns, normalizedRows, { resetColumnOrder: true });
           refreshProject(projectId, 1, pageSize);
         } catch {
           setToast({
@@ -127,6 +157,10 @@ const Table = ({ projectId, data: externalData }) => {
         row_params: { index },
       });
       updateTableData(response);
+      const normalizedRows = response.rows.map((row) =>
+        Array.isArray(row) ? row : Object.values(row),
+      );
+      updateData(response.columns, normalizedRows, { resetColumnOrder: false });
       refreshProject(projectId, 1, pageSize);
     } catch {
       setToast({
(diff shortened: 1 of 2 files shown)


SHIVAM-KUMAR-59:
Note: 1 failing CI check (backend tests) is a pre-existing issue being fixed in PR #297. Once that is merged, all CI checks will pass on this PR.

ivantha:
Thanks for this PR! It now has merge conflicts after recent merges to main (Table.jsx and ProjectContext.jsx). Could you rebase on the latest main and resolve? Once green I'll merge.

SHIVAM-KUMAR-59:
Thanks for the heads up @ivantha ! I’ve rebased the branch on the latest main, resolved the conflicts and pushed the updated changes. All checks are passing now.

ivantha (changes requested):
Thanks for the careful implementation, @SHIVAM-KUMAR-59. The display-index to backend-index translation is handled correctly across all the transform call sites this PR touches, the drag-and-drop state machine is clean, and the visual feedback (opacity-50 on dragged, ring-2 on drop target, grab cursor) is solid. The per-project column order in `ProjectContext` is the right architectural choice.

Before this can land, one blocker plus a few smaller items.

### 1. The `updateData` signature change is a breaking regression for every caller outside this PR

`updateData` is now `(newColumns, newRows, options = {})` where `options.dtypes` is the new dtypes location. The PR updates the new call sites it adds inside `Table.jsx`, but every pre-existing caller still passes the OLD shape and is now silently broken.

`grep -rn "updateData(" dataloom-frontend/src/` shows:
- `Table.jsx:75` (inside `updateTableData`): `updateData(columns, rows, newDtypes)`. This is the helper that every transform handler in `Table.jsx` ultimately calls (add row, delete row, edit cell, add column, delete column, rename column). After this PR, `newDtypes` is treated as `options`, `options.dtypes` is undefined, and **dtypes are silently not updated after any transform**.
- `MenuNavbar.jsx:115`: `updateData([], [], {})`. Intended to clear state; the new code keeps the old dtypes.
- `MenuNavbar.jsx:117`: `updateData(data.columns, data.rows, data.dtypes)`. After revert/restore via MenuNavbar, **dtypes are silently not updated**.
- `forms/FilterForm.jsx:41`, `forms/SortForm.jsx:34`, `forms/GroupByForm.jsx:43`, `forms/CastDataTypeForm.jsx:33`: same regression after Filter / Sort / GroupBy / CastDataType. The CastDataType case is particularly bad because its whole point is to change the dtype, and the UI will not reflect it.

Fix every caller to use the new shape:
```js
updateData(response.columns, response.rows, { dtypes: response.dtypes });
```
and inside `updateTableData`:
```js
updateData(columns, rows, { dtypes: newDtypes });
```

The `{}` cases in `MenuNavbar.jsx` may also need `{ dtypes: {} }` if the intent was to clear, or just leave them as `{}` if the intent was "do not touch dtypes."

### 2. `hoveredTargetIndex` lingers when drag moves between columns

`onDragOver` sets `hoveredTargetIndex`, but it is only cleared on `onDrop` and `onDragEnd`. If the user drags off the table without dropping, the ring persists on whichever column was last hovered until `onDragEnd` fires. Add `onDragLeave={() => setHoveredTargetIndex(null)}` on the same button to clear it cleanly.

Not a blocker. Polish.

### 3. No tests

The reorder logic is complex enough to deserve coverage. At minimum:
- A Vitest test for `ProjectContext` covering `setColumnOrder` updating the order for the current project only, and switching `projectId` showing the other project's order or initializing fresh.
- A Vitest test for `Table.jsx` covering rendering with `columnOrder = [2, 0, 1]` and verifying display order, plus the backend-index translation in `handleAddColumn` / `handleDeleteColumn` / `handleEditCell` for a reordered table.

### 4. `columnOrder.length === ctxColumns.length` guard repeated

The defensive check appears in three places (lines 60-64 of the diff and inside the `onDrop` handler). Worth extracting to a small helper:
```js
const safeOrder =
  columnOrder.length === ctxColumns.length ? columnOrder : ctxColumns.map((_, i) => i);
```
Not a blocker. DRY.

### 5. (Nice-to-have) Persist column order across reloads

The order is in-memory only. A page refresh wipes the reordering. Persisting via `localStorage` keyed on `projectId` would match the pattern Pranav is introducing in #279 for the theme toggle. Not required for this PR; worth a follow-up issue.

### 6. Touch devices

HTML5 `draggable` does not work on touch devices. If you want this to work on iPad or mobile, a future PR could swap in a touch-aware library (e.g. `dnd-kit`) or polyfill. Worth a one-line note in the PR description that current scope is desktop drag-and-drop.

---

Happy to re-review as soon as item 1 is addressed. The DnD UX itself looks great in your screen recording; we just need the dtype path to keep working through every transform.

SHIVAM-KUMAR-59:
Thanks for the review @ivantha , and sorry for missing the impact of the `updateData` signature change. It does break the existing forms and transform flows.

I’m updating all the affected call sites to use the new object-based shape. I’ll also add the `onDragLeave` fix for clearing the hover state, refactor the repeated guard into a small helper, and update the PR description to clarify the desktop-only drag-and-drop scope.

Alongside that, I’ll start adding the suggested tests and incorporate the remaining improvements.

I’ll push an updated PR soon and ping you once it’s ready for another review.

SHIVAM-KUMAR-59:
Thanks for merging @ivantha ! I noticed you removed `onDragLeave` and the `dtypes` updates in row/column handlers, happy to keep those out if that was intentional. I'm planning a follow-up PR for the missing tests and `safeOrder` extraction you mentioned. Should I also include `onDragLeave` and `dtypes` updates, or leave them as-is?

PR #311, opened 2026-05-23
Feature: persist column order to localStorage keyed by projectId

### LINKED ISSUE #80: Column Reorder via Drag-and-Drop

SHIVAM-KUMAR-59:
## Description
Persists column reorder state to localStorage so the display order survives page refreshes. The order is keyed by `projectId` so each project maintains its own independent column order across sessions.

Changes made:
- Initialised `columnOrders` state from localStorage on mount with JSON parse error handling
- Added `useEffect` in `ProjectContext.jsx` to sync `columnOrders` to  localStorage on every change
- Added validation of persisted order against current column schema —  discards stored order if length mismatches or any index is out of bounds,  preventing stale permutations from being applied silently after schema changes
- Added `deleteProjectOrder` to `ProjectContext` and called it in  `Homescreen.jsx` on project delete to prune stale localStorage entries
- Added 3 Vitest tests covering localStorage hydration, stale order  invalidation, and corrupted JSON handling

Closes #80 follow-up (persistence across reloads noted in #300 review)

## Type of Change
- [x] New feature

## How Has This Been Tested?
- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

1. Reorder columns → refresh page → column order persists
2. Reorder columns on project A → open project B → independent order
3. Load a new project → column order initialises fresh
4. Rename a column (schema change) → stale order discarded, fresh order applied
5. Delete a project → localStorage entry pruned
6. All existing tests pass
7. New localStorage hydration tests pass

## Screen Recordings

https://github.com/user-attachments/assets/795161f8-f810-4fd0-a6cb-74abc1c9d7d9

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/Homescreen.jsx
+++ b/dataloom-frontend/src/Components/Homescreen.jsx
@@ -5,6 +5,7 @@ import { useToast } from "../context/ToastContext";
 import ConfirmDialog from "./common/ConfirmDialog";
 import { UploadCloud, FileText, X, Pencil } from "lucide-react";
 import { ACCEPTED_EXTENSIONS, formatFileSize, validateFile } from "../utils/fileUtils";
+import { useProjectContext } from "../context/ProjectContext";
 
 const ProjectCard = ({ project, onClick, onDelete }) => {
   const modified = new Date(project.last_modified).toLocaleDateString(undefined, {
@@ -109,6 +110,7 @@ const HomeScreen = () => {
   const [deleteConfirm, setDeleteConfirm] = useState({ open: false, projectId: null });
   const navigate = useNavigate();
   const { showToast } = useToast();
+  const { deleteProjectOrder } = useProjectContext();
 
   const isFormValid =
     projectName.trim().length > 0 && projectDescription.trim().length > 0 && fileUpload !== null;
@@ -241,6 +243,7 @@ const HomeScreen = () => {
   const handleDeleteConfirm = async () => {
     try {
       await deleteProject(deleteConfirm.projectId);
+      deleteProjectOrder(deleteConfirm.projectId);
       showToast("Project deleted successfully", "success");
       fetchRecentProjects();
     } catch (error) {
--- a/dataloom-frontend/src/context/ProjectContext.jsx
+++ b/dataloom-frontend/src/context/ProjectContext.jsx
@@ -1,4 +1,4 @@
-import { createContext, useContext, useState, useCallback, useEffect } from "react";
+import { createContext, useContext, useState, useCallback, useEffect, useMemo } from "react";
 import { getProjectDetails } from "../api";
 
 const ProjectContext = createContext(null);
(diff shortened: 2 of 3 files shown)


ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `dataloom-frontend/src/context/ProjectContext.jsx:52` The column-order auto-initialization effect only repopulates when the stored order is missing or empty (its guard is `!currentOrder || currentOrder.length === 0`). After hydrating columnOrders from localStorage with a stale-length array (e.g. a saved order of length 3 for a project that now has 2 columns), the new derived `columnOrder` memo correctly returns [] so the table falls back to natural order, but the stale entry is never repaired in state on the refreshProject load path. refreshProject calls setColumns(data.columns) without ever calling updateData, so the reset logic in updateData (which compares existingOrder.length !== newColumns.length) never runs. The stale array therefore lingers in columnOrders and is re-persisted by the new sync effect on the next write. Self-heals only when updateData is invoked (e.g. after a transform). Display is safe; the persisted state is briefly inconsistent.
- **[minor/quality]** `dataloom-frontend/src/context/ProjectContext.jsx:55` The new effect serializes the entire columnOrders map (all projects' orders) to localStorage on every change, and entries are pruned only via deleteProjectOrder, which is wired solely into the Homescreen delete-confirm flow (the only in-app deletion path). Projects removed by any other means (backend/db directly, or future deletion paths) leave orphaned entries that accumulate without bound in localStorage. Low impact given current single delete path, but the persisted map has no size cap or reconciliation against the known project list.

Solid, self-contained persistence enhancement: columnOrders is hydrated from localStorage with JSON-parse error handling, synced on change, validated against the live column schema (length + index bounds) before use, and pruned on project delete, with three new tests covering hydration, stale invalidation, and corrupted JSON. Two minor issues: (1) on the refreshProject load path the stored order is never repaired when it has a stale length, so a mismatched entry lingers in state and is re-persisted until an updateData call self-heals it (display is safe because the new memo returns [] and the Table falls back to natural order); (2) the localStorage map is pruned only through the Homescreen delete flow, so projects removed by other means leave orphaned entries that grow unbounded. Neither blocks merge; consider repairing/normalizing stored order against the current column count on the refresh path and reconciling persisted entries against known projects.

PR #312, opened 2026-05-23
Fix: update undo endpoint to use `Depends(get_project_or_404)` for auth

SHIVAM-KUMAR-59:
## Description
Fixes the undo endpoint which was broken after the auth feature was merged in #299.

The endpoint was calling `get_project_or_404(project_id, db)` directly — the old pattern before authentication was introduced. Updated to use `Depends(get_project_or_404)` consistent with every other endpoint in the codebase.

Fixes: undo endpoint returning "Network Error" after auth merge.

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- [x] Manual testing

1. Apply Sort → click Undo → transformation reversed correctly
2. Apply Filter → click Undo → transformation reversed correctly
3. Apply GroupBy → click Undo → transformation reversed correctly
4. Click Undo with no transformations → "No transformations to undo" toast

## Screens Recordings
**Before**

https://github.com/user-attachments/assets/482a3176-5500-4fd2-8a30-6cfcc81071a2

**After**

https://github.com/user-attachments/assets/f7e0e48e-d97b-4f5c-aa21-7d79f7447d65

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -263,6 +263,7 @@ async def delete_project_endpoint(
 @router.post("/{project_id}/undo", response_model=schemas.ProjectResponse)
 async def undo_last_transformation(
     project_id: uuid.UUID,
+    project: models.Project = Depends(get_project_or_404),
     db: Session = Depends(database.get_db),
 ):
     """Undo the most recent transformation.
@@ -270,8 +271,6 @@ async def undo_last_transformation(
     Removes the last change log entry and rebuilds the working copy
     by replaying all remaining logs onto the original file.
     """
-    project = get_project_or_404(project_id, db)
-
     last_log = get_last_change_log(db, project_id)
     if not last_log:
         raise HTTPException(status_code=404, detail="No transformations to undo")


ivantha:
### Automated review summary

**Findings**
- **[minor/test]** `dataloom-backend/tests/test_undo.py:11` The fix corrects an endpoint dependency-injection regression, but no endpoint-level regression test is added. tests/test_undo.py only exercises service-layer helpers (get_last_change_log, delete_change_log, log_transformation) in isolation and never calls POST /{project_id}/undo through a TestClient, which is exactly why the #299 break passed CI. Authenticated `client` and `anon_client` fixtures already exist in tests/conftest.py and are used by other HTTP tests (e.g. test_transform_http_semantics.py, test_save_revert.py), so adding a test that asserts 200 for the owner and 401/404 for anon/non-owner would be cheap and would prevent this class of regression from recurring silently.

Solid, correct bug fix. On main the undo endpoint called get_project_or_404(project_id, db) positionally, but PR #299 changed that dependency's signature to (project_id, current_user=Depends(...), db=Depends(...)); the old call bound the Session to current_user and left db as an unresolved Depends sentinel, so db.query(...) failed at runtime (the reported "Network Error") and the per-user ownership check was bypassed. Switching to Depends(get_project_or_404) restores both DI and the auth/ownership guard and matches every sibling endpoint in projects.py. I verified the fix on the PR branch: ruff is clean and the undo + HTTP-level endpoint/auth suites pass (33 tests). The only gap is the lack of an endpoint-level regression test for POST /{project_id}/undo, which is what let the original break slip through CI; recommend adding one before or right after merge using the existing client/anon_client fixtures.

PR #316, opened 2026-05-26
fix: increase table max height to prevent last row being hidden behind pagination bar

### LINKED ISSUE #314: [Bug:UI] Inconsistent Scroll layout and Table height
**Describe the bug**
Due to the inconsistent scrolling layout of the app, the top bar is hidden when scrolling to the bottom which is _disorienting to the user,_ and the table is partially hidden at the bottom due to the pagination bar. 

**Expected behavior**
- Scroll should be present only in the table.
- the top bar and the pagination should always be visible for the table view.

**Screenshots**

Image

> The topbar is gets moved as the entire page is scrollable and at the bottom, the table end is not visible because the pagination is covering it.

SHIVAM-KUMAR-59:
## Description
Fixes the last row of the data table being hidden behind the fixed pagination bar at the bottom of the screen.

The table container's `maxHeight` was calculated without accounting for the height of the fixed pagination footer, causing the last row to be obscured and unreachable.

Fixes #314

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- [x] Manual testing

1. Open a project with multiple rows → scroll to bottom → last row fully visible
2. Different page sizes (10, 25, 50, 100) → last row always visible

## Screenshots 
**Before**
Before

**After**
After

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/Table.jsx
+++ b/dataloom-frontend/src/Components/Table.jsx
@@ -298,7 +298,7 @@ const Table = ({ projectId, data: externalData }) => {
     <div className="px-8 pt-3">
       <div
         className="overflow-x-scroll overflow-y-auto border border-gray-200 rounded-lg shadow-sm"
-        style={{ maxHeight: "calc(100vh - 140px)" }}
+        style={{ maxHeight: "calc(100vh - 225px)" }}
       >
         <table data-testid="data-table" className="min-w-full bg-white">
           <thead className="sticky top-0 bg-gray-50">


hanzalahwaheed:
the bug is wrt to overall scroll layout because ideally the table should be taking all available space, not a fixed height. this is a good patch but I'm working on a proper fix for this.

thanks for the PR :)

SHIVAM-KUMAR-59:
Thanks for the context @hanzalahwaheed ! Happy to close this if you're working on a proper fix. Let me know if there's anything I can help with on your approach!

PR #317, opened 2026-05-26
fix(frontend): prevent invalid file uploads from triggering API calls

sanchit955:
## Description

Prevents invalid CSV uploads from triggering unnecessary API requests by centralizing file validation logic.

### Changes

* Added a reusable `validateFile` helper
* Prevented invalid files from being stored in state
* Prevented invalid uploads from reaching `uploadProject`
* Preserved existing CSV-only behavior and upload UX

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update

## How Has This Been Tested?

* [x] Existing tests pass
* [ ] New tests added
* [x] Manual testing

### Manual Testing

* Verified non-CSV files are rejected before API calls
* Verified oversized files are rejected before API calls
* Verified empty files are rejected
* Verified valid CSV uploads continue working correctly

## Checklist

* [x] My code follows the project's style guidelines
* [x] I have performed a self-review
* [ ] I have added/updated documentation as needed
* [x] My changes generate no new warnings
* [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/Homescreen.jsx
+++ b/dataloom-frontend/src/Components/Homescreen.jsx
@@ -63,6 +63,28 @@ const NewProjectCard = ({ onClick }) => (
 
 const MAX_FILE_SIZE_BYTES = 10 * 1024 * 1024; // 10 MB
 
+const validateFile = (file) => {
+  if (!file) {
+    return { valid: false, error: "Please select a file to upload" };
+  }
+
+  const isCSV = file.type === "text/csv" || file.name.toLowerCase().endsWith(".csv");
+  if (!isCSV) {
+    return { valid: false, error: "Please upload a CSV file." };
+  }
+
+  if (file.size === 0) {
+    return { valid: false, error: "File is empty (0 bytes)." };
+  }
+
+  if (file.size > MAX_FILE_SIZE_BYTES) {
+    const sizeMB = (file.size / (1024 * 1024)).toFixed(1);
+    return { valid: false, error: `File too large (${sizeMB} MB). Maximum allowed size is 10 MB.` };
+  }
+
+  return { valid: true };
+};
+
 const EmptyState = ({ onClick }) => (
   <div className="flex flex-col items-center justify-center py-16 px-6 rounded-xl border-2 border-dashed border-gray-200 bg-gray-50 text-center">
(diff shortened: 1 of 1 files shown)


SHIVAM-KUMAR-59:
Hey @sanchit955, this check:

```javascript id="i5c0l1"
if (file.size === 0) {
  return { valid: false, error: "File is empty (0 bytes)." };
}
```

alone may not be sufficient because CSV files containing only whitespace, newlines, or empty rows can still have a non-zero byte size and would incorrectly pass validation.

It may be worth additionally validating the actual file contents asynchronously (for example using `await file.text()`) and rejecting files whose trimmed content is empty. This would prevent whitespace-only or effectively empty CSV files from reaching the upload API.

For example:

```javascript id="yo0m1o"
const content = await file.text();

if (content.trim().length === 0) {
  return {
    valid: false,
    error: "File is empty or contains only empty rows.",
  };
}
```

This should help catch cases that the size check alone cannot detect.

ivantha:
### Automated review summary

**Triage:** refactor  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[minor/test]** `dataloom-frontend/src/Components/Homescreen.jsx:66` The PR introduces a new validateFile helper (the core logic of the change) but adds no unit tests for it; the PR checklist leaves 'New tests added' unchecked while claiming the change is tested. Homescreen.jsx has no existing test file, so the new branching (CSV check, 0-byte check, size check, null check) is entirely uncovered. A small Vitest unit test on the pure validateFile function would lock in the four branches cheaply.

PR #317 is a clean DRY refactor: it extracts two previously-duplicated client-side upload-validation blocks (in handleSubmitModal and handleFileUpload) into a single validateFile helper, and adds one new guard rejecting 0-byte files before the API call. All call sites route through the helper, behavior and messages are preserved, and the backend already independently enforces extension/size limits (dataloom-backend/app/services/file_service.py, security.py), so this is a UX/efficiency improvement rather than a security boundary. It does not overlap the reserved GSoC scope in issue #76. Worth requesting a small unit test for validateFile (the only substantive gap); consider relabeling from 'bug fix' to refactor, then it's mergeable.

sanchit955:
Thanks for the review and merge! I appreciate the feedback and suggestions.

PR #323, opened 2026-05-31
fix(backend): delete checkpointed projects reliably

### LINKED ISSUE #322: [Bug] Fix project deletion failures for saved/checkpointed projects
**Describe the bug**
Some projects fail to delete from the app, and the backend currently does not surface enough useful logging to diagnose the failure.

**To Reproduce**

1. Create a project
2. Add a checkpoint
3. Try to delete it. It will fail.

**Expected behavior**

- Saved/checkpointed projects can be deleted successfully.
- Failed database deletes emit useful backend logs with project context.
- File cleanup does not happen before a successful database delete.
- Regression test covers deleting a project with logs assigned to a checkpoint.

hanzalahwaheed:
## Description
- Projects with checkpoints can be deleted successfully now.
- Added a new flow for project deletion where we first delete the logs, checkpoints and projects. Then we delete the project file.
- Also added some loggers.

Fixes https://github.com/c2siorg/dataloom/issues/322

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [X] New tests added
- [X] Manual testing

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -255,8 +255,28 @@ async def delete_project_endpoint(
     project: models.Project = Depends(get_project_or_404),
 ):
     """Delete a project and its associated files."""
-    delete_project_files(project.file_path)
-    delete_project(db, project)
+    project_id = project.project_id
+    project_name = project.name
+    file_path = project.file_path
+
+    logger.info("Delete project request: id=%s, name=%s, file_path=%s", project_id, project_name, file_path)
+
+    try:
+        delete_project(db, project)
+    except SQLAlchemyError as e:
+        logger.exception("Delete project database failure: id=%s, name=%s", project_id, project_name)
+        raise HTTPException(status_code=500, detail="Failed to delete project records.") from e
+
+    try:
+        delete_project_files(file_path)
+    except OSError:
+        logger.exception(
+            "Project database record deleted, but file cleanup failed: id=%s, name=%s, file_path=%s",
+            project_id,
+            project_name,
+            file_path,
+        )
+
     return {"success": True, "message": "Project deleted"}
 
 
--- a/dataloom-backend/app/services/project_service.py
+++ b/dataloom-backend/app/services/project_service.py
@@ -3,6 +3,7 @@
 import uuid
 from datetime import UTC, datetime
 
+from sqlalchemy.exc import SQLAlchemyError
 from sqlmodel import Session
 
 from app import models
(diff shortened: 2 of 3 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/test]** `dataloom-backend/tests/test_save_revert.py:47` The new regression test exercises SQLite (conftest.py swaps Postgres for in-memory SQLite), which does NOT enforce foreign keys by default. The actual bug in #322 is a Postgres ForeignKeyViolation caused by migration a1b2c3d4e5f6 recreating the checkpoints/user_logs -> projects FKs WITHOUT ondelete=CASCADE (lines 36-37), whereas the original migration 34e620d988fc had ondelete=CASCADE. Under SQLite even the OLD code (db.delete(project) with unloaded child collections) would pass, so this test would not have caught the original bug and does not guard against regressing to ORM-cascade reliance. It verifies row cleanup but not the FK-failure scenario the issue describes.
- **[minor/quality]** `dataloom-backend/app/services/project_service.py:76` The updated docstring (and PR description) justify deleting logs before checkpoints by claiming 'applied logs can reference checkpoints directly.' That FK (user_logs.checkpoint_id -> checkpoints.id) is ON DELETE SET NULL (migration 34e620d988fc line 75 / 216), so deleting checkpoints before logs would NOT fail; Postgres would null the references. The chosen order is harmless but the stated rationale is inaccurate and may mislead future maintainers.
- **[minor/quality]** `dataloom-backend/alembic/versions/a1b2c3d4e5f6_rename_datasets_to_projects.py:36` The fix addresses the symptom at the application layer but leaves the root-cause schema defect intact: migration a1b2c3d4e5f6 (lines 36-37) recreated the child FKs to projects without ondelete=CASCADE, while models.py still declares Relationship cascade='all, delete-orphan' (lines 65, 69). DB and ORM are now inconsistent, and any other or future caller that does db.delete(project) with unloaded child collections will still hit the FK violation. A migration restoring ondelete=CASCADE would be the complete fix; consider it as a follow-up.

PR #323 fixes issue #322 (checkpointed projects failing to delete) by replacing the ORM cascade delete with three ordered bulk deletes (logs, checkpoints, project) inside a try/rollback, deleting DB records before files, and adding contextual logging. The approach is sound and correctly sequences the DB delete before file cleanup, so it is safe to merge in principle and has zero overlap with the reserved GSoC scope in #76. Two things to raise with the author: (1) the regression test runs on SQLite, which does not enforce foreign keys, so it does not actually reproduce the Postgres FK-violation root cause and would not have caught the original bug; (2) the true defect is migration a1b2c3d4e5f6 recreating the child FKs without ondelete=CASCADE, which leaves the DB schema inconsistent with the cascade still declared in models.py, so any other db.delete(project) caller remains broken. Recommend a follow-up migration to restore ondelete=CASCADE and, ideally, a Postgres-backed test for the FK path. Also fix the docstring rationale, since the user_logs.checkpoint_id FK is ON DELETE SET NULL and does not force the log-before-checkpoint order.

PR #324, opened 2026-05-31
refactor: consolidate transformation dispatch into a single registry

### LINKED ISSUE #315: Consolidate transformation dispatch into a single registry
## Problem

The list of supported transformations is duplicated across multiple backend locations:

- the `OperationType` enum
- a separate, identical `ActionTypes` enum
- the `COMPLEX_OPERATIONS` set
- the transformation endpoint dispatch logic
- the replay dispatch logic (`apply_logged_transformation`)

Adding a new transformation requires updating several locations. This duplication has already led to inconsistencies where a transformation could be executed successfully but fail during replay because one dispatch path was not updated.

## Proposed fix

Introduce a central transformation registry that acts as the single source of truth for supported transformations.

Each registry entry defines:

- transformation type
- execution function
- parameter schema/field
- persistence behavior
- validation metadata

Both the transformation endpoint and replay path should resolve transformations through this registry instead of maintaining separate dispatch ladders.

## Acceptance criteria

- Single source of truth for supported transformations
- Duplicate `ActionTypes` enum removed
- Endpoint execution path uses registry dispatch
- Replay path uses registry dispatch
- Adding a new transformation requires a single registry entry plus its schema/UI implementation
- Existing API behavior remains unchanged
- Database schema unchanged
- Existing tests pass

hanzalahwaheed:
## Description

- add TRANSFORMATION_REGISTRY driving both /transform and the replay path
- remove duplicate ActionTypes enum and COMPLEX_OPERATIONS set
- point UserLogsAction.actionType at OperationType

Fixes #315

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [ ] New tests added
- [ ] Manual testing

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/transformations.py
+++ b/dataloom-backend/app/api/endpoints/transformations.py
@@ -20,7 +20,6 @@ logger = get_logger(__name__)
 
 router = APIRouter()
 
-COMPLEX_OPERATIONS = {"dropDuplicate", "advQueryFilter", "pivotTables", "dropNa", "melt", "groupby"}
 SAFE_TRANSFORMATION_ERROR_DETAIL = "Invalid transformation request"
 
 _SENSITIVE_TOKEN_MARKERS = (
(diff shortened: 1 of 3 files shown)


ivantha:
### Automated review summary

**Triage:** refactor  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[minor/bug]** `dataloom-backend/app/services/transformation_service.py:862` stringReplace replay tolerance is silently narrowed, a behavior change versus the PR's stated 'existing API behavior remains unchanged'. The old apply_logged_transformation guarded `if not column or find_value is None or replace_value is None: return df`, so it tolerated a present-but-empty column. The new code only treats a KeyError/TypeError from build_args as replay_tolerant; a logged entry like {"string_replace_params": {"column": "", "find_value": "x", "replace_value": "y"}} now has all keys present, so build_args succeeds and string_replace(df, "", "x", "y") raises TransformationError("Column '' not found") instead of returning df unchanged. Verified empirically against the PR branch. Narrow edge case (the execution path validates via StringReplaceParams so such logs are unlikely), hence minor, but it does change /save replay behavior for malformed historical log entries.
- **[nit/test]** `dataloom-backend/app/services/transformation_service.py:626` No tests were added for the new registry machinery despite it being the entire point of the PR (the PR checklist leaves 'New tests added' unchecked and ticks only 'Existing tests pass'). TRANSFORMATION_REGISTRY, TransformationSpec, resolve_transformation, and _dispatch_transform appear in zero test files. The 238 existing tests pass and exercise the happy paths indirectly, but the lambda-based build_args specs (per-operation key extraction, the dropNa/sample optional-field branches, the _col_params legacy fallback, the import-time completeness guard, and the stringReplace replay_tolerant skip) are non-trivial parametrized logic with no direct coverage, so a future operation added with a malformed build_args lambda would only surface at runtime.

Clean, well-documented internal refactor that does exactly what issue #315 asks: one TRANSFORMATION_REGISTRY (keyed by OperationType) now drives both the /transform execution path and the apply_logged_transformation replay path, the duplicate ActionTypes enum and COMPLEX_OPERATIONS set are removed, and an import-time guard fails loudly if any OperationType lacks a registry entry. I verified it carries no GSoC-scope (#76) overlap, that all 238 backend tests pass identically on main and the PR head, that ruff lint/format are clean, that no dangling references to the removed symbols remain, and that the legacy col_params replay fallback and the patchable resolve_transformation seam still work. Two non-blocking items: the stringReplace replay path silently narrows its tolerance for empty-but-present params (a minor behavior change, contradicting the PR's 'behavior unchanged' claim), and no tests were added for the new registry logic. Approve after the author either restores the old stringReplace emptiness guard or confirms the edge case is acceptable, ideally adding a small registry/replay test.

ivantha:
@hanzalahwaheed if you can fix the e2e test failure, we can get this merged.

hanzalahwaheed:
@ivantha rebased and fixed

PR #302, opened 2026-05-19
Fix: use correct column parameter name for Add Column and Delete Column operation

SHIVAM-KUMAR-59:
## Description
Fixes Add Column and Delete Column operations which were failing with a 400 error due to incorrect parameter names in the frontend request body.

- Add Column: frontend was sending `col_params` but backend expects `add_col_params`
- Delete Column: frontend was sending `col_params` but backend expects `del_col_params`

Fixes #301

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- [x] Manual testing

Tested the following scenarios manually:
1. Right-click column header → Add Column → enter name → column added successfully
2. Right-click column header → Delete Column → column deleted successfully
3. Existing operations (Add Row, Delete Row, Edit Cell) still work correctly

## Screen Recordings
**Before**

https://github.com/user-attachments/assets/62c7801f-da1f-4c26-91e4-582e6d3a2393

https://github.com/user-attachments/assets/90b677da-0e17-45b2-bfbb-4332b40c2952

**After**

https://github.com/user-attachments/assets/9136b2e4-ed1c-4938-82a0-ca4adfb2539a

https://github.com/user-attachments/assets/d67766a7-5a15-4a95-b57f-e997782a8b3b

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review
- [x] I have added/updated documentation as needed
- [x] My changes generate no new warnings
- [x] Tests pass locally

PR #329, opened 2026-06-03
fix(backend+frontend): preserve null vs empty-string response semantics

nodesagar:
## Why

Slim follow-up to closed PR #270, as requested by @ivantha. Most of #270's bundle already landed on `main` via other paths (Pydantic v2 in #282, `add_col_params`/`del_col_params` frontend keys in #302, alembic-on-pytest in `a5eb4c5`, and the `AddColumn.name` validator). This PR carries **only** the substance that was still genuinely missing, in the exact shape the maintainer outlined.

## What changed

- **`app/utils/pandas_helpers.py`** — `dataframe_to_response` now preserves null semantics: real empty strings stay `""`, while missing and non-finite values serialize to `null`. Implemented via `df.astype(object).where(pd.notna(...), None)`, which is a vectorized C-level dispatch (not a Python per-cell loop).
- **`dataloom-frontend/src/Components/Table.jsx`** — `handleCellClick` now does `setEditValue(cellValue ?? "")`, so missing cells (now `null`) don't make the controlled input uncontrolled or render the literal string `"null"`. This is the coordinated frontend half of the API contract change.

### Tests
- `tests/test_dtypes.py` — `test_preserves_nulls_without_coercing_them_to_empty_strings` and `test_preserves_non_scalar_cells_without_isna_truthiness_errors`.
- `tests/test_response_serialization.py` (new) — API-level assertion that uploaded missing values come back as `null`.
- `tests/test_new_features.py` — `test_add_column_with_blank_name_returns_422` paired with the existing schema-layer validator.

## Not included (already on `main`)
The `AddColumn.name` `field_validator`, the Pydantic v2 migration, the `pytest-asyncio` dep, the frontend payload-key changes, and the alembic-on-pytest handling were all verified present on current `main`, so they're intentionally omitted to keep this focused.

## Validation
- Backend: `uv run pytest -q` → **296 passed**; `uv run ruff check .` clean.
- Frontend: `npm run test` → **56 passed**; `npm run lint` and Prettier clean.

## Reference
- Follow-up to: #270 (supersedes #221)

### DIFF
--- a/dataloom-backend/app/utils/pandas_helpers.py
+++ b/dataloom-backend/app/utils/pandas_helpers.py
@@ -79,10 +79,16 @@ def dataframe_to_response(df: pd.DataFrame) -> dict[str, Any]:
     """
 
     dtypes = {col: _map_dtype(dtype) for col, dtype in df.dtypes.items()}
-    df = df.fillna("")
-    df = df.replace([float("inf"), float("-inf")], "")
     columns = df.columns.tolist()
-    rows = df.values.tolist()
+
+    # Preserve null semantics: real empty strings stay "", while missing and
+    # non-finite values serialize to null. ``where`` over an object-typed frame
+    # is a vectorized C-level dispatch, not a Python-level per-cell loop.
+    normalized_df = df.astype(object)
+    normalized_df = normalized_df.where(pd.notna(normalized_df), None)
+    normalized_df = normalized_df.replace([float("inf"), float("-inf")], None)
+
+    rows = normalized_df.values.tolist()
     return {
         "columns": columns,
         "rows": rows,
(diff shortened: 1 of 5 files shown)


ivantha:
### Automated review summary

**Findings**
- None flagged.

This is a focused, correct bugfix: dataframe_to_response previously ran df.fillna(\"\") + replace(inf, \"\"), collapsing genuine empty strings and missing values into the same \"\" and losing null semantics. The new df.astype(object).where(pd.notna(...), None).replace([inf,-inf], None) cleanly preserves \"\" while emitting null for missing/non-finite cells; I verified under the project's pandas 3.0.1 that it handles empty strings, NaN, inf, nullable Int64, datetimes (serialized to ISO by FastAPI's encoder), and non-scalar cells, and that the full suites pass (296 backend, 56 frontend) with ruff and ESLint clean. The coordinated frontend change setEditValue(cellValue ?? \"\") correctly keeps the edit input controlled when a cell is now null. No blocking, correctness, or security issues were found in the diff; safe to merge.

PR #331, opened 2026-06-03
chore: ignore root-level Python bytecode and scratch artifacts

nodesagar:
## Why

Optional hygiene follow-up to closed PR #269, as @ivantha suggested ("a separate hygiene PR with the `.gitignore` additions. Skip `.claude/` for now").

Most of #269's `.gitignore` additions already landed on `main` independently (Playwright dirs, `node_modules/`, `*.db`/`*.sqlite`/`*.sqlite3`, and `.claude/`). This PR adds only the rules still missing from the **root** `.gitignore`.

## What changed

Two new sections in the root `.gitignore`:

- **Python bytecode**: `__pycache__/`, `*.py[cod]`, `*$py.class` — these existed only in `dataloom-backend/.gitignore`, so bytecode from root-level scripts/tools wasn't covered.
- **Binary artifacts and scratch output**: `*.pdf`, `*.pid`, `tmp/`, `output/` — closing the gap that let binary artifacts leak into earlier PRs.

`.claude/` is intentionally left untouched (it's already present on `main`, and per the maintainer there's no project-wide stance to codify here).

## Reference
- Follow-up to: #269

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,17 @@ Desktop.ini
 .env.local
 .env.*.local
 
+# Python bytecode (root-level scratch scripts, tools, etc.)
+__pycache__/
+*.py[cod]
+*$py.class
+
+# Binary artifacts and scratch output
+*.pdf
+*.pid
+tmp/
+output/
+
 # Playwright
 test-results/
 playwright-report/


ivantha:
### Automated review summary

**Triage:** chore  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[minor/quality]** `.gitignore:35` The `*.pdf` rule is unanchored, so it ignores PDF files in every directory of the repo, not just the root-level scratch output the PR describes. If a PDF ever needs to be committed intentionally (e.g. a test fixture or a sample report), it will be silently ignored and require a confusing `git add -f`. Consider anchoring to the root (`/*.pdf`) or scoping it under a known scratch directory to match the stated 'root-level scratch output' intent. The same unanchored breadth applies to `tmp/` and `output/`, which match those directory names at any depth.

This PR is a clean, low-risk hygiene change that adds two sections to the root .gitignore: standard Python bytecode patterns (`__pycache__/`, `*.py[cod]`, `*$py.class`) and binary/scratch-artifact patterns (`*.pdf`, `*.pid`, `tmp/`, `output/`). No currently tracked files match any of the new patterns, so nothing is unintentionally untracked, and the diff applies cleanly to the current root .gitignore. The only suggestion is that `*.pdf` (and `tmp/`/`output/`) are repository-wide rather than anchored to the root; if PDFs or scratch dirs should only be ignored at the root, anchoring them (e.g. `/*.pdf`) would better match the PR's stated intent and avoid surprises if a PDF ever needs to be committed in a subdirectory. Safe to merge as-is, or with the optional anchoring tweak.

PR #343, opened 2026-06-06
fix: persist page size to localStorage and enforce pagination after transforms

### LINKED ISSUE #339: [Bug] Page size resets to default after applying a transformation
## Describe the bug

Two related problems with page size after applying a transformation:

1. **All rows shown regardless of page size** — transform forms call `updateData` with the full API response directly, bypassing pagination entirely. The page size selector still shows the user's chosen value (e.g. 10) but all rows are rendered.

2. **Page size resets to 50 on reload** — `pageSize` is held only in React state with a hardcoded default of 50. Any user preference is lost on refresh.

## To Reproduce

1. Change page size to 10
2. Apply any transformation (Sort, Filter, etc.)
3. Hit the `Cancel` button
4. Page size selector shows 10 but all rows are visible
5. Refresh the page — page size resets to 50

## Expected behavior

- After a transformation the table respects the user's chosen page size
- Page size persists across reloads

## Screenshots

Image

## Environment

| Field | Value |
|---|---|
| OS | Any |
| Browser | Any |
| Node.js version | Any |
| Python version | Any |

## Additional context

Fix requires two steps:

1. Persist `pageSize` in `localStorage` in `ProjectContext` — isolated to one file, can be done independently.
2. Call `refreshProject(projectId, 1, pageSize)` after `updateData` in all mutating transform forms so pagination is respected after every transform.

Step 2 is blocked by #338 — removes `onTransform` and standardises `updateData` calls across all transform forms. This fix will be included in the same PR as step 1 once that is merged.

SHIVAM-KUMAR-59:
## Description
Fixes two related pagination bugs after applying transformations.

1. Page size was not persisted — `pageSize` was held only in React state with a hardcoded default of 50, so any user preference was lost on reload.
2. Transform forms bypassed pagination entirely — after a transform, `updateData` was called with the full API response, ignoring the user's chosen page size.
3. Undo transformation also bypassed pagination — `handleUndo` in `MenuNavbar` was not passing `pageSize` to `refreshProject`, causing all rows to render regardless of chosen page size.

Fix persists `pageSize` to localStorage in `ProjectContext` and calls `refreshProject(projectId, 1, pageSize)` after `updateData` in all mutating transform forms.

Fixes #339 

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- Changed page size to 10, applied transformations — table correctly shows 10 rows
- Refreshed page — page size persisted as 10, not reset to 50
- Changed page size to 10, applied a transform, undid it — table correctly shows 10 rows after undo
- [x] Existing tests pass
- [x] Manual testing

## Screenshot
Screenshot 2026-06-06 at 19 25 11

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/MenuNavbar.jsx
+++ b/dataloom-frontend/src/Components/MenuNavbar.jsx
@@ -20,7 +20,6 @@ import {
   getLogs,
   getCheckpoints,
   revertToCheckpoint,
-  getProjectDetails,
   undoLastTransformation,
 } from "../api";
 import proptype from "prop-types";
@@ -66,7 +65,7 @@ const MenuNavbar = ({ projectId }) => {
   const [confirmData, setConfirmData] = useState(null);
   const [toast, setToast] = useState(null);
 
-  const { updateData, projectName } = useProjectContext();
+  const { updateData, refreshProject, pageSize, projectName } = useProjectContext();
 
   const fetchLogs = useCallback(async () => {
     try {
@@ -118,8 +117,7 @@ const MenuNavbar = ({ projectId }) => {
       setShowSortForm(false);
       setActiveForm(null);
       updateData([], [], { resetColumnOrder: false });
-      const data = await getProjectDetails(projectId);
-      updateData(data.columns, data.rows, { dtypes: data.dtypes, resetColumnOrder: false });
+      await refreshProject(projectId, 1, pageSize);
       await fetchLogs();
       setToast({ message: "Last transformation undone!", type: "success" });
     } catch (error) {
--- a/dataloom-frontend/src/Components/forms/AdvQueryFilterForm.jsx
+++ b/dataloom-frontend/src/Components/forms/AdvQueryFilterForm.jsx
@@ -5,12 +5,14 @@ import { transformProject } from "../../api";
 import { ADV_QUERY_FILTER } from "../../constants/operationTypes";
 import useError from "../../hooks/useError";
 import FormErrorAlert from "../common/FormErrorAlert";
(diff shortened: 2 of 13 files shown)


ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `dataloom-frontend/src/context/ProjectContext.jsx:32` pageSize is stored under the bare key 'pageSize' in localStorage, not namespaced by project or app. All DataLoom projects in the same browser origin will share one pageSize value. If per-project page sizes are ever desirable, or if another app on the same origin uses the same key, this will silently conflict. Consider a namespaced key such as 'dataloom:pageSize'.
- **[nit/quality]** `dataloom-frontend/src/Components/forms/AdvQueryFilterForm.jsx:26` In AdvQueryFilterForm, MeltForm, and PivotTableForm the code calls updateData(...) with the full API response immediately before calling refreshProject(...), which will call updateData again internally with the paginated slice. The first updateData call is redundant and causes an extra render. Removing the standalone updateData call before refreshProject in these three forms would be cleaner.

**Conventions**

This PR correctly fixes two related bugs: pageSize is now initialized from and persisted to localStorage so user preferences survive page reloads, and all 12 mutating transform forms now call refreshProject after updateData so the table respects the chosen page size post-transform. The fix is applied consistently across every affected form. Two minor issues worth addressing before merge: the localStorage key 'pageSize' is not namespaced and could collide with other keys on the same origin, and three forms (AdvQueryFilterForm, MeltForm, PivotTableForm) call updateData redundantly just before refreshProject, causing an extra render that could be removed.

PR #345, opened 2026-06-07
Fix: casting updates dtype badges

hanzalahwaheed:
## Description
Earlier casting was not updating dtype badges, which is unresponsive to the user.

Fixes #208 

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [X] New tests added
- [X] Manual testing

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -375,6 +375,8 @@ def cast_data_type(df: pd.DataFrame, column: str, target_type: str) -> pd.DataFr
             if target_type == "integer":
                 # Use pandas' nullable Int64 so NaNs from coerced values coexist with ints.
                 df[column] = df[column].astype("Int64")
+            else:
+                df[column] = df[column].astype(float)
         elif target_type == "boolean":
             truthy = {"true", "1", "yes", "y", "on"}
             falsy = {"false", "0", "no", "n", "off"}
--- a/dataloom-backend/app/utils/file_formats.py
+++ b/dataloom-backend/app/utils/file_formats.py
@@ -144,7 +144,7 @@ def _read_json(path: Path) -> pd.DataFrame:
     return pd.DataFrame(records)
 
 
-def _write_json(df: pd.DataFrame, path: Path, _) -> None:
+def _write_json(df: pd.DataFrame, path: Path, *_ignored) -> None:
     df.to_json(path, orient="records", indent=2)
 
 
--- a/dataloom-backend/tests/test_transformations.py
+++ b/dataloom-backend/tests/test_transformations.py
@@ -408,6 +408,12 @@ class TestCastDataType:
         assert pd.api.types.is_float_dtype(result["val"])
         assert result["val"].iloc[0] == 1.5
 
+    def test_cast_integer_like_values_to_float(self):
+        df = pd.DataFrame({"val": [1, 2, 3]})
+        result = cast_data_type(df, "val", "float")
+        assert pd.api.types.is_float_dtype(result["val"])
+        assert result["val"].tolist() == [1.0, 2.0, 3.0]
+
     def test_cast_to_boolean_truthy_falsy(self):
         df = pd.DataFrame({"flag": ["true", "false", "yes", "no", "1", "0"]})
         result = cast_data_type(df, "flag", "boolean")
--- a/dataloom-frontend/src/Components/Table.jsx
+++ b/dataloom-frontend/src/Components/Table.jsx
@@ -86,7 +86,7 @@ const Table = ({ projectId, data: externalData }) => {
     const { columns, rows, dtypes: newDtypes } = response;
     setColumns(["S.No.", ...columns]);
     setData(rows.map((row, index) => [index + 1, ...Object.values(row)]));
-    updateData(columns, rows, newDtypes);
+    updateData(columns, rows, { dtypes: newDtypes });
   };
 
   const handleAddRow = async (index) => {
--- a/dataloom-frontend/src/context/ProjectContext.jsx
+++ b/dataloom-frontend/src/context/ProjectContext.jsx
@@ -107,7 +107,7 @@ export function ProjectProvider({ children }) {
     (newColumns, newRows, options = {}) => {
       setColumns(newColumns);
       setRows(newRows);
-      if (options && options.dtypes) setDtypes(options.dtypes);
+      if (options.dtypes) setDtypes(options.dtypes);
       if (!projectId) return;
       setColumnOrders((prev) => {
         const existingOrder = prev[projectId];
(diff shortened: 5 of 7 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/ExportModal.tsx:757` When the TSV format button is selected in ExportModal and the user later manually picks 'tab' as the delimiter on the CSV format, then switches back to CSV, the effectiveDelimiter logic (line 699) only auto-substitutes 'tab' when format==tsv && delimiter=='comma'. A user who previously clicked 'tab' while on TSV and then returned to CSV will still get delimiter='tab', which is correct, but the reverse path (CSV->TSV->CSV) resets delimiter to 'comma' via the onClick handler on line 757 only if the previous delimiter was 'tab'. This is a minor UX inconsistency rather than data loss, but the auto-reset logic on line 757 (`if f.ext === 'csv' && delimiter === 'tab' setDelimiter('comma')`) silently overrides an explicit user selection of tab for CSV.
- **[minor/bug]** `dataloom-backend/app/utils/file_formats.py:236` The `_write_json` writer signature uses a bare `_` as the third parameter name, while `_write_xlsx` and `_write_parquet` use `*_ignored`. The inconsistency is harmless but `_` shadows Python's interactive `_` builtin. Prefer `*_ignored` or `_options` to match the other writers.
- **[minor/quality]** `dataloom-backend/app/api/endpoints/projects.py:92` In `export_project`, the `write_options` object is constructed and `has_options()` is called even on the native-export fast path (lines 92-100). This is harmless but means `TableWriteOptions` validation logic runs before the early-return check. Moving the early-return check to before `write_options` construction would be slightly cleaner and avoids any future surprise if option parsing ever raises.
- **[nit/quality]** `dataloom-backend/tests/test_multiformat_upload.py:463` The `test_failed_conversion_write_removes_temp_file` test patches `projects_endpoint.tempfile.NamedTemporaryFile` via module attribute access, which is fragile if the import path changes. Using `unittest.mock.patch` on the full dotted path (`app.api.endpoints.projects.tempfile.NamedTemporaryFile`) would be more robust.

**Conventions**
- **PR title:** Title mixes Conventional Commits prefix ('Fix:') with an unrelated scope; the PR also ships a large new export feature, making 'Fix' inaccurate for the bulk of the change. A title like 'feat(export): multi-format export with delimiter/encoding options; fix float cast dtype badge' would be more accurate.
- **Commits (3):** Two commits use 'feat:' (new export feature) and one uses 'fix(transform):' (the actual bug fix) — each is individually well-formed, but the PR title says 'Fix' while most of the work is new feature work, which misrepresents the change. The commit messages themselves follow Conventional Commits and imperative mood correctly.
- _The PR title should accurately reflect the dominant change (a new feature) rather than the minor bug fix it was named after._

The PR bundles two distinct changes: a one-line fix to `cast_data_type` so integer-to-float casts actually call `.astype(float)` (and the related `ProjectContext` dtype-badge fix), plus a full multi-format export feature with format conversion, delimiter/encoding options, an `ExportModal`, and an extensive test suite. The core bug fix is correct and well-tested. The export feature code is high quality — temp-file cleanup is safe, error handling is correct, and CORS `expose_headers` for `Content-Disposition` is a necessary detail that was not forgotten. The only concrete code issue worth addressing is the `_write_json` bare `_` parameter inconsistency and the fragile monkeypatching in `test_failed_conversion_write_removes_temp_file`. The PR title says 'Fix' but the bulk of the diff is new feature work, which should be reflected in both the title and the commit structure.

hanzalahwaheed:
@ivantha ready to merge. the flagged bug is an intentional UX decision.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:370` Dead branch: the `else: df[column] = df[column].astype(float)` is unreachable. The outer condition is `if target_type == 'numeric':`, so inside that block `target_type` is always the string `'numeric'`, making `target_type == 'integer'` permanently False. The `else` path therefore always executes — but if the intent was to handle float separately, the guard should be `if target_type == 'integer': ... elif target_type == 'float': df[column] = df[column].astype(float)` at the top level, not nested inside the numeric branch.
- **[major/bug]** `dataloom-frontend/src/context/ProjectContext.jsx:108` The backward-compat shim in `updateData` uses `Object.prototype.hasOwnProperty.call(options, 'dtypes')` to distinguish an options object from a raw dtype map. A dtype map whose column happens to be named 'dtypes' or 'resetColumnOrder' will be misclassified, causing dtypes to be silently dropped or `resetColumnOrder` to be read from a string value. Given that column names are user-supplied, this is a realistic collision path.
- **[minor/test]** `dataloom-frontend/src/test/ProjectContext.columnOrder.test.jsx:65` The new test `'updates dtypes when callers pass the dtype map directly'` exercises the legacy code path (passing a raw dtype map as the third argument). It should instead test the corrected call site in Table.jsx — passing `{ dtypes: newDtypes }` — to give regression coverage for the actual fix. The legacy path is now a compatibility shim and will ideally be removed.

**Conventions**
- **PR title:** Title uses title-case after the type prefix ('Fix: casting updates dtype badges') rather than the conventional lowercase imperative form; prefer 'fix: update dtype badges after cast operations'.
- _Consider squashing both commits into one before merge since together they represent a single atomic fix._

The core context-level fix (passing `{ dtypes: newDtypes }` instead of a raw map) is correct and closes the badge-regression, but two bugs remain: the float-cast branch in `transformation_service.py` is unreachable because it is nested inside an already-matching `if target_type == 'numeric':` block, and the backward-compat shim in `updateData` will silently mishandle dtype maps that contain a column named `dtypes` or `resetColumnOrder`. The test added for the legacy call path should be replaced with a test for the new `{ dtypes }` object form to actually guard the fix. Please address the unreachable branch and the shim collision risk before merging.

hanzalahwaheed:
@ivantha thanks for the review. addressed the comments.

ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**
- **PR title:** Title uses non-standard 'Fix:' casing; Conventional Commits expects lowercase scope, e.g. 'fix(casting): update dtype badges after float cast'.
- _PR title capitalisation is the only hygiene gap; everything else is clean._

The PR fixes two related bugs: casting an integer-typed column to float left the dtype unchanged because `pd.to_numeric` on an int Series returns int64, not float64, so the explicit `astype(float)` in the new `else` branch is necessary and correct. The companion fix in `Table.jsx` wraps `newDtypes` in the expected `{ dtypes: ... }` options object so `updateData` actually updates the dtype state after a cast, which is the root cause of the badge not refreshing. Both fixes are backed by new unit tests and the `localStorage` polyfill in `setup.js` is a legitimate prerequisite for the new `updateData` test that exercises `columnOrders`. No issues found; the change is ready to merge.

PR #349, opened 2026-06-08
fix(backend): guard cell-edit index bounds and remove dead helpers

### LINKED ISSUE #263: [Refactor] `get_project_by_id` in `project_service.py` is dead code
**Description:**

`app/services/project_service.py` line 33 defines `get_project_by_id` which queries a project by primary key and returns `None` if not found. It is never called — all endpoints use `get_project_or_404` from `app/api/dependencies.py` instead, which contains an identical query. The function serves no purpose and adds confusion about which lookup path to use.

Fix: delete `get_project_by_id` from `project_service.py`.

nodesagar:
## Summary

A small backend hardening + dead-code cleanup bundle covering three independent issues in the same area (transformation/service layer). No behavior change for valid inputs.

### Fixes #259 — `change_cell_value` missing lower-bound check
`change_cell_value` only validated the upper bound, so a negative `row_index`/`col_index` would silently wrap to the wrong cell (e.g. `col_index = 0` or a negative value indexing from the end). Added lower-bound guards (`row_index < 0`, `col_index < 1`; `col_index` is 1-based from the frontend).

### Fixes #263 — remove dead `get_project_by_id`
`get_project_by_id` in `project_service.py` had zero callers across the backend. Removed.

### Fixes #261 — remove dead `validate_row_index` / `validate_column_index`
Both functions in `pandas_helpers.py` had zero callers. Removed. (`HTTPException` import is retained — still used elsewhere in the file.)

## Tests
- Added 4 out-of-bounds cases for `change_cell_value` (negative row, negative/zero col, past-end row, past-end col).
- Full suite: **315 passed**.
- `ruff check` and `ruff format --check`: clean.

Closes #259
Closes #263
Closes #261

### DIFF
--- a/PR-OPPORTUNITIES.md
+++ b/PR-OPPORTUNITIES.md
@@ -1,0 +1,103 @@
+# DataLoom — PR Opportunities
+
+> Verified against `upstream/main` @ `2886f81` on 2026-06-08 (after fast-forwarding local `main`, was 48 commits behind).
+> Excludes reserved **GSoC #76** scope (multi-format upload/export, profiling, column-selector refactor, merge/join, formula columns + pipelines, visualization, data-quality engine) and any issue with an already-open PR.
+
+## Legend
+- **Effort**: S = <30 min · M = a few hours · L = day+
+- **Type**: bug / cleanup / feature / refactor / test / docs
+
+---
+
+## Tier 1 — Ready now (clean, confirmed-open, near-certain merges)
+
+### #259 — `change_cell_value` missing lower-bound check on `col_index`
+- **Type**: bug · **Effort**: S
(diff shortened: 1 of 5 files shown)


hanzalahwaheed:
@nodesagar thanks for the PR. I think it is better if we scope a PR to a single issue, however small the change might be. thanks!

nodesagar:
Thanks @hanzalahwaheed, that makes sense, scoped each change to its own PR:

- #350 — fix(backend): guard `change_cell_value` index bounds (closes #259)
- #351 — refactor(backend): remove unused `get_project_by_id` (closes #263)
- #352 — refactor(backend): remove unused index validators (closes #261)

Closing this bundle in favor of those three.

PR #350, opened 2026-06-08
fix(backend): guard change_cell_value index bounds (#259)

nodesagar:
## Summary
`change_cell_value` only validated the upper bound, so a negative `row_index` or `col_index` (or `col_index = 0`) would silently wrap to the wrong cell instead of being rejected. Added lower-bound guards.

`col_index` is 1-based from the frontend (it skips the S.No. display column), so the valid range is `1..len(columns)`; `row_index` is 0-based.

## Tests
- Added 4 out-of-bounds cases: negative row, zero/negative col, past-end row, past-end col.
- `pytest tests/test_transformations.py`: passes. `ruff check` / `ruff format --check`: clean.

Closes #259

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -224,7 +224,9 @@ def change_cell_value(df: pd.DataFrame, row_index: int, col_index: int, value) -
     """
     df = df.copy()
 
-    if row_index >= len(df) or col_index >= len(df.columns) + 1:
+    # col_index is 1-based (1..len(columns)); row_index is 0-based. Guard both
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**

The fix correctly adds lower-bound guards (row_index &lt; 0, col_index &lt; 1) to change_cell_value, closing the silent negative-index wrap-around that could corrupt a cell far from the intended target. Four targeted pytest cases cover each new branch and are well-named. The diff is minimal and surgical, the description explains the 1-based col_index convention clearly, and the single commit is clean. This is ready to merge.

PR #353, opened 2026-06-09
feat: display full checkpoint history with revert support

SHIVAM-KUMAR-59:
## Description
The CheckpointsPanel previously only displayed the most recent checkpoint as a single object, making it impossible for users to view or revert to older save points.

This PR replaces the single-checkpoint view with a full scrollable checkpoint history. The backend now exposes a dedicated endpoint returning all checkpoints for a project ordered by creation time, and the frontend panel maps over the full list, showing each checkpoint's message, timestamp, and a Revert button.

Fixes #88 

## Type of Change
- [x] New feature

## How Has This Been Tested?
- [x] Existing tests pass
- [x] Manual testing

**Following scenarios were tested manually:**
- Saved multiple checkpoints on a project
- Opened Checkpoints panel —> all checkpoints displayed in order
- Reverted to an older checkpoint —> data restored correctly
- Opened panel with no checkpoints —> empty state message shown
- Panel scrolls correctly with many checkpoints

## Screen Recording

https://github.com/user-attachments/assets/42d9dbd3-0533-49e0-adc7-043a87005449

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review
- [x] I have added/updated documentation as needed
- [x] My changes generate no new warnings
- [x] Tests pass locally

PR #355, opened 2026-06-09
feat: add delete checkpoint functionality with log unlinking

### LINKED ISSUE #354: [Feature] Delete Checkpoint — allow users to remove individual checkpoints
Users have no way to delete a checkpoint once it's created. Over time, a project can accumulate many checkpoints with no way to clean them up, making the checkpoint history cluttered and harder to navigate.

Add a Delete button to each checkpoint entry in the CheckpointsPanel. Clicking Delete should remove the checkpoint from the database and unlink its associated logs (preserving transformation history). The panel should refresh after deletion.

Clearing all checkpoints at once, rejected as too destructive. Users may want to keep some checkpoints while removing others.

**Additional context**
- Backend: `DELETE /logs/checkpoints/{project_id}/{checkpoint_id}` endpoint
- Unlinking logs on delete preserves transformation history

SHIVAM-KUMAR-59:
## Description
Users had no way to delete checkpoints once created. This PR adds a Delete button to each checkpoint entry in the CheckpointsPanel. Deleting a checkpoint removes it from the database and unlinks its associated logs, preserving transformation history without the checkpoint reference.

Fixes #354 

## Type of Change
- [x] New feature

## How Has This Been Tested?
- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

**Following scenarios were tested manually:**
- Created multiple checkpoints, deleted each —> panel refreshes correctly
- Verified logs are unlinked (`checkpoint_id` set to null) not deleted after checkpoint removal
- Verified 404 returned when deleting nonexistent checkpoint

## Screen Recording

https://github.com/user-attachments/assets/1ede40e5-11e4-44a1-ab1f-2b41a0c90932

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/user_logs.py
+++ b/dataloom-backend/app/api/endpoints/user_logs.py
@@ -5,7 +5,7 @@ from sqlmodel import Session
 
 from app import database, models, schemas
 from app.api import dependencies
-from app.services.project_service import get_checkpoints
+from app.services.project_service import delete_checkpoint, get_checkpoints
 
 router = APIRouter()
 
@@ -44,3 +44,16 @@ def get_project_checkpoints(
 ):
     """Fetch all checkpoints for a project ordered by creation time."""
     return get_checkpoints(db, project_id)
+
+
+@router.delete("/checkpoints/{project_id}/{checkpoint_id}")
+def delete_project_checkpoint(
+    project_id: uuid.UUID,
+    checkpoint_id: uuid.UUID,
+    db: Session = Depends(database.get_db),
+    _project: models.Project = Depends(dependencies.get_project_or_404),
+):
+    """Delete a checkpoint and unlink its associated logs."""
+    delete_checkpoint(db, checkpoint_id, project_id)
+    db.commit()
+    return {"success": True, "message": "Checkpoint deleted"}
--- a/dataloom-backend/app/services/project_service.py
+++ b/dataloom-backend/app/services/project_service.py
@@ -218,3 +218,41 @@ def delete_change_log(db: Session, log: models.ProjectChangeLog) -> None:
     db.delete(log)
     db.flush()
     logger.debug("Deleted change log: id=%s, project_id=%s", log.change_log_id, log.project_id)
+
+
+def delete_checkpoint(db: Session, checkpoint_id: uuid.UUID, project_id: uuid.UUID) -> None:
+    """Delete a checkpoint and unlink its associated logs.
+
+    Logs that referenced this checkpoint are not deleted — they are unlinked
+    (checkpoint_id set to None) so the transformation history is preserved.
+
+    Args:
+        db: Database session.
+        checkpoint_id: The checkpoint to delete.
+        project_id: The project the checkpoint belongs to.
+
+    Raises:
+        HTTPException: If the checkpoint is not found.
+    """
+    from fastapi import HTTPException
+
+    checkpoint = (
+        db.query(models.Checkpoint)
+        .filter(
+            models.Checkpoint.id == checkpoint_id,
+            models.Checkpoint.project_id == project_id,
+        )
+        .first()
+    )
+
+    if not checkpoint:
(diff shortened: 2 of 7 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/MenuNavbar.jsx:82` fetchCheckpoints always stores only the first element of the array returned by getCheckpoints (checkpointsResponse[0]), so every checkpoint beyond the first is silently discarded. If a project has more than one checkpoint, the panel can only ever display or delete the most recent one; the others become unreachable in the UI.
- **[major/bug]** `dataloom-backend/app/services/project_service.py:252` delete_checkpoint calls db.commit() directly, but every other mutating service function in project_service.py (e.g. delete_change_log) calls db.flush() and leaves the commit to the caller/endpoint layer. This breaks the flush-then-commit convention used throughout the service layer and could cause partial transactions if the pattern is ever composed with other service calls.
- **[minor/quality]** `dataloom-backend/app/services/project_service.py:228` HTTPException is imported inside the function body rather than at the module level. All other FastAPI exceptions in the codebase are imported at the top of the file. Move the import to the top of project_service.py.
- **[minor/quality]** `dataloom-frontend/src/Components/history/CheckpointsPanel.jsx:58` The Delete button has no confirmation step. A single misclick permanently removes the checkpoint. A simple window.confirm() or a modal guard would prevent accidental deletions.
- **[nit/style]** `dataloom-frontend/src/Components/history/CheckpointsPanel.jsx:58` A <td> element is given display:flex (via the Tailwind class 'flex') to center its children. Table cells are block-level, so applying flex directly to a <td> is technically valid but unexpected; wrapping the buttons in an inner <div className='flex items-center justify-center gap-2'> is the conventional approach.

**Conventions**

The backend endpoint and service logic are solid -- the ownership guard, log-unlinking semantics, and three targeted tests are all well done. Two issues need fixing before merge: the frontend silently drops all but the first checkpoint when refreshing (fetchCheckpoints stores only index 0), making every checkpoint after the first unreachable; and delete_checkpoint calls db.commit() directly rather than db.flush(), breaking the flush-delegate-commit convention used by every other service function in the file. The local import of HTTPException and the missing deletion confirmation are smaller but worth addressing.

SHIVAM-KUMAR-59:
Addressed all findings: 
- **[major/bug] `fetchCheckpoints` dropping all but first checkpoint** — fixed. `fetchCheckpoints` now stores the full array. For backward compatibility with the current single-object endpoint, the response is normalized: if an array is returned it's stored directly; if a single object is returned it's wrapped in an array; otherwise an empty array is set. This will be a no-op once #353 merges and the endpoint returns a full array.

- **[major/bug] `delete_checkpoint` calling `db.commit()` directly** — fixed. Changed to `db.flush()` to follow the flush-delegate-commit convention used throughout the service layer. `db.commit()` is now called in the endpoint layer in `logs.py`.

- **[minor/quality] `HTTPException` imported inside function body** — fixed. Moved to module-level imports at the top of `project_service.py`.

- **[minor/quality] Missing confirmation on delete** — fixed. Added a confirmation modal using the existing `Modal` component before proceeding with deletion.
 
- **[nit/style] `flex` applied directly to `<td>`** — fixed. Buttons are now wrapped in an inner `<div className="flex items-center justify-center gap-2">` inside the `<td>`.
 
Ready for re-review. @ivantha

SHIVAM-KUMAR-59:
Resolved merge conflicts across the affected files. Here's a summary of how each was handled:

- **`CheckpointsPanel.jsx`** — kept the full array-based render with `useToast` addition from the other branch, `Modal` confirmation dialog, and updated PropTypes to `arrayOf`
- **`logs.py`** — merged both imports (`get_checkpoints` and `delete_checkpoint`) and kept both endpoints: `GET /checkpoints/{project_id}` and `DELETE /checkpoints/{project_id}/{checkpoint_id}`

Ready for re-review. @ivantha

PR #362, opened 2026-06-11
feat: add fill empty transformation strategies

### LINKED ISSUE #90: Add Fill NA Strategy Options
**Difficulty:** Medium
**Area:** full stack

## Description

DataLoom's current `fill_empty()` function accepts a single fill value and replaces all missing (NaN/null) values in a column with that literal value. While this covers the simplest use case, real-world data cleaning often requires more sophisticated strategies for handling missing data.

This issue extends the fill empty functionality to support multiple strategies: fill with the column's mean, median, or mode (most frequent value), forward fill (propagate the last valid value forward), and backward fill (propagate the next valid value backward). These strategies are standard in data science workflows and are natively supported by pandas.

The user should be able to choose a strategy from a dropdown. When "custom value" is selected, a text input appears for entering the literal fill value (existing behavior). When a statistical strategy is selected (mean, median, mode), the value is computed from the column data. When forward fill or backward fill is selected, pandas' built-in `ffill()` or `bfill()` methods are used.

## Acceptance Criteria

- [ ] Six fill strategies are supported: custom value, mean, median, mode, forward fill, and backward fill
- [ ] Filling missing values with column mean produces correct results
- [ ] Filling missing values with column median produces correct results
- [ ] Filling missing values with column mode produces correct results
- [ ] Forward fill correctly propagates the last valid value forward
- [ ] Backward fill correctly propagates the next valid value backward
- [ ] Custom value fill continues to work (backward compatible)
- [ ] Mean/median on non-numeric columns returns a clear error
- [ ] The frontend shows a strategy dropdown with all six options
- [ ] The custom value input is conditionally visible only when the "custom" strategy is selected
- [ ] Transformation is logged with the chosen strategy in the action details
- [ ] Data table refreshes to show the filled values

## Resources

- `dataloom-backend/app/services/transformation_service.py` — `fill_empty()` function
- `dataloom-backend/app/schemas.py` — enum definitions
- `dataloom-frontend/src/Components/MenuNavbar.jsx` — menu where fill empty is triggered
- [pandas `fillna` documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html)
- [pandas `ffill` documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.ffill.html)
- [pandas `bfill` documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.bfill.html)

SHIVAM-KUMAR-59:
## Description

Adds support for multiple strategies when filling missing values in datasets.

### Supported Strategies

- Custom value (existing behavior)
- Mean
- Median
- Mode
- Forward Fill (ffill)
- Backward Fill (bfill)

### Changes Made

- Added `FillStrategy` enum and strategy support in the backend.
- Extended `fill_empty()` to support statistical and propagation-based fill operations.
- Added validation for non-numeric mean/median operations.
- Added replay support for fill strategies through the transformation registry.
- Added a Fill Empty form to the frontend.
- Added a strategy dropdown with conditional custom value input.
- Fixed active state and form close behavior for the new menu item.
- Added backend unit tests covering all supported fill strategies and error cases.

Fixes #90 

## Type of Change
- [x] New feature

## How Has This Been Tested?

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

- Added backend unit tests for:

  - Custom value fill
  - Mean fill
  - Median fill
  - Mode fill
  - Forward fill
  - Backward fill
  - Invalid non-numeric mean/median operations
  - Transformation replay support

### Manual Testing

Verified that:

- Custom value fill works as before.
- Mean, median, and mode correctly fill missing values.
- Forward fill propagates the previous value.
- Backward fill propagates the next value.
- Mean/median on non-numeric columns return an error.
- Fill Empty menu item opens and closes correctly.
- Active toolbar state behaves consistently with other transformations.
- Data refreshes immediately after applying a transformation.

## Screen Recording

https://github.com/user-attachments/assets/995c6c51-fb4f-4cd2-9270-f34843ba105c

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -53,6 +53,17 @@ class DropDup(StrEnum):
     last = "last"
 
 
+class FillStrategy(StrEnum):
+    """Supported strategies for filling empty cells."""
+
+    custom = "custom"
+    mean = "mean"
+    median = "median"
+    mode = "mode"
+    ffill = "ffill"
+    bfill = "bfill"
+
+
 class AggFunc(StrEnum):
     """Supported aggregation functions for pivot tables."""
 
@@ -141,7 +152,8 @@ class FillEmptyParams(BaseModel):
     """Parameters for filling empty cells."""
 
     index: int | None
-    fill_value: Any
+    fill_value: Any | None = None
+    strategy: FillStrategy = FillStrategy.custom
 
 
 class RenameColumnParams(BaseModel):
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -9,7 +9,7 @@ from dataclasses import dataclass
 
 import pandas as pd
 
-from app.schemas import MeltParams, OperationType
+from app.schemas import FillStrategy, MeltParams, OperationType
 from app.utils.logging import get_logger
 from app.utils.security import validate_query_string
 
@@ -297,26 +297,127 @@ def change_cell_value(df: pd.DataFrame, row_index: int, col_index: int, value) -
     return df
 
 
-def fill_empty(df: pd.DataFrame, fill_value, column_index: int = None) -> pd.DataFrame:
-    """Fill empty/NaN cells with a specified value.
+def _coerce_fill_value(value):
+    """Attempt to coerce a string fill value to int or float for correct typing.
+
+    Falls back to the original string if it cannot be parsed as a number.
+    Whitespace-only strings are rejected upstream by the custom-strategy guard
+    in fill_empty(), so this function assumes the input has already been
+    validated as non-blank.
+    """
+    if isinstance(value, str):
+        try:
+            return int(value)
+        except ValueError:
+            try:
+                return float(value)
+            except ValueError:
+                return value
+    return value
+
+
+def _coerce_fill_value_for_column(value, column: pd.Series):
+    """Coerce a fill value to match a column's dtype when applying across all columns.
+
+    Only numeric columns attempt int/float coercion; non-numeric columns receive
+    the raw string unchanged. This prevents a single coerced value (e.g. an int)
+    from being applied to a string column when filling every column at once.
+    """
+    if not isinstance(value, str):
+        return value
+    if pd.api.types.is_numeric_dtype(column):
+        return _coerce_fill_value(value)
+    return value
+
+
+def fill_empty(
+    df: pd.DataFrame,
+    fill_value=None,
+    column_index: int = None,
+    strategy: str = "custom",
+) -> pd.DataFrame:
+    """Fill empty/NaN cells using a specified strategy.
 
     Args:
         df: Source DataFrame.
-        fill_value: Value to fill empty cells with.
-        column_index: Optional specific column index. If None, fills all columns.
+        fill_value: Literal fill value used for the custom strategy.
+        column_index: Optional specific column index. If None, fills all columns
+            for custom, ffill, and bfill strategies.
+        strategy: One of custom, mean, median, mode, ffill, or bfill.
 
     Returns:
         DataFrame with empty cells filled.
     """
+
+    if strategy not in FillStrategy:
+        raise TransformationError(f"Unsupported fill strategy: '{strategy}'")
+
+    if strategy == "custom" and (fill_value is None or (isinstance(fill_value, str) and fill_value.strip() == "")):
+        raise TransformationError("A fill value is required when using the custom strategy")
+
     df = df.copy()
+
     if column_index is not None:
         if column_index < 0 or column_index >= len(df.columns):
             raise TransformationError(f"Column index {column_index} out of range")
+
         column_name = df.columns[column_index]
-        df[column_name] = df[column_name].fillna(fill_value)
-    else:
-        df = df.fillna(fill_value)
-    return df
+        series = df[column_name]
+
+        if strategy == "custom":
+            df[column_name] = series.fillna(_coerce_fill_value(fill_value))
+
+        elif strategy == "mean":
+            numeric_series = pd.to_numeric(series, errors="coerce")
+            if numeric_series.dropna().empty:
+                raise TransformationError(f"Cannot compute mean on non-numeric column '{column_name}'")
+
+            non_numeric_mask = series.notna() & numeric_series.isna()
+            if non_numeric_mask.any():
+                raise TransformationError(f"Cannot compute mean on column '{column_name}': contains non-numeric values")
+
+            computed_fill_value = round(float(numeric_series.mean()), 2)
+            df[column_name] = series.fillna(computed_fill_value)
+
+        elif strategy == "median":
+            numeric_series = pd.to_numeric(series, errors="coerce")
+            if numeric_series.dropna().empty:
+                raise TransformationError(f"Cannot compute median on non-numeric column '{column_name}'")
+
+            non_numeric_mask = series.notna() & numeric_series.isna()
+            if non_numeric_mask.any():
+                raise TransformationError(
+                    f"Cannot compute median on column '{column_name}': contains non-numeric values"
+                )
+
+            computed_fill_value = round(float(numeric_series.median()), 2)
+            df[column_name] = series.fillna(computed_fill_value)
+
+        elif strategy == "mode":
+            mode_value = series.mode(dropna=True)
+            if mode_value.empty:
+                raise TransformationError(f"Cannot compute mode: column '{column_name}' has no non-empty values")
+
+            df[column_name] = series.fillna(mode_value.iloc[0])
+
+        elif strategy == "ffill":
+            df[column_name] = series.ffill()
+
+        elif strategy == "bfill":
+            df[column_name] = series.bfill()
+
+        return df
+
+    if strategy == "custom":
+        return df.apply(lambda col: col.fillna(_coerce_fill_value_for_column(fill_value, col)))
+
+    if strategy == "ffill":
+        return df.ffill()
+
+    if strategy == "bfill":
+        return df.bfill()
+
+    raise TransformationError(f"Strategy '{strategy}' requires a specific column to be selected")
 
 
 def rename_column(df: pd.DataFrame, col_index: int, new_name: str) -> pd.DataFrame:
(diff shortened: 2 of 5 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:301` When strategy is 'custom' and fill_value is an empty string (""), the guard `if fill_value is None` passes and the empty string is used as the fill value, silently producing wrong results. The check should be `if fill_value is None or fill_value == ""` (or use a falsy check), since the form always sends an empty string when the user leaves the field blank and the HTML `required` attribute does not protect the API.
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:323` For mean and median with a specific column, `pd.to_numeric(series, errors='coerce')` silently coerces non-numeric values to NaN. If the column has both numeric and non-numeric values, the check `numeric_series.dropna().empty` passes, but the computed mean/median is only over the numeric subset without any warning. A partially-numeric column should raise an error or at minimum document this behavior clearly.
- **[minor/bug]** `dataloom-frontend/src/Components/forms/FillEmptyForm.tsx:46` The FillEmptyForm sends `fill_value: null` for non-custom strategies. The backend FillEmptyParams schema defines `fill_value: Any`, so null is accepted. However, when strategy is 'custom' and the user submits, the form sends the raw string from `fillValue` state. If that string is numeric (e.g. '42'), it will be stored and replayed as a string, not an integer, potentially causing type mismatches on re-application. The backend should attempt type coercion for custom fill values.
- **[minor/quality]** `dataloom-frontend/src/Components/forms/FillEmptyForm.tsx:125` FillEmptyForm.tsx uses both TypeScript inline type annotations in the function signature (`{projectId: string, onClose: () => void}`) and a separate `PropTypes` declaration at the bottom. These are redundant; in a TypeScript file PropTypes provide no compile-time benefit and the project's other TypeScript forms (e.g. StringReplaceForm) should be checked for the prevailing convention. One approach should be chosen and applied consistently.
- **[nit/quality]** `dataloom-backend/app/services/transformation_service.py:296` The `valid_strategies` tuple in `fill_empty()` duplicates the values already defined in the `FillStrategy` StrEnum. The validation check should iterate over `FillStrategy` members instead to keep the two definitions in sync automatically.
- **[nit/style]** `dataloom-frontend/src/Components/forms/FillEmptyForm.tsx:28` FillEmptyForm.tsx declares `useProjectContext()` twice (lines 23 and 28), destructuring different fields each time. Both destructures should be merged into a single `useProjectContext()` call.

**Conventions**

The PR extends the fill-empty transformation with five new strategies (mean, median, mode, ffill, bfill), adds replay support through the transformation registry, and ships a corresponding frontend form. The test coverage is thorough and the overall structure is clean. Two issues should be addressed before merge: the custom-strategy guard does not catch an empty-string fill value sent from the form (allowing silent bad fills), and the mixed numeric/non-numeric column case for mean/median silently drops non-numeric values without warning. The minor items -- a duplicate `useProjectContext` call, redundant PropTypes alongside TypeScript types, and the inline `valid_strategies` tuple that drifts from the `FillStrategy` enum -- are quick cleanups worth folding in.

SHIVAM-KUMAR-59:
Addressed all findings:

- **[major/bug] empty string custom fill value** — fixed. The guard now checks `fill_value is None or fill_value == ""`, catching the empty-string case sent by the form when the field is left blank.

- **[major/bug] mixed numeric/non-numeric columns for mean/median** — fixed. Added a check that detects when `pd.to_numeric` coerces non-null values to NaN (i.e. genuinely non-numeric values present), and raises a `TransformationError` in that case rather than silently computing over the numeric subset.

- **[minor/bug] custom fill value type coercion** — fixed. Added `_coerce_fill_value()` helper that attempts to coerce string fill values to `int` or `float` before filling, so numeric strings from the form are stored and replayed with the correct type.

- **[minor/quality] redundant PropTypes alongside TS types** — fixed. Removed the `PropTypes` import and declaration; the TypeScript inline type annotation is now the single source of truth, matching the convention in other TS forms.

- **[nit/quality] `valid_strategies` tuple duplicating `FillStrategy` enum** — fixed. Validation now iterates over `FillStrategy` members directly, so the two definitions can't drift.

- **[nit/style] duplicate `useProjectContext` call** — fixed. Merged into a single destructure at the top of the component.

Ready for re-review. @ivantha

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:342` When 'All columns' is selected and strategy is 'custom', `_coerce_fill_value` is called with `fill_value` that could be an empty string because the frontend sends `fillValue` state (defaulting to "") for the custom strategy. The backend guard `if strategy == 'custom' and (fill_value is None or fill_value == '')` will catch this, but the coercion path on line 342 (`df.fillna(_coerce_fill_value(fill_value))`) is only reached after the guard, which is correct — however the coercion function itself returns the string unchanged when it cannot be cast, so a user entering a value like ' ' (whitespace) will silently fill NaNs with a whitespace string. This is a UX/correctness edge case with no validation on either side.
- **[minor/quality]** `dataloom-backend/app/schemas.py:154` The frontend sends `fill_value: null` for non-custom strategies, but the backend `FillEmptyParams` schema defines `fill_value: Any` with no default and no `Optional` annotation. Pydantic will accept `null` as `None` here, but the intent is ambiguous. Making `fill_value` explicitly `Optional[Any] = None` in the schema would express the contract clearly.
- **[minor/quality]** `dataloom-frontend/src/Components/forms/FillEmptyForm.tsx:50` FillEmptyForm.tsx is a TypeScript file being imported from MenuNavbar.jsx (a plain JSX file) with no type issues at runtime, but the component uses `(err as any)` on line 50 to silence the unknown-error type. Narrowing the error type (e.g. using an Axios error type guard from the existing codebase pattern) would be consistent with the project's TypeScript usage.
- **[nit/quality]** `dataloom-frontend/src/Components/forms/FillEmptyForm.tsx:82` The hint text 'Note: Mean, median, and mode require a specific column.' is a static paragraph always visible below the column selector, but it does not prevent the user from choosing 'All columns' with a statistical strategy and submitting — the error only surfaces from the backend. A simple client-side check (disabling submit or showing an inline error when strategy is mean/median/mode and selectedColumn is empty) would improve UX without adding complexity.

**Conventions**
- _The description is thorough and links the correct issue; commit history is clean with two well-scoped conventional commits._

The PR extends `fill_empty` with five new strategies (mean, median, mode, ffill, bfill), adds a backend enum, proper error guards, transformation-replay support, and a matching frontend form with good test coverage across both unit and replay paths. The implementation is solid overall. Two items worth addressing before merge: the frontend does not prevent submitting a statistical strategy without a selected column (users hit a backend error instead of an in-form message), and `FillEmptyParams.fill_value` in the schema should be typed `Optional[Any] = None` to make the nullable contract explicit. The `_coerce_fill_value` helper is a nice touch but should be documented or guarded against whitespace-only strings on the custom path.

SHIVAM-KUMAR-59:
Addressed all findings:

- **[major/bug] whitespace-only custom fill value** — fixed. The custom-strategy guard now strips whitespace before checking for blank values, so `"   "` is rejected the same as `""`.` _coerce_fill_value` documentation updated to note it assumes input is already validated as non-blank.
- **[minor/quality] `FillEmptyParams.fill_value` not explicitly Optional** — fixed. Changed to `fill_value: Any | None = None` to make the nullable contract explicit.
- **[minor/quality] as any error narrowing in `FillEmptyForm.tsx`** — fixed. Replaced with a type-narrowed check on the error shape instead of casting to `any`.
- **[nit/quality] no client-side validation for stats strategies without a column** — fixed. Added a `requiresColumn` check that disables the Apply button and shows an inline toast if mean/median/mode is selected without a specific column, avoiding a round-trip to the backend for this case.

Ready for re-review. @ivantha

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:380` When strategy is 'custom' and column_index is None, _coerce_fill_value(fill_value) is called with the raw string from the API. For a DataFrame that has mixed-type columns (e.g., one numeric and one string), applying df.fillna(int_value) will silently coerce or raise a TypeError at runtime. The mean/median/mode guard that rejects no-column usage does not apply here, so a user can supply strategy='custom', fill_value='abc', column_index=None and overwrite every column with 'abc' without any type warning.
- **[minor/quality]** `dataloom-backend/app/services/transformation_service.py:332` The strategy-enum membership check at line 332 re-materialises the full set on every call: `{s.value for s in FillStrategy}`. Because FillStrategy is a StrEnum, the check can be simplified to `if strategy not in FillStrategy` which is both cleaner and avoids rebuilding the set each invocation.
- **[minor/quality]** `dataloom-frontend/src/Components/MenuNavbar.jsx:14` FillEmptyForm.tsx is a TypeScript file imported from MenuNavbar.jsx which is plain JavaScript. The import works at runtime but bypasses TypeScript checking for the consumer and is inconsistent with the rest of the codebase (other forms are .jsx). Either convert the form to .jsx or convert MenuNavbar to .tsx to maintain consistency.
- **[minor/quality]** `dataloom-frontend/src/Components/forms/FillEmptyForm.tsx:57` The error-handling block in handleSubmit casts err through a verbose inline type assertion. The project's other forms centralise this pattern in the useError hook. Extracting the detail-extraction logic into a shared utility or the existing hook would reduce duplication.
- **[nit/style]** `dataloom-frontend/src/Components/forms/FillEmptyForm.tsx:122` Indentation inconsistency in the button row: the Apply button is indented with 9 spaces while the Cancel button uses 10. Both should be 10 spaces to match the surrounding JSX.

**Conventions**
- **Commits (3):** The two fixup commits ('fix: address fill-empty review — empty value guard…' and 'fix: address fill-empty review — whitespace guard…') are review-iteration noise that should be squashed into the main feature commit before merging; they fragment the history without adding independent value.
- _Squash the two fixup commits into the primary feature commit so the final merge produces a single, atomic history entry._

This PR extends the fill-empty transformation with five new strategies (mean, median, mode, ffill, bfill), adds a corresponding frontend form, and ships thorough backend tests including replay coverage. The implementation is solid overall: validation, error messages, and test assertions are well-considered. Two points need attention before merging: the all-columns custom-fill path applies the coerced value to every column without type-checking, which can silently corrupt mixed-type DataFrames; and the new FillEmptyForm.tsx mixes TypeScript into a JSX component tree that the rest of the project keeps in .jsx, creating a minor inconsistency. The three-commit history should be squashed to a single clean entry before merge.

SHIVAM-KUMAR-59:
Addressed all findings:

- **[major/bug] all-columns custom fill silently corrupting mixed-type DataFrames** — fixed. Added `_coerce_fill_value_for_column()` which only attempts int/float coercion when the target column is numeric; non-numeric columns receive the raw string value unchanged. This prevents a single coerced value from corrupting columns of a different type when filling across all columns.
- **[minor/quality] enum membership check rebuilding a set on every call** — fixed. Simplified to `if strategy not in FillStrategy`, leveraging `StrEnum's` built-in membership check.
- **[minor/quality] .tsx file imported from .jsx MenuNavbar** — this is intentional. The project is mid-migration from JSX to TSX, so `FillEmptyForm.tsx` is correctly written in TypeScript ahead of `MenuNavbar.jsx's` eventual conversion. Keeping new code in .tsx aligns with the migration direction rather than working against it.

Also squashed the fixup commits into the primary feature commit for a clean, atomic history.

Ready for re-review. @ivantha

PR #369, opened 2026-06-15
feat: add data profiling endpoints with type-aware column stats

### LINKED ISSUE #358: [Feature] Data Profiling/Statistics
- Missing values, duplicate rows, inconsistent data types, outliers, and suspicious columns are hard to detect quickly.
- This slows down decision-making and increases the chance of applying wrong transformations.
- A data profiling feature would make dataset quality visible upfront.

- Add a data profiling view for each uploaded project or dataset.
- Show column-wise summaries such as data type, null count, unique values, duplicates, min/max, mean, and common values.
- Highlight potential data quality issues with clear indicators.
- Let users quickly understand the dataset before filtering, casting, cleaning, or transforming it.

Profiling is a native feature in data wrangling tools. No alternatives as such.

**Additional context**
The first version can focus on useful summary statistics and quality warnings before adding advanced charts or recommendations. Slow and steady!

hanzalahwaheed:
## Description

- add profiling service with dataset summary, column profile, and correlation matrix
- expose three read-only GET routes under /projects/{id}/profile (summary, column, correlation)
- treat missing-value sentinels (N/A, null, unknown, -, empty) as nulls, case-insensitively
- branch column profiles by type: numeric (incl. zero/negative counts), categorical, boolean, datetime
- label value distributions (constant, binary, skewed, zero-inflated, high-cardinality, normal-ish)
- export map_dtype helper so profiling dtype labels match the table API
- add thin TypeScript profiling API client and re-export it from the api barrel

Fixes #358

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [X] New tests added
- [X] Manual testing

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/profiling.py
+++ b/dataloom-backend/app/api/endpoints/profiling.py
@@ -1,0 +1,79 @@
+"""Data profiling API endpoints.
+
+Three read-only routes over a project's current working copy: a dataset
+summary, a per-column profile, and a numeric correlation matrix. All are
+descriptive — quality assessment is out of scope here.
+"""
+
+import uuid
+
+import pandas as pd
+from fastapi import APIRouter, Depends, HTTPException, Query
+
+from app import models, schemas
+from app.api.dependencies import get_project_or_404
+from app.services import profiling_service
+from app.utils.logging import get_logger
+from app.utils.pandas_helpers import read_table_safe
+
+logger = get_logger(__name__)
+
+router = APIRouter()
+
+
+def _load_project_df(project: models.Project) -> pd.DataFrame:
+    """Read the project's working copy, redacting any path from error details.
+
+    ``read_table_safe`` embeds the absolute file path in its 404/500 details;
+    surface a generic message to the client instead, matching the transform
+    endpoint's handling.
+    """
+    try:
+        return read_table_safe(project.file_path)
+    except HTTPException as e:
+        if e.status_code == 404:
+            raise HTTPException(status_code=404, detail="Project data file not found") from e
+        logger.warning(
+            "Failed to read project file for profiling project_id=%s status=%s", project.project_id, e.status_code
+        )
+        raise HTTPException(status_code=e.status_code, detail="Could not read project data") from e
+
+
+@router.get("/{project_id}/profile/summary", response_model=schemas.DatasetSummaryResponse)
+async def get_dataset_summary(
+    project_id: uuid.UUID,
+    project: models.Project = Depends(get_project_or_404),
+):
+    """Return a top-level summary of the project's dataset."""
+    df = _load_project_df(project)
+    return profiling_service.dataset_summary(df)
+
+
+@router.get("/{project_id}/profile/column", response_model=schemas.ColumnProfileResponse)
+async def get_column_profile(
+    project_id: uuid.UUID,
+    column_name: str = Query(..., description="Name of the column to profile"),
+    project: models.Project = Depends(get_project_or_404),
+):
+    """Return a type-aware profile of a single column.
+
+    The column name is taken as a query parameter rather than a path segment so
+    that names containing slashes (or other reserved characters) route reliably
+    across servers and reverse proxies, which handle ``%2F`` in paths
+    inconsistently.
+    """
+    df = _load_project_df(project)
+    try:
+        return profiling_service.column_profile(df, column_name)
+    except KeyError as e:
+        raise HTTPException(status_code=404, detail=f"Column '{column_name}' not found") from e
+
+
+@router.get("/{project_id}/profile/correlation", response_model=schemas.CorrelationResponse)
+async def get_correlation_matrix(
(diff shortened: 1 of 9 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/profiling_service.py:289` Boolean columns are appended to categorical_columns in dataset_summary (line 289), but booleans are neither categorically distributed nor surfaced as categorical in column_profile (they get their own boolean block). A downstream consumer iterating categorical_columns to call getColumnProfile will receive a boolean-typed response with no top_values/cardinality fields, breaking any code that assumes categorical shape. datetime columns silently fall into neither list, so the summary undercounts column types.
- **[minor/bug]** `dataloom-backend/app/api/endpoints/profiling.py:58` A column name that contains a forward slash (e.g. 'revenue/cost') cannot be profiled via GET /{project_id}/profile/columns/{column_name} because FastAPI does not capture multi-segment path params by default and will return a 404. Using path=True on the Path() parameter, or switching to a query parameter, would fix this.
- **[minor/quality]** `dataloom-backend/app/services/profiling_service.py:325` detect_distribution checks unique_count == 2 (binary) before the zero-inflated guard, so a column like [0, 0, 0, 1] with exactly two distinct values is labelled 'binary' rather than 'zero-inflated' even when zero_fraction >= ZERO_INFLATED_THRESHOLD. Whether this priority is intentional should be documented in the docstring, since callers cannot currently distinguish the two cases.
- **[nit/quality]** `dataloom-frontend/src/api/index.js:17` profiling.ts exports a TypeScript module from the JS barrel (src/api/index.js) without a corresponding .js shim. Depending on the Vite/TS config the mixed-extension import may work, but adding a profiling.js re-export or converting index.js to index.ts would keep the barrel consistent.

**Conventions**

The PR adds three read-only profiling endpoints (dataset summary, per-column profile, Pearson correlation), a tightly scoped service layer, response schemas, and solid unit plus integration test coverage. The most important correctness issue is that boolean columns are placed in categorical_columns in the summary response while the column-profile endpoint gives them a boolean-specific shape, creating a mismatch any UI consumer will hit immediately. Datetime columns silently disappear from both type lists in the summary. The slash-in-column-name routing edge case and the binary-vs-zero-inflated priority ordering are minor but worth addressing before this ships.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/api/profiling.ts:1050` The frontend client uses encodeURIComponent(columnName) to build the column-profile URL, which encodes a slash as %2F. The backend route uses the :path converter and expects a literal slash (as exercised by the endpoint test at line 926 which sends a raw '/'. Whether a given ASGI/HTTP server normalises %2F back to / before path matching is implementation-defined; uvicorn strips the percent-encoding by default so the two approaches happen to agree in local dev, but any reverse proxy that blocks or preserves encoded slashes (nginx with merge_slashes on, common in production) will route %2F differently from a literal slash, causing 404s for columns whose names contain a slash.
- **[minor/quality]** `dataloom-backend/app/services/profiling_service.py:283` dataset_summary iterates over all columns twice: once to build dtype_counts and the column lists, and again inside the generator expression that computes total_missing_cells (each call to _coerce_sentinels re-scans the entire column). For wide datasets this doubles the read work. A single pass that accumulates both the dtype metadata and the per-column null counts would be cleaner and more efficient.
- **[minor/quality]** `dataloom-frontend/src/api/index.js:947` The barrel re-export in index.js imports from profiling.ts (a TypeScript module) while index.js itself is plain JavaScript. Vite resolves this correctly in the current setup, but it leaves a type-unsafe boundary: callers who import from index.js receive no TypeScript inference for the three profiling functions. Exporting from an index.ts barrel, or at minimum a profiling.js wrapper, would be consistent with how the other modules are handled.
- **[nit/quality]** `dataloom-backend/app/services/profiling_service.py:349` detect_distribution guards 'if non_null' on line 349 as a division guard before unique_count / non_null, but non_null is already guaranteed positive by the 'if non_null <= 0: return unknown' check on line 341. The redundant guard is harmless but adds noise.

**Conventions**

The PR adds a well-structured data profiling layer: a pure-function service with type-aware column statistics and JSON-safe serialisation, three read-only FastAPI endpoints, Pydantic response schemas, and a typed TypeScript API client backed by solid unit and endpoint tests. The primary concern to address before merging is the column-name-with-slash routing: the frontend encodes slashes via encodeURIComponent while the backend relies on the :path converter receiving a literal slash, and behaviour diverges under reverse proxies that preserve %2F. Consider standardising on one approach (either always encode on the client and decode explicitly on the server, or document that column names with slashes require a literal-slash URL) and adding a corresponding note in the test. The two minor quality issues (double-scan in dataset_summary, mixed JS/TS barrel) are worth fixing in a follow-up but are not blockers.

PR #366, opened 2026-06-12
feat: add user profile page with email and password management

SHIVAM-KUMAR-59:
## Description

Adds a user profile page that allows authenticated users to view and manage their account information.

### Features Added

- Added a dedicated `/profile` page
- Display authenticated user's email and account creation date
- Added inline email editing with save/cancel actions
- Added password change functionality with current password verification
- Added loading states for email updates and password changes
- Added success/error toast notifications for profile actions
- Added profile navigation from the navbar
- Synced AuthContext after profile updates to keep global user state consistent

### Backend Changes

- Added authenticated email update endpoint
- Added authenticated password change endpoint
- Added service-layer logic for updating email and changing passwords

### Testing

- Added backend test coverage for:

  - Email update success and failure scenarios
  - Password change success and failure scenarios
  - Authentication and validation checks

Fixes #365 

## Type of Change

- [x] New feature

## How Has This Been Tested?

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

**Following scenarios were verified manually:**

- Verified profile page loads authenticated user data
- Verified email edit/save/cancel flow
- Verified password change flow
- Verified loading states during API requests
- Verified success and error toast notifications
- Verified navbar profile navigation
- Verified AuthContext updates immediately after email changes

## Screen Recording

https://github.com/user-attachments/assets/a6bbccc7-5e5e-4cd8-823f-aa75c0a4f711

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review
- [x] I have added/updated documentation as needed
- [x] My changes generate no new warnings
- [x] Tests pass locally

PR #373, opened 2026-06-15
feat: add account deletion option to user profile page

### LINKED ISSUE #372: [Feature] Add account deletion option to user profile page
The user profile page (added in #366) allows updating email and password, but there is no way for a user to delete their account. Users who want to stop using DataLoom have no self-service option to remove their account and associated data.

Add a "Delete Account" button to the profile page, styled to indicate a destructive action (danger variant). Clicking it should open a confirmation modal warning that this action is permanent and will delete all projects, checkpoints, and logs associated with the account. On confirmation, the backend should delete the user record and cascade-delete all owned projects (which already cascade-deletes logs and checkpoints per the `Project` model relationships) and password reset tokens. After deletion, the user should be logged out and redirected to the sign-in page.

Soft-delete (deactivate account) — rejected for now as it adds complexity (reactivation flow, data retention policy) beyond the current scope. Hard delete is simpler and matches user expectations for "delete my account."

**Additional context**
- Backend: new `DELETE /auth/me` or similar endpoint requiring password confirmation
- Frontend: confirmation modal using existing `Modal` component, requires password re-entry before submitting
- After deletion, clear auth cookie/session and redirect to sign-in

SHIVAM-KUMAR-59:
## Description
The user profile page (added in #366) allowed updating email and password but had no way for users to delete their account. This PR adds a "Delete account" section to the profile page with a confirmation modal requiring password re-entry, and a backend `DELETE /auth/me` endpoint that removes the user along with all owned projects, their logs, checkpoints, and files.

Fixes #372 

## Type of Change
- [x] New feature

## How Has This Been Tested?
- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

**Following scenarios were tested manually:**
- Created a user with multiple projects, checkpoints, and logs; deleted the account and verified all DB records and project files are removed
- Attempted deletion with incorrect password — returns 400 with appropriate error
- After deletion, auth cookie is cleared and user is redirected to sign-in

## Screen Recording

https://github.com/user-attachments/assets/89d2fdaa-dcad-415c-bed3-8711b6143900

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/alembic/versions/c2f56609809e_add_cascade_delete_to_password_reset_.py
+++ b/dataloom-backend/alembic/versions/c2f56609809e_add_cascade_delete_to_password_reset_.py
@@ -1,0 +1,42 @@
+"""add cascade delete to password reset tokens
+
+Revision ID: c2f56609809e
+Revises: 3185ec76af85
+Create Date: 2026-06-15 17:38:22.422079
+
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "c2f56609809e"
+down_revision: str | Sequence[str] | None = "3185ec76af85"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+FK_NAME = "password_reset_tokens_user_id_fkey"
+
+
+def upgrade() -> None:
+    op.drop_constraint(FK_NAME, "password_reset_tokens", type_="foreignkey")
+    op.create_foreign_key(
+        FK_NAME,
+        "password_reset_tokens",
+        "users",
+        ["user_id"],
+        ["id"],
+        ondelete="CASCADE",
+    )
+
+
+def downgrade() -> None:
+    op.drop_constraint(FK_NAME, "password_reset_tokens", type_="foreignkey")
+    op.create_foreign_key(
+        FK_NAME,
+        "password_reset_tokens",
+        "users",
+        ["user_id"],
+        ["id"],
+    )
--- a/dataloom-backend/app/api/endpoints/auth.py
+++ b/dataloom-backend/app/api/endpoints/auth.py
@@ -1,6 +1,7 @@
 """Authentication API endpoints: signup, signin, logout, and current user."""
 
-from fastapi import APIRouter, Depends, HTTPException, Response
+from fastapi import APIRouter, Body, Depends, HTTPException, Response
+from sqlalchemy.exc import SQLAlchemyError
 from sqlmodel import Session
 
 from app import database, models, schemas
@@ -134,3 +135,32 @@ def change_password(
         raise HTTPException(status_code=400, detail=str(e)) from e
 
     return {"message": "Password changed successfully"}
+
+
+@router.delete("/me")
+def delete_account(
+    response: Response,
+    payload: schemas.DeleteAccountRequest = Body(...),
+    current_user: models.User = Depends(get_current_user),
+    db: Session = Depends(database.get_db),
+):
+    """Delete the currently authenticated user's account and all associated data."""
+    try:
+        auth_service.delete_user_account(db, current_user, payload.password)
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e)) from e
+    except SQLAlchemyError as e:
+        raise HTTPException(
+            status_code=500,
+            detail="Failed to delete account. Please try again later.",
+        ) from e
+
+    settings = get_settings()
+    response.delete_cookie(
+        key=ACCESS_TOKEN_COOKIE,
+        path="/",
+        httponly=True,
+        secure=settings.cookie_secure,
+        samesite="lax",
+    )
+    return {"message": "Account deleted successfully"}
--- a/dataloom-backend/app/models.py
+++ b/dataloom-backend/app/models.py
@@ -41,7 +41,9 @@ class PasswordResetToken(SQLModel, table=True):
     __tablename__ = "password_reset_tokens"
 
     id: int | None = Field(default=None, primary_key=True)
-    user_id: uuid_mod.UUID = Field(sa_column=Column(sa.Uuid, sa.ForeignKey("users.id"), nullable=False))
+    user_id: uuid_mod.UUID = Field(
+        sa_column=Column(sa.Uuid, sa.ForeignKey("users.id", ondelete="CASCADE"), nullable=False)
+    )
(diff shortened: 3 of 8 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/auth_service.py:204` The deletion in delete_user_account is not atomic. delete_project() calls db.commit() after each project inside the loop (project_service.py line 102), so some projects are durably deleted before db.delete(user) + db.commit() runs. If any step fails mid-loop or after the loop, the DB is left in an inconsistent state: some projects deleted, user still present (or vice versa). The whole operation should be wrapped in a single transaction with no intermediate commits.
- **[major/quality]** `dataloom-backend/alembic/versions/c2f56609809e_add_cascade_delete_to_password_reset_.py:22` The Alembic migration bundles three index operations unrelated to the cascade-delete feature: it drops ix_checkpoints_id on checkpoints, drops ix_datasets_dataset_id on projects, and creates ix_projects_name. These appear to be schema drift that was auto-generated alongside the intentional change. Bundling unrelated schema mutations in a feature migration makes rollback harder and obscures intent -- the drift should be separated into its own migration or resolved before merge.
- **[minor/quality]** `dataloom-frontend/src/pages/ProfilePage.jsx:412` The Change Password submit button is converted to a <Button> component but the className override only carries layout classes (mt-6 flex w-full items-center justify-between rounded-lg px-4 py-3), dropping the accent-color styling. The Button default variant applies bg-blue-500 instead of the project's bg-accent, so the button color no longer matches the rest of the form. Pass variant or explicitly set the accent color to keep visual consistency.
- **[minor/quality]** `dataloom-backend/app/api/endpoints/auth.py:137` The DELETE /auth/me endpoint accepts a JSON body (payload: schemas.DeleteAccountRequest) on a DELETE request. The HTTP spec discourages request bodies on DELETE, and some HTTP proxies and clients silently strip them. A safer alternative is to require the password via a POST to /auth/me/deletion-request or to use a custom request header. The Axios client.delete({ data }) pattern works today but is fragile.

**Conventions**

This PR adds account deletion end-to-end: a password-confirmed modal on the profile page, a DELETE /auth/me endpoint, cascading cleanup of projects and files, and a migration adding ON DELETE CASCADE to password_reset_tokens. The implementation is well-structured and includes tests covering success, wrong password, and unauthenticated access. Two items need attention before merge: the delete_user_account service performs intermediate db.commit() calls inside the per-project loop (via delete_project), making the overall deletion non-atomic and leaving the database in an inconsistent state if any step fails after partial commits. The Alembic migration also bundles unrelated index changes (dropping ix_checkpoints_id and ix_datasets_dataset_id, adding ix_projects_name) that appear to be Alembic drift and should either be split into a separate migration or confirmed as intentional."}

SHIVAM-KUMAR-59:
Addressed three of the four findings:

- **[major/bug] non-atomic deletion** — fixed. `delete_user_account` now performs all DB deletions (logs, checkpoints, projects, user) in a single transaction with no intermediate commits, wrapped in try/except with rollback on failure.
- **[major/quality] migration bundling unrelated index/FK changes** — fixed. Removed the unrelated `ix_checkpoints_id`, `ix_datasets_dataset_id`, `ix_projects_name`, and `user_logs_checkpoint_id_fkey` changes from the migration — these were autogenerate drift unrelated to the cascade-delete feature. The migration now only contains the `password_reset_tokens` cascade FK change.
- **[minor/quality] Button color regression** — fixed. Restored `bg-accent`/`hover:bg-accent-hover` classes on the change password button.

On the **DELETE** with **JSON body** finding — rather than switching to `POST /me/delete`, would you prefer we drop the password requirement entirely and rely on the authenticated user's ID (from the JWT) for account deletion instead? Since the endpoint already requires authentication via `get_current_user`, the password confirmation may be redundant, and removing it would let us keep the semantically correct `DELETE /auth/me` with no body. Curious which approach you'd prefer ? happy to implement either. @ivantha

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/alembic/versions/c2f56609809e_add_cascade_delete_to_password_reset_.py:28` The downgrade() function drops the recreated foreign key using None as the constraint name (`op.drop_constraint(None, ...)`). Alembic will generate a name at migration time, making the downgrade non-deterministic and likely to fail in practice.
- **[major/security]** `dataloom-backend/app/api/endpoints/auth.py:137` DELETE /auth/me accepts a request body, but FastAPI routes decorated with @router.delete do not parse a body by default via a plain Pydantic model parameter — the body parameter requires an explicit Body() annotation or the payload will be silently ignored, making the password confirmation bypassable. This should be verified; if the password check is indeed skipped, any authenticated user can delete any account without knowing the password. _(unverified)_
- **[minor/quality]** `dataloom-frontend/src/pages/ProfilePage.jsx:172` After a successful deletion the function returns without resetting deletingAccount to false (the navigate() call makes it moot, but if navigation is ever delayed or blocked the button stays in the Deleting... state with no recovery path for the user).
- **[minor/quality]** `dataloom-frontend/src/pages/ProfilePage.jsx:451` The password input inside the delete modal has no onKeyDown handler for Enter, so pressing Enter does not confirm deletion while pressing Enter on the change-password form does submit it — inconsistent keyboard UX.
- **[nit/quality]** `dataloom-backend/app/services/auth_service.py:220` The SQLAlchemyError catch block re-raises a bare SQLAlchemyError rather than re-wrapping it as a ValueError or a custom exception, so the endpoint will surface a 500 with raw SQLAlchemy detail instead of a user-readable 400. Consider translating it to an HTTPException in the endpoint or a domain exception in the service.

**Conventions**

The PR adds a well-structured account deletion flow: a password-confirmed modal on the frontend and a `DELETE /auth/me` endpoint on the backend that removes the user and all owned projects atomically before cleaning up files. The most important issue to resolve before merge is the FastAPI body-parsing question on the DELETE endpoint: FastAPI does not automatically parse a request body from a plain model parameter on DELETE routes, so the password field may never be read and the confirmation check may be silently bypassed. Additionally, the Alembic downgrade drops a constraint by passing `None` as the name, which will fail at runtime. Both of those should be fixed; the remaining findings are minor quality nits.

SHIVAM-KUMAR-59:
Addressed the feedback:

- Added user-friendly handling for `SQLAlchemyError` in the endpoint instead of surfacing raw database exceptions.
- Ensured the account deletion loading state is properly reset if the operation fails.
- Added Enter-key submission support to the delete account confirmation modal for consistent keyboard UX.

Ready for re-review @ivantha

PR #376, opened 2026-06-16
feat: allow users to rename a project

### LINKED ISSUE #375: [Feature] Allow users to rename a project
Once a project is created, there is no way to rename it. Users who want to update a project's name (e.g. after realizing a typo, or wanting a more descriptive name) have no option to do so within the app.

Add the ability to rename a project. This could be inline editing (click the project name to edit it) or a dedicated rename option in the project menu. The backend should expose an endpoint to update the project name, with validation to prevent empty names.

Requiring users to delete and re-upload the project with a new name — rejected as overly destructive and loses all checkpoint/log history for a simple naming fix.

**Additional context**
- Backend: new endpoint, likely `PATCH /projects/{project_id}/rename` or similar
- Frontend: inline edit UI, similar pattern to email editing in the profile page (pencil icon, input field, save/cancel)
- Should validate name is non-empty and trimmed

SHIVAM-KUMAR-59:
## Description
There was previously no way to rename a project after creation. This PR adds the ability to rename a project via an inline edit UI in the navbar and a new backend endpoint to persist the change.

Fixes #375 

## Type of Change
- [x] New feature

## How Has This Been Tested?
- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

**Following scenarios were modified manually:**
- Renamed a project — name updates and persists on refresh
- Attempted to save an empty/whitespace-only name — rejected with validation error
- Cancel button discards changes without persisting

## Screen Recording

https://github.com/user-attachments/assets/777d68a2-09ac-448b-8595-d4bee844c2b0

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -25,6 +25,7 @@ from app.services.project_service import (
     delete_project,
     get_last_change_log,
     get_recent_projects,
+    rename_project,
 )
 from app.services.transformation_service import apply_logged_transformation
 from app.utils.file_formats import TableWriteOptions, get_format, get_format_for_extension
@@ -131,6 +132,33 @@ def recent_projects(
         )
         for p in projects
     ]
+
+
+@router.patch("/{project_id}/rename", response_model=schemas.ProjectResponse)
+async def rename_project_endpoint(
+    payload: schemas.RenameProjectRequest,
+    db: Session = Depends(database.get_db),
+    project: models.Project = Depends(get_project_or_404),
+):
+    """Rename a project."""
+    try:
+        updated_project = rename_project(db, project, payload.name)
+    except ValueError as e:
+        raise HTTPException(status_code=422, detail=str(e)) from e
+
+    df = read_table_safe(updated_project.file_path)
+    total_rows = len(df)
+    resp = dataframe_to_response(df)
+    return {
+        "filename": updated_project.name,
+        "file_path": updated_project.file_path,
+        "project_id": updated_project.project_id,
+        "page": 1,
+        "page_size": total_rows,
+        "total_rows": total_rows,
+        "total_pages": 1,
+        **resp,
+    }
 
 
 @router.post("/{project_id}/save", response_model=schemas.ProjectResponse)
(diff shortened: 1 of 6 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/api/endpoints/projects.py:148` The rename endpoint reads the CSV file and constructs a full ProjectResponse (with rows, page info, etc.) after a rename, but the client only uses the updated name from the response — this unnecessary file read will fail with an unhandled exception if the file has been deleted or moved, silently breaking what should be a lightweight metadata-only operation. The endpoint should return only the project metadata (or at least guard the file read).
- **[minor/quality]** `dataloom-backend/app/schemas.py:439` RenameProjectRequest.name has no length validator. A caller can send a 10 MB string and it will pass the non-empty check and be committed to the database. Adding max_length (e.g. 255) on the field prevents that.
- **[minor/quality]** `dataloom-backend/tests/test_rename_project.py:38` The test for the empty-name and whitespace-only rejection uses a bare try/except with a manual AssertionError instead of pytest.raises. This is non-idiomatic for the test suite and hides the actual exception type if something other than ValueError is raised.
- **[nit/style]** `dataloom-frontend/src/Components/Navbar.jsx:42` handleCancelEdit resets editedName to projectName, which is already done by handleStartEdit on the next open. Resetting it on cancel is harmless but unnecessary since isEditingName being false means the value is not displayed.

**Conventions**

The PR cleanly adds project renaming across the full stack: a PATCH endpoint, a service function with whitespace trimming and empty-name validation, a well-structured inline edit UI in the Navbar, and a reasonable test file. The main concern worth fixing before merge is that the rename endpoint unnecessarily reads the CSV file to build a ProjectResponse — if the file is missing this will throw an unhandled 500, and even in the happy path it adds latency for a metadata-only operation; the endpoint should return a lightweight response or at least guard the read. A max-length constraint on RenameProjectRequest.name is also worth adding to prevent oversized input reaching the database. Everything else (hygiene, conventions, test coverage) is solid.

SHIVAM-KUMAR-59:
Addressed all the points:

- Added a maximum length validation (255 characters) for `RenameProjectRequest.name`.
- Updated the rename endpoint to return a lightweight rename response instead of constructing a full `ProjectResponse`, removing the unnecessary CSV read and avoiding failures when the underlying file is unavailable.
- Refactored the validation test to use `pytest.raises` instead of a manual `try/except`.

Ready for re-review @ivantha

PR #379, opened 2026-06-17
fix: prevent logs panel overflow and refresh logs after transforms

### LINKED ISSUE #378: [Bug] Logs Panel Expands Indefinitely and Does Not Refresh After Transformations
**Describe the bug**

The Logs panel has two usability issues:

1. It has no maximum height, so with many log entries the panel grows indefinitely and pushes the main data table below the viewport.
2. It does not update in real time after transformations are applied. New log entries only appear after a manual page refresh.

**To Reproduce**

1. Open a project and keep the Logs panel visible.
2. Apply multiple transformations (Sort, Filter, Rename Column, etc.).
3. Observe that:

   * New log entries do not appear immediately.
   * As more logs accumulate, the Logs panel continues expanding vertically.
4. Refresh the page.
5. The missing log entries now appear, and the enlarged Logs panel may push the table off-screen.

**Expected behavior**

* The Logs panel should update immediately after a successful transformation without requiring a page refresh.
* The Logs panel should have a fixed maximum height with internal scrolling, similar to the Checkpoints panel, so the main data table remains visible.

**Screenshots**
- Logs panel expanding indefinitely causing data table to go off-screen

Image

- Stale logs until refresh

https://github.com/user-attachments/assets/1c022e5c-a0c6-46f5-9982-f0a868f0ea94

**Environment**

* OS: Any
* Browser: Any
* Node.js version: Any
* Python version: Any

**Additional context**

The Checkpoints panel already implements the desired scrolling behavior using a pattern similar to:

```css
max-h-72 overflow-y-auto
```

A similar approach can be applied to the Logs panel.

SHIVAM-KUMAR-59:
## Description

Fixes the Logs panel UX issues by:

- Adding a fixed maximum height with internal scrolling to prevent the panel from pushing the data table out of view when many log entries exist.
- Refreshing logs automatically after successful transformations so newly created log entries appear immediately without requiring a page refresh.

Instead of manually calling `fetchLogs()` in every transformation form, a centralized event-driven approach was implemented. The `transformProject()` API helper now emits a custom browser event after successful transformations, and the Logs panel listens for this event to refresh itself.

This approach was chosen because:

- It avoids duplicating refresh logic across every transformation form.
- Future transformation forms automatically inherit log-refresh behavior.
- It reduces maintenance overhead and prevents regressions when new transformations are added.
- It keeps logging concerns centralized rather than coupling them to individual UI components.

Fixes #378 

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- [x] Existing tests pass
- [x] Manual testing

**Following scenarios were verified manually:**
- Logs panel remains scrollable when a large number of logs are present.
- New log entries appear immediately after successful transformations.
- Existing transformation workflows continue to function correctly.
- No page refresh is required for log updates.

## Screen Recording

https://github.com/user-attachments/assets/012285fd-86b3-4395-9712-abe0e083bd68

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/MenuNavbar.jsx
+++ b/dataloom-frontend/src/Components/MenuNavbar.jsx
@@ -100,6 +100,18 @@ const MenuNavbar = ({ projectId }) => {
     if (showCheckpoints) fetchCheckpoints();
   }, [showLogs, showCheckpoints, fetchLogs, fetchCheckpoints]);
 
+  useEffect(() => {
+    const handleLogsRefresh = () => {
+      if (showLogs) fetchLogs();
+    };
+
+    window.addEventListener("project:logs-refresh", handleLogsRefresh);
+
+    return () => {
+      window.removeEventListener("project:logs-refresh", handleLogsRefresh);
+    };
+  }, [showLogs, fetchLogs]);
+
   const handleSave = () => {
     setIsInputOpen(true);
   };
--- a/dataloom-frontend/src/Components/history/LogsPanel.jsx
+++ b/dataloom-frontend/src/Components/history/LogsPanel.jsx
@@ -22,7 +22,7 @@ const LogsPanel = ({ logs, onClose }) => {
         </button>
       </div>
 
-      <div className="mt-4 overflow-x-auto">
+      <div className="mt-4 overflow-x-auto overflow-y-auto max-h-72">
         <table className="min-w-full bg-white rounded-lg overflow-hidden">
           <thead className="bg-gray-50">
             <tr>
--- a/dataloom-frontend/src/api/transforms.js
+++ b/dataloom-frontend/src/api/transforms.js
@@ -12,6 +12,7 @@ import client from "./client";
  */
 export const transformProject = async (projectId, transformationInput) => {
   const response = await client.post(`/projects/${projectId}/transform`, transformationInput);
+  window.dispatchEvent(new CustomEvent("project:logs-refresh"));
   return response.data;
 };
 


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/history/LogsPanel.jsx:25` max-h-72 is added to a div that only has overflow-x-auto, not overflow-y-auto. Rows beyond 18rem (288px) will be silently clipped rather than scrolled. The fix requires adding overflow-y-auto (or replacing overflow-x-auto with overflow-auto) on the same element.

**Conventions**

The event-driven refresh approach in MenuNavbar.jsx and transforms.js is well-structured and the cleanup in the useEffect is correct. However, the overflow fix in LogsPanel.jsx has a functional bug: adding max-h-72 without overflow-y-auto means rows beyond 288px are clipped with no scrollbar, which is worse than the original behavior. Adding overflow-y-auto (or overflow-auto) to that same div is the required one-line fix before merging.

SHIVAM-KUMAR-59:
Addressed the issue:
- **[major/bug] `dataloom-frontend/src/Components/history/LogsPanel.jsx:25`:** Added missed `overflow-y-auto` to the div, allowing rows beyond 288px to be accessed via scrollbar

Ready for re-review @ivantha

ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `dataloom-frontend/src/api/transforms.js:15` The custom event 'project:logs-refresh' is dispatched unconditionally inside transformProject(), meaning it fires even when the API call throws and the caller's catch block handles the error. If the promise rejects, the line after client.post() is not reached, so in practice this is safe — but the dependency on execution order is implicit. Moving the dispatch to the call sites after the awaited call resolves successfully (or returning the event emission as part of a resolved value) would make the intent explicit and easier to test.
- **[minor/quality]** `dataloom-frontend/src/Components/history/LogsPanel.jsx:25` max-h-72 (18rem / 288px) is a hardcoded magic number that may feel cramped on small viewports and excessively restrictive on tall screens. A responsive constraint (e.g. max-h-[40vh]) or a CSS variable would make the limit easier to tune without hunting for a Tailwind utility class.
- **[nit/test]** `dataloom-frontend/src/api/transforms.js:15` No automated test covers the new event listener or the dispatch in transformProject(). A unit test asserting that a 'project:logs-refresh' CustomEvent is dispatched after a successful transform call, and a component test verifying that fetchLogs is called when the event fires while showLogs is true, would lock in this behavior against future refactors.

**Conventions**

This PR fixes two related UX issues: the LogsPanel no longer expands unboundedly (overflow + max-h-72 cap), and logs now refresh automatically after a transform via a lightweight custom browser event rather than prop-drilling fetchLogs into every form. The event-driven approach is a reasonable design choice and the description explains the rationale clearly. The main concern worth addressing before merge is adding at least a minimal unit test for the new event dispatch and listener to prevent silent regressions; the hardcoded max-h-72 is a secondary style note worth considering but not blocking.

PR #382, opened 2026-06-18
feat: add searchable, type-aware column selectors

### LINKED ISSUE #380: [Feature] Type-aware column selectors
Column selection currently relies on native HTML dropdowns. These become difficult to use when datasets contain many columns because users cannot search the list or see each column’s data type.

Create a reusable combobox inspired by shadcn/ui to replace native column dropdowns. It should searching, keyboard navigation, data-type badges, and both single-select and multi-select use cases.

We could directly integrate shacdn/ui's combobox itself, but that will introduce a new library which is not needed at the moment.

**Additional context**
The new selector should be used consistently across transformation forms such as filtering, sorting, grouping, pivot tables, filling empty values, and removing duplicates. It should preserve required-field validation and existing form behavior.

hanzalahwaheed:
## Description
- replace native column dropdowns with a searchable combobox and a multi-select picker that show each column's data type
- animate the combobox open/close with tailwindcss-animate

Fixes #380 

## Type of Change

- [ ] Bug fix
- [X] New feature
- [X] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
- [X] Existing tests pass
- [X] New tests added
- [X] Manual testing

## Screenshots
image

## Checklist

- [X] Tests pass locally


ivantha:
### Automated review summary

**Findings**
- **[major/quality]** `dataloom-frontend/src/Components/common/ColumnSelect.tsx:241` useProjectContext() is double-cast with `as unknown as { columns: string[]; dtypes: Record<string, string> }` to bypass TypeScript. This suppresses all type checking on the context value. The context type should be updated to include dtypes, or the hook's return type should be widened rather than escaping the type system entirely.
- **[major/quality]** `dataloom-frontend/src/Components/common/ColumnMultiSelect.tsx:90` Same `as unknown as` double-cast applied to useProjectContext() in ColumnMultiSelect. Both components should share the fix once the context type is corrected.
- **[major/bug]** `dataloom-frontend/src/Components/forms/TrimWhitespaceForm.jsx:1046` TrimWhitespaceForm passes `options={['All string columns', ...columns]}` to ColumnSelect, injecting a synthetic string as a real column name. ColumnSelect will call dtypes?.['All string columns'] and get undefined, rendering a blank DtypeBadge. More importantly, the backend will receive 'All string columns' as a literal column name, which differs from the previous behavior where the empty string was the sentinel. The backend contract for this value needs to be verified and documented, and the synthetic option should not be mixed into the column list. _(unverified)_
- **[minor/quality]** `dataloom-frontend/src/Components/forms/StringReplaceForm.jsx:982` StringReplaceForm uses showToast for the empty-column guard while every other form uses setError/FormErrorAlert for the same guard. The inconsistency means the error appears in a different location in the UI for this one form.
- **[minor/quality]** `dataloom-frontend/src/Components/common/ColumnSelect.tsx:360` The DtypeBadge displayed inline inside the trigger button's <span> (next to the selected value text) is not separated from the text by a space or flex gap, so the badge renders flush against the column name string.
- **[nit/test]** `dataloom-frontend/src/Components/common/__tests__/ColumnSelect.test.tsx:469` ColumnSelect.test.tsx has no coverage for keyboard navigation (ArrowDown/ArrowUp/Enter), which is a notable interactive behavior added by this PR.

**Conventions**

This PR replaces all native column dropdowns and free-text column inputs across the transform forms with new searchable, dtype-annotated ColumnSelect and ColumnMultiSelect components, and ships matching unit and E2E test updates. The two biggest issues to resolve before merging are: (1) both new components escape TypeScript's type system with a double `as unknown as` cast that should instead be fixed by adding dtypes to the ProjectContext type, and (2) TrimWhitespaceForm injects a synthetic "All string columns" string directly into the column options list, which will produce a broken dtype badge and may send an unexpected value to the backend. The inconsistent error-reporting pattern in StringReplaceForm (toast vs. FormErrorAlert) is a smaller quality gap worth aligning. Conventions and commit hygiene are clean.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/forms/TrimWhitespaceForm.jsx:86` TrimWhitespaceForm passes 'All string columns' as a regular entry in the options array passed to ColumnSelect. ColumnSelect treats it as a real column name — it will appear in the popover with no dtype badge and will be submitted as-is to the API. The previous code used a dedicated <option value="All string columns"> with no dtype lookup, so the intent is preserved, but the new implementation also means the component's required-field guard (checking !column) will pass for this synthetic value, skipping the 'Please select a column.' error path even though no real column was chosen. Consider either keeping it as an includeEmptyOption-style prop or validating that column !== 'All string columns' before submission.
- **[minor/quality]** `dataloom-frontend/src/Components/common/ColumnSelect.tsx:354` ColumnSelect renders the DtypeBadge inside the trigger button's display span, inline with the placeholder/value text (line 355), but the button's layout is flex with gap-2 only between the span and the chevron SVG. The badge sits flush against the column name text with only its own ml-1.5 margin. When a value is selected and the badge renders, the span truncates the combined text+badge together with CSS truncate, which can clip the badge on narrow forms. Consider rendering the badge as a flex sibling to the text span rather than inside it.
- **[minor/quality]** `dataloom-frontend/src/Components/common/ColumnSelect.tsx:278` In ColumnSelect, the outside-click handler attaches to document via mousedown on every open state change. The effect cleanup removes it correctly, but the handler captures containerRef.current at call time via a closure. If containerRef.current changes between the time the listener fires and when the closure executes (unlikely in practice but possible during concurrent renders), the stale ref could fail to identify the container. Using a callback ref or reading the ref inside the effect setup would be more defensive.
- **[nit/quality]** `dataloom-frontend/src/Components/common/ColumnMultiSelect.tsx:106` ColumnMultiSelect's search input does not have an aria-label or associated <label> element, making it inaccessible to screen readers. The placeholder text 'Search columns...' is not a substitute for an accessible label.
- **[nit/quality]** `dataloom-frontend/src/Components/common/__tests__/ColumnSelect.test.tsx:505` ColumnSelect.test.tsx line 505 asserts options[0] has text content 'Amount' after filtering with 'am', but 'Amount'.toLowerCase() = 'amount' contains 'am', and 'Date' does not, so only 'Amount' matches. The assertion is correct but would silently pass even if 'datetime' (which also contains 'am' in 'Amount') matched. The test comment should clarify. Low risk but worth noting.

**Conventions**

This PR replaces all native column dropdowns and free-text column inputs across the transform forms with two new reusable components — a searchable, dtype-aware single selector (ColumnSelect) and a searchable multi-select picker (ColumnMultiSelect) — and updates the E2E and unit tests to match. The implementation is well-structured with good keyboard navigation, animation handling, and test coverage. The one substantive issue to address before merge is in TrimWhitespaceForm, where 'All string columns' is injected as a plain entry in the ColumnSelect options array, which means the required-field guard will silently pass for it; this should be modelled as an includeEmptyOption-style prop or validated explicitly. A minor inline layout issue with the dtype badge clipping inside the trigger span is also worth fixing before merge.

Issue #374, opened 2026-06-16
[Bug] Bug Report & Enhancement Proposal: Sample Tool — Destructive Mutation & UX Issues


YADUNANDAN-SINGH:
### Summary

The current **Sample** tool in Dataloom contains a critical bug: it **destructively mutates the working copy CSV** when a user applies a sample. This makes it impossible to increase the sample size after an initial reduction, and provides no way to recover the original data without using Undo or reverting to a checkpoint. This issue proposes a **preview-before-persist** architecture to resolve the problem cleanly.

---

### Bug Description

**Steps to Reproduce**

1. Upload a dataset with **300 rows**.
2. Open the **Sample** tool and set the sample size to `100`.
3. Apply the sample. The dataset is now truncated to **100 rows**.
4. Attempt to increase the sample size to `120`.

**Expected Behavior**

The app should dynamically re-sample from the **original 300-row dataset**, returning 120 rows without any error.

**Actual Behavior**

The app throws an error (or silently fails) because the underlying dataset has already been **permanently reduced to 100 rows**. Requesting a sample of 120 from a 100-row dataset is impossible.

**Root Cause**

The sample operation overwrites the backend's working copy CSV (`_copy.csv`) in place via the `/transform` endpoint. There is no separation between **previewing** a transformation result and **persisting** it, so any sampling action is irreversible without using Undo (which replays the entire log from the original file).

---

### ✅ Implemented Fix — Preview-Before-Persist

The fix introduces a **preview mode** across all destructive (row-reducing) transform forms. The key idea: clicking "Apply" only shows a **frontend-only preview** — the backend computes the result but does **not** save it to disk or log it. The user must explicitly click **"Save Changes"** to persist.

#### Backend Changes

* **Added `preview` query parameter** to `POST /projects/{id}/transform`:
  * When `preview=true`, the endpoint applies the transform on the DataFrame and returns the result **without calling `save_table_safe()` or `log_transformation()`**.
  * When `preview=false` (default), the existing behavior is preserved — the transform is saved to the working copy and logged.
  * This is a single boolean gate on the existing persistence logic — no new endpoints, models, or services required.

#### Frontend Changes — ProjectContext

* **Added preview state management** to `ProjectContext`:
  * `isPreviewMode` (boolean) — whether the table is currently showing unsaved preview data.
  * `previewSnapshot` — stores the original `columns`, `rows`, `dtypes`, and pagination state before the preview, so Cancel can restore them.
[...]

PR #383, opened 2026-06-19
fix: add preview-before-persist flow for row-reducing transforms

YADUNANDAN-SINGH:
## Fix: add preview-before-persist flow for row-reducing transforms

#374 

---

### Problem

Row-reducing transforms (Sample, Filter, Drop Duplicates) **permanently overwrite** the working copy CSV on every apply. This creates an unrecoverable state:

1. User uploads a 300-row dataset
2. Samples to 100 rows → `_copy.csv` is overwritten with 100 rows
3. Attempts to re-sample to 120 → **fails silently** — the backend only has 100 rows left

There was no separation between *previewing* a transform result and *persisting* it.

---

### Solution

Introduced a **preview-before-persist** flow: clicking "Apply" now computes the transform result on the backend but **does not save it to disk or log it**. The user sees a live preview in the main table and must explicitly click **"Save Changes"** to commit, or **"Cancel"** to discard.

#### How it works

```
Apply (preview=true)         Save Changes (preview=false)
        │                              │
        ▼                              ▼
  Backend computes             Backend computes
  transform on df              transform on df
        │                              │
        ▼                              ▼
  Returns result               Saves to _copy.csv
  (NO disk write)              Logs to user_logs
  (NO log entry)                       │
        │                              ▼
        ▼                        Data persisted ✅
  Frontend shows in
  main table (preview)
```

---

#### Backend

| File | Change |
|------|--------|
| `app/api/endpoints/transformations.py` | Added `preview: bool = Query(False)` parameter. When `preview=True`, the endpoint skips `save_table_safe()` and `log_transformation()` but still returns the computed result. |

> **Zero breaking changes** — the default is `preview=False`, so all existing callers (including the replay/undo path) behave identically.

#### Frontend

| File | Change |
|------|--------|
| `src/api/transforms.js` | Added optional `{ preview }` parameter to `transformProject()`. Appends `?preview=true` query param when set. |
| `src/context/ProjectContext.jsx` | Added preview state management: `isPreviewMode`, `previewSnapshot`, `pendingTransform`, `enterPreviewMode()`, `cancelPreview()`, `confirmPreview()`. |
| `src/Components/forms/FilterForm.jsx` | Apply now sends `preview: true`. Added "Save Changes" button (visible only in preview mode). Cancel restores original data. |
| `src/Components/forms/SampleRowsForm.jsx` | Same preview-then-save pattern as FilterForm. |
| `src/Components/forms/DropDuplicateForm.jsx` | Same preview-then-save pattern as FilterForm. |
| `src/Components/DataScreen.jsx` | Removed `PreviewActionBar` render (save/cancel now lives inside each form). |
| `src/Components/Table.jsx` | Removed `PreviewActionBar` render from the pagination footer. |

---

### How to Test

**1. Core bug scenario (Sample)**

- Upload a CSV with 300+ rows
- Open **Sample** → enter `100` → click **Apply Sample**
- Verify: table shows 100 rows, **"Save Changes"** button appears next to Apply
- Click **Cancel** → verify table restores to original 300 rows
- Re-open Sample → enter `120` → click **Apply Sample**
- Verify: table shows 120 rows (sampled from original 300, not from 100) ✅
- Click **Save Changes** → verify data is persisted (refresh page to confirm)

**2. Filter preview**

- Open **Filter** → set a filter condition → click **Apply Filter**
- Verify: filtered data shows in table, "Save Changes" appears
- Click **Cancel** → verify original data is restored
- Re-apply filter → click **Save Changes** → verify it persists

**3. No regression on non-preview transforms**

- Verify Sort, Rename Column, Cell Edit, Add/Delete Row still work as before (they don't use preview mode)

**4. Backend verification**

- Apply a preview → check that `_copy.csv` has NOT changed on disk
- Apply a preview → check that `user_logs` table has NO new entry
- Click Save Changes → confirm both file and log are updated

---

### Checklist

- [x] Backend tests pass (`uv run pytest -v`)
- [x] Frontend tests pass (`npm run test`)
- [x] Linting passes (backend: `uv run ruff check .`, frontend: `npm run lint`)
- [x] Manual testing completed for Sample, Filter, and Drop Duplicates

### DIFF
--- a/dataloom-backend/app/api/endpoints/transformations.py
+++ b/dataloom-backend/app/api/endpoints/transformations.py
@@ -6,7 +6,7 @@ All transformations are handled through a single unified /transform endpoint.
 import re
 import uuid
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, HTTPException, Query
 from sqlmodel import Session
 
 from app import database, models, schemas
@@ -104,6 +104,7 @@ def _dispatch_transform(df, transformation_input):
 async def transform_project(
     project_id: uuid.UUID,
     transformation_input: schemas.TransformationInput,
+    preview: bool = Query(False, description="If true, return transformation data without saving."),
     db: Session = Depends(database.get_db),
     project: models.Project = Depends(get_project_or_404),
 ):
@@ -121,7 +122,7 @@ async def transform_project(
 
         result_df, should_save = _dispatch_transform(df, transformation_input)
 
-        if should_save:
+        if should_save and not preview:
             save_table_safe(result_df, project.file_path)
             try:
                 log_transformation(db, project_id, operation_type, transformation_input.dict())
--- a/dataloom-frontend/src/Components/MenuNavbar.jsx
+++ b/dataloom-frontend/src/Components/MenuNavbar.jsx
@@ -68,7 +68,7 @@ const MenuNavbar = ({ projectId }) => {
   const [confirmData, setConfirmData] = useState(null);
   const [toast, setToast] = useState(null);
 
-  const { updateData, refreshProject, pageSize, projectName } = useProjectContext();
+  const { updateData, refreshProject, pageSize, projectName, isPreviewMode } = useProjectContext();
 
   const fetchLogs = useCallback(async () => {
     try {
@@ -297,54 +297,63 @@ const MenuNavbar = ({ projectId }) => {
             label: "Filter",
             icon: LuFilter,
             formType: "FilterForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("FilterForm"),
           },
           {
             label: "Sample",
             icon: LuDice5,
             formType: "SampleRowsForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("SampleRowsForm"),
           },
           {
             label: "Sort",
             icon: LuArrowUpDown,
             formType: "SortForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("SortForm"),
           },
           {
             label: "Drop Dup",
             icon: LuCopyMinus,
             formType: "DropDuplicateForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("DropDuplicateForm"),
           },
           {
             label: "GroupBy",
             icon: LuGroup,
             formType: "GroupByForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("GroupByForm"),
           },
           {
             label: "Cast Type",
             icon: LuRefreshCw,
             formType: "CastDataTypeForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("CastDataTypeForm"),
           },
           {
             label: "Trim Space",
             icon: LuScissors,
             formType: "TrimWhitespaceForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("TrimWhitespaceForm"),
           },
           {
             label: "Replace",
             icon: LuReplace,
             formType: "StringReplaceForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("StringReplaceForm"),
           },
           {
             label: "Fill Empty",
             icon: LuEraser,
             formType: "FillEmptyForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("FillEmptyForm"),
           },
         ],
@@ -356,18 +365,21 @@ const MenuNavbar = ({ projectId }) => {
             label: "Adv Query",
             icon: LuCode,
             formType: "AdvQueryFilterForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("AdvQueryFilterForm"),
           },
           {
             label: "Pivot Table",
             icon: LuTable2,
             formType: "PivotTableForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("PivotTableForm"),
           },
           {
             label: "Melt (Unpivot)",
             icon: LuLayoutList,
             formType: "MeltForm",
+            disabled: isPreviewMode,
             onClick: () => handleMenuClick("MeltForm"),
           },
         ],
@@ -407,9 +419,12 @@ const MenuNavbar = ({ projectId }) => {
                       key={item.label}
                       data-testid={`toolbar-${item.label.toLowerCase().replace(/ /g, "-")}`}
                       onClick={item.onClick}
-                      className={`flex flex-col items-center gap-1 px-3 py-1.5 rounded-md ${
-                        isActive ? "bg-blue-50 text-blue-600" : "hover:bg-gray-100"
-                      }`}
+                      disabled={item.disabled}
+                      className={`flex flex-col items-center gap-1 px-3 py-1.5 rounded-md transition-colors ${
+                        isActive
+                          ? "bg-blue-50 text-blue-600"
+                          : "hover:bg-gray-100 disabled:hover:bg-transparent"
+                      } ${item.disabled ? "opacity-50 cursor-not-allowed" : ""}`}
                     >
                       <item.icon
                         className={`w-5 h-5 ${isActive ? "text-blue-600" : "text-gray-600"}`}
--- a/dataloom-frontend/src/Components/common/ColumnMultiSelect.tsx
+++ b/dataloom-frontend/src/Components/common/ColumnMultiSelect.tsx
@@ -51,7 +51,9 @@ const ColumnMultiSelect = ({
         className="border border-gray-300 rounded-md w-full px-3 py-2 mb-1 text-sm bg-white text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none"
       />
       <div className="border border-gray-300 rounded-md max-h-40 overflow-y-auto p-2 bg-gray-50">
-        {filtered.length === 0 && <p className="px-1 py-1 text-sm text-gray-400">No columns found</p>}
+        {filtered.length === 0 && (
+          <p className="px-1 py-1 text-sm text-gray-400">No columns found</p>
+        )}
         {filtered.map((col) => (
           <label
             key={col}
--- a/dataloom-frontend/src/Components/forms/AdvQueryFilterForm.jsx
+++ b/dataloom-frontend/src/Components/forms/AdvQueryFilterForm.jsx
@@ -1,41 +1,48 @@
 import { useState } from "react";
 import PropTypes from "prop-types";
-import TransformResultPreview from "./TransformResultPreview";
 import { transformProject } from "../../api";
 import { ADV_QUERY_FILTER } from "../../constants/operationTypes";
 import useError from "../../hooks/useError";
+import usePreviewSave from "../../hooks/usePreviewSave";
 import FormErrorAlert from "../common/FormErrorAlert";
 import { useProjectContext } from "../../context/ProjectContext";
 import Button from "../common/Button";
 
 const AdvQueryFilterForm = ({ projectId, onClose }) => {
   const [query, setQuery] = useState("");
-  const [result, setResult] = useState(null);
   const [loading, setLoading] = useState(false);
   const { error, clearError, handleError } = useError();
-  const { updateData, refreshProject, pageSize } = useProjectContext();
+  const { isPreviewMode, enterPreviewMode, cancelPreview } = useProjectContext();
+  const { saving, handleSave } = usePreviewSave({
+    clearError,
+    handleError,
+    onClose,
+  });
 
   const handleSubmit = async (e) => {
     e.preventDefault();
     setLoading(true);
     clearError();
     try {
-      const response = await transformProject(projectId, {
+      const payload = {
         operation_type: ADV_QUERY_FILTER,
         adv_query: { query },
-      });
-      setResult(response);
-      updateData(response.columns, response.rows, {
-        dtypes: response.dtypes,
-        resetColumnOrder: false,
-      });
-      await refreshProject(projectId, 1, pageSize);
+      };
+      const response = await transformProject(projectId, payload, { preview: true });
+      enterPreviewMode(response.columns, response.rows, response.dtypes, { projectId, payload });
     } catch (err) {
       console.error("Error applying query:", err.message);
       handleError(err);
     } finally {
       setLoading(false);
     }
+  };
+
+  const handleClose = () => {
+    if (isPreviewMode) {
+      cancelPreview();
+    }
+    onClose();
   };
 
   return (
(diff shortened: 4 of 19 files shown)


ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `dataloom-frontend/src/Components/forms/AdvQueryFilterForm.jsx:32` AdvQueryFilterForm enters preview mode but has no 'Save Changes' button and the PreviewActionBar is never rendered anywhere, so there is no code path to persist an advanced-query filter preview. The save is unreachable.
- **[major/bug]** `dataloom-frontend/package.json:121` playwright is added to production dependencies instead of devDependencies. Playwright is a 200 MB+ E2E test runner and must not be bundled into the production frontend artifact.
- **[major/bug]** `dataloom-frontend/src/Components/common/PreviewActionBar.jsx:1` PreviewActionBar.jsx is created as a new file but is no longer imported by DataScreen or Table (both removed it). The file is dead code and will never render.
- **[major/bug]** `dataloom-backend/app/api/endpoints/transformations.py:27` The return value _should_save from _dispatch_transform is silently discarded. The original logic used should_save to skip writes for certain operations (e.g. no-op transforms). Replacing it with 'if not preview' means every preview=False call now unconditionally writes to disk regardless of the operation's intent, potentially corrupting state for operations that returned should_save=False.
- **[major/bug]** `playwright.config.js:1679` The global test timeout (90_000 ms) was removed from playwright.config.js with no replacement. Tests that do not individually set a timeout now fall back to Playwright's 30 s default, which is likely to cause CI failures for slower integration tests that previously relied on the 90 s ceiling.
- **[nit/style]** `e2e/transformations.spec.js:1669` File is missing a newline at end of file.
- **[nit/style]** `playwright.config.js:1698` File is missing a newline at end of file.

**Conventions**
- **PR title:** Title reads like an issue description ('Sample Tool — Destructive Mutation & UX Issues') rather than a Conventional Commits imperative summary of what the PR does; something like 'fix: add preview-before-persist flow for row-reducing transforms' would be appropriate.
- _The PR description is thorough and well-structured; the only hygiene gap is the non-Conventional-Commits PR title._

The core idea — gating row-reducing transforms behind a preview-before-persist flow — is sound and the backend change is minimal and backward-compatible. However, there are two blocking issues: AdvQueryFilterForm enters preview mode but the only component that could confirm/save the preview (PreviewActionBar) was removed from all render trees, leaving advanced-query filter previews with no save path; and playwright was accidentally added as a production dependency rather than a devDependency. Additionally, discarding the _should_save return value from _dispatch_transform silently changes semantics for any operation that previously signaled it should not write, and the removal of the global 90 s Playwright timeout without a per-test replacement risks CI instability. These should be resolved before merge.

YADUNANDAN-SINGH:
@ivantha Thanks for the thorough review! I've gone ahead and addressed all the feedback:

- **`AdvQueryFilterForm.jsx`**: Added the `Save Changes` button and logic so it now properly follows the preview-before-persist flow, just like the other forms.
- **`package.json`**: Moved `playwright` out of production `dependencies` and into `devDependencies`.
- **Dead Code**: Deleted `PreviewActionBar.jsx` since it's no longer used anywhere.
- **`transformations.py`**: Updated the logic to `if _should_save and not preview:` to ensure operations that are not meant to save (no-ops) don't accidentally write to disk.
- **`playwright.config.js`**: Restored the `timeout: 90000` global timeout to prevent slower integration tests from failing.
- **Formatting**: Fixed the missing trailing newlines in both `e2e/transformations.spec.js` and `playwright.config.js`.

I've also squashed everything down into a single commit with the updated Conventional Commits title (`fix: add preview-before-persist flow for row-reducing transforms`) as requested. 

Let me know if there's anything else needed!

ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `dataloom-frontend/package.json:121` playwright is listed as a runtime dependency in package.json rather than a devDependency. The Playwright package (~100 MB, installs Chromium/Firefox/WebKit browser binaries) does not belong in the production bundle. This inflates the production install and may break deployment environments. It should be under devDependencies alongside vitest and other test tooling.
- **[major/bug]** `dataloom-frontend/src/context/ProjectContext.jsx:785` enterPreviewMode captures the snapshot only when previewSnapshot is null (the guard `if (!previewSnapshot)` on line 185). If the user clicks Apply a second time while already in preview mode (e.g., re-applies a different sample size without clicking Cancel first), the snapshot is NOT refreshed and still points to the original pre-preview data. This is the intended behaviour, but the problem is that the new preview result is displayed while pendingTransform is updated to the new payload. If the user then clicks Save Changes, the second transform is re-executed against the backend, which is fine. However if the user clicks Cancel, they correctly restore from the original snapshot. The subtle bug is that no UX guard prevents the user from clicking Apply again while in preview mode; the toolbar buttons are disabled but the Apply button inside the open form remains active. A user can therefore layer multiple Apply calls, each overwriting the displayed data, before clicking Save or Cancel, and only Cancel correctly restores from the original snapshot. The described scenario should be tested or the Apply button should also be disabled after the first preview.
- **[major/bug]** `dataloom-frontend/src/Components/forms/SampleRowsForm.jsx:654` The save path in all four forms (FilterForm, SampleRowsForm, DropDuplicateForm, AdvQueryFilterForm) reads the payload from pendingTransform.payload and re-posts it to the backend with preview=false. For Sample specifically, if the transform involves a random sample without a fixed seed (randomSeed is blank), the backend will draw a different random sample on the second call. The user sees one set of rows in the preview and persists a different set. This is a semantic correctness issue inherent to the double-request design.
- **[minor/quality]** `dataloom-frontend/src/Components/forms/FilterForm.jsx:537` The four handleSave implementations (FilterForm, SampleRowsForm, DropDuplicateForm, AdvQueryFilterForm) are byte-for-byte identical: guard on pendingTransform.payload, setSaving(true), transformProject with preview:false, confirmPreview(), refreshProject(), onClose(). This pattern should be extracted into a shared hook (e.g., usePreviewSave) or a helper in ProjectContext to avoid the four-way duplication.
- **[minor/quality]** `dataloom-frontend/src/Components/MenuNavbar.jsx:239` The disabled styling on toolbar buttons is incomplete: the `disabled:hover:bg-transparent` Tailwind class is applied via the `className` string but toolbar items are rendered as plain `<button>` elements that also inherit text colour from the item icon's conditional class. When `item.disabled` is true the icon class adds `text-gray-400`, but the label `<span>` still uses `text-gray-700`, so the label remains full-contrast while the icon is dimmed. A consistent `opacity-50` or a unified disabled colour class on the wrapping element would be cleaner.
- **[nit/style]** `dataloom-backend/app/api/endpoints/transformations.py:27` The variable renaming from `should_save` to `_should_save` in transformations.py is unnecessary. The underscore-prefix convention in Python signals 'intentionally unused', but `_should_save` is in fact used in the conditional on the next line. Using the original name `should_save` is more readable and avoids misleading the reader.
- **[nit/style]** `playwright.config.js:5` playwright.config.js: the `timeout` field was moved from line 12 (inside the config body, after testDir) to line 5 (before testDir). The functional result is identical but the churn in the config file is unnecessary and mixes a config fix (DATABASE_URL default credentials) with a cosmetic reorder.

**Conventions**
- **PR title:** The title 'Sample Tool — Destructive Mutation & UX Issues' reads like a problem statement or issue title, not a Conventional Commits imperative summary of what the PR does. A clearer title would be something like 'fix: add preview-before-persist flow for row-reducing transforms'.
- _Single clean commit with a well-formed Conventional Commits message; the PR title is the only hygiene gap._

This PR fixes a real and impactful data-loss bug where row-reducing transforms permanently overwrote the working CSV, making it impossible to re-sample or re-filter from the original data. The backend change is minimal and correct; the preview flag is backward-compatible. The most important issue to resolve before merging is that playwright is accidentally installed as a runtime dependency instead of a devDependency, which will bulk up and potentially break production installs. The random-seed double-request issue (two separate sample calls without a fixed seed can yield different rows in preview vs persisted result) is a semantic correctness concern that should be acknowledged or addressed. The four handleSave implementations are also identical and should be consolidated into a shared hook before the codebase accumulates more forms with this pattern.

YADUNANDAN-SINGH:
@ivantha Thanks for the detailed review! I've addressed all the findings:

**Blocking / Major fixes:**

- **`playwright` in prod deps** — This was already moved to `devDependencies` in a prior push; the bot may have been reviewing a stale diff. Confirmed it's only in `devDependencies` now.
- **Apply button not disabled during preview** — All four form Apply/Submit buttons now include `|| isPreviewMode` in their `disabled` prop, preventing layered previews.
- **Random sample double-request divergence** — `SampleRowsForm` now auto-generates a random seed (`Math.floor(Math.random() * 4294967295)`) when the user leaves the seed field blank. This ensures the preview and save requests use the same seed and produce identical results.
- **Duplicate `handleSave` across 4 forms** — Extracted into a shared `usePreviewSave` hook (`src/hooks/usePreviewSave.js`). All four forms (FilterForm, SampleRowsForm, DropDuplicateForm, AdvQueryFilterForm) now use it.

**Minor / Nit fixes:**

- **Inconsistent disabled styling** — Replaced the per-element `text-gray-400` on the icon with a unified `opacity-50 cursor-not-allowed` on the wrapping `<button>`, so both icon and label dim consistently.
- **`_should_save` naming** — Renamed back to `should_save` since it's actively used in the conditional.
- **`timeout` position in `playwright.config.js`** — Moved back to its original position (after `reporter`) to avoid unnecessary config churn.
- **PR title** — Updated to `fix: add preview-before-persist flow for row-reducing transforms` to follow Conventional Commits format.

Everything is squashed into a single commit. All 72 unit tests pass locally and all 7 CI checks are green. Let me know if anything else needs attention!

ivantha:
@YADUNANDAN-SINGH Could you please squash all the commits and rebase?

YADUNANDAN-SINGH:
Done @ivantha! Squashed all commits into a single commit and rebased onto `main`. Please let me know if anything else needs to be addressed. 🙏

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/context/ProjectContext.jsx:724` enterPreviewMode does not update totalRows/totalPages to reflect the preview result. The snapshot captures the original values, but those page-level counters are not overwritten with values from the preview response, so the pagination UI continues to display the original row/page count while the preview data is shown in the table, misleading users about how many rows the filtered/sampled result contains.
- **[major/bug]** `dataloom-frontend/src/Components/forms/SampleRowsForm.jsx:586` In SampleRowsForm.handleSubmit, a random auto-seed is generated on each click of 'Apply Sample', and that same seed is stored in the payload and replayed on 'Save Changes'. However, because the seed is generated inside the submit handler each time the user clicks 'Apply Sample' again, a second preview click before saving will generate a fresh seed, making the preview result diverge from the pending payload still in pendingTransform. The auto-seed fix only protects a single preview-then-save round trip; clicking Apply a second time without saving first silently orphans the first pendingTransform. _(unverified)_
- **[minor/quality]** `dataloom-backend/app/api/endpoints/transformations.py:122` The log_transformation call is skipped when preview=True, but if an exception is thrown inside the try block after save_table_safe succeeds (e.g. a DB error in log_transformation), the saved file and the log will already be out of sync even in non-preview mode. This pre-existing issue is not introduced by this PR but the new code path makes the asymmetry more visible; no regression here, just worth noting.
- **[minor/quality]** `dataloom-frontend/src/api/transforms.js:684` transforms.js removes the window.dispatchEvent(new CustomEvent('project:logs-refresh')) call that previously fired on every transform. For non-preview transforms ('Save Changes' path) the logs refresh is now handled by refreshProject inside usePreviewSave, but for transforms that bypass the preview flow entirely (Sort, Rename, etc.) the logs panel will no longer auto-refresh after they are applied, since those forms still call transformProject directly without the preview flag.
- **[nit/quality]** `dataloom-frontend/package.json:105` playwright is added as a devDependency to dataloom-frontend/package.json, but the e2e tests already live at the repo root (not inside dataloom-frontend) and the root-level playwright.config.js references a separate fixtures setup. Installing playwright inside the frontend package is redundant and adds ~100 MB to frontend CI install time.

**Conventions**
- **PR title:** Title uses an em dash and sentence-case phrasing ('Sample Tool — Destructive Mutation & UX Issues') rather than a Conventional Commits prefix such as 'fix: add preview-before-persist flow for row-reducing transforms'. The commit message is correct; the PR title is not aligned with it.
- _The single commit message is well-formed; align the PR title to match it._

This PR correctly diagnoses and fixes the destructive-mutation bug: row-reducing transforms (Sample, Filter, Drop Duplicates, AdvQuery Filter) now go through a preview-before-persist flow backed by a new ?preview=true query parameter on the backend, and a shared usePreviewSave hook keeps the Save Changes logic DRY. The approach is sound and the description is thorough. Two issues need attention before merging: enterPreviewMode does not update totalRows/totalPages from the preview response, so the pagination counter stays wrong during preview; and removing the project:logs-refresh custom event from transformProject breaks the log panel auto-refresh for non-preview transforms (Sort, Rename, etc.) that do not go through usePreviewSave. The playwright devDependency added inside dataloom-frontend is redundant with the root-level Playwright setup and should be removed.

YADUNANDAN-SINGH:
@ivantha Thank you for the thorough review! I've addressed all the findings:

---

**[major/bug] `enterPreviewMode` not updating `totalRows`/`totalPages`**
Fixed — `enterPreviewMode` in `ProjectContext.jsx` now sets `totalRows = previewRows.length`, `totalPages = 1`, and `page = 1` when entering preview mode, so the pagination UI correctly reflects the filtered/sampled result count. `cancelPreview` already restored the original values from the snapshot, so that path was already correct.

---

**[major/bug] Auto-seed diverging on second Apply click in `SampleRowsForm`**
Already guarded — the `Apply Sample` button has `disabled={... || isPreviewMode}`, which prevents a second click while a preview is active. The snapshot guard in `enterPreviewMode` also preserves the original snapshot on re-entry, so no divergence is possible.

---

**[minor/quality] `project:logs-refresh` event removed from `transforms.js`**
Fixed — restored `window.dispatchEvent(new CustomEvent("project:logs-refresh"))` inside `transformProject`, but **only** when `preview === false`. This ensures Sort, Rename, and other non-preview transforms continue to trigger a log panel refresh, while preview-mode calls (which are not persisted) don't cause spurious refetches.

---

**[nit/quality] Redundant `playwright` in `dataloom-frontend/package.json`**
Fixed — removed `playwright` from the frontend `devDependencies`.

---

**PR title**
Updated to `fix: add preview-before-persist flow for row-reducing transforms` to align with the commit message and Conventional Commits convention.

ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `dataloom-frontend/src/hooks/usePreviewSave.js:24` confirmPreview() fires before the await refreshProject() call. If the refresh request throws, preview mode is already exited (snapshot and pendingTransform cleared) and the form closes, so the user has no retry path from the UI. The saved data on disk is correct, but the user sees an error toast with no clear recovery. Moving confirmPreview() to after the refreshProject() await keeps preview mode active on failure and lets the user retry.
- **[minor/test]** `dataloom-frontend/src/hooks/usePreviewSave.js:1` The new usePreviewSave hook has no unit tests. The happy path (transformProject succeeds, confirmPreview fires, refreshProject succeeds, onClose called) and the error path (transformProject or refreshProject throws, handleError called, saving reset to false) are not directly covered. Only FilterForm's enterPreviewMode call is tested; the save flow is exercised by no test.
- **[nit/quality]** `dataloom-backend/app/api/endpoints/transformations.py:107` preview is exposed as a query parameter via Query(False) on a POST endpoint. REST convention expects all non-path control flags on POST requests to live in the request body, and some caches or reverse proxies key on query strings even for POST. Adding preview to the TransformationInput schema (defaulting to False) would keep the POST body self-contained with zero breaking changes.

**Conventions**

The PR correctly fixes the row-reducer overwrite bug by introducing a two-step preview-then-persist flow: the backend preview flag skips both the disk write and the log entry, and the frontend context stores a snapshot for cancel/restore. The main correctness concern is in usePreviewSave.js: confirmPreview() fires before the refreshProject() await, so a failed refresh exits preview mode without any retry path from the UI (data integrity on disk is preserved, but the UX lands in an ambiguous state). Test coverage for the new hook's save-success and save-error paths is absent. The preview query parameter on a POST endpoint is a minor convention deviation worth a quick team discussion before merge.

PR #385, opened 2026-06-20
feat: add searchable project listing across all user projects

### LINKED ISSUE #384: [Feature] Support searching and browsing all projects, not just the 10 most recent
The Home Screen only fetches and displays the user's 10 most recently modified projects via `getRecentProjects()`. Once a user has more than 10 projects, older ones become completely inaccessible from the Home Screen, there's no pagination or way to browse beyond the most recent 10. A prior attempt (#280, closed due to inactivity) added a client-side search bar, but it only filtered within this already-limited 10-item list, so it couldn't help users find older projects either.

Add a backend endpoint supporting searchable, paginated project listing, matching on both project name and description. Update the Home Screen to query this endpoint as the user types (debounced), replacing the fixed `/projects/recent` list when a search is active.

Increasing the `recent` limit from 10 to a larger fixed number — rejected as it doesn't scale and still imposes an arbitrary cutoff.

**Additional context**
- Supersedes #252 / #280 — fixes the same UX gap with the correct underlying scope
- Current implementation: `get_recent_projects(db, owner_id, limit=10)` in `project_service.py`, called from `GET /projects/recent`
- Search should be case-insensitive and match name and description

SHIVAM-KUMAR-59:
## Description
The Home Screen previously only fetched and displayed the user's 10 most recently modified projects via `/projects/recent`, with no way to find older projects. A prior attempt (#280, closed due to inactivity) added client-side search, but it only filtered within this already-limited 10-item list.

This PR adds a proper backend-powered search: a new `search_projects()` service function and `GET /projects/search?q=<query>` endpoint that searches across **all** of a user's projects by name or description (case-insensitive), not just the recent 10. The Home Screen now debounces search input (500ms) and queries this endpoint, falling back to the existing recent-projects view when the search box is empty.

Fixes #384 

## Type of Change
- [x] New feature

## How Has This Been Tested?
- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

**Verified following scenarios manually:**
- Created 12+ projects, confirmed the oldest (outside the recent-10 window) is findable via search
- Searched by partial project name and by description text — both match correctly
- Verified search is case-insensitive
- Verified search respects user ownership — no cross-user project leakage
- Verified empty/whitespace query returns no results without erroring
- Verified projects with `description = None` don't break the search query
- Verified debounce prevents excessive API calls while typing
- Verified search bar is hidden when the user has zero projects
- Verified "No projects found" empty state renders for non-matching queries

## Screen Recording

https://github.com/user-attachments/assets/af0d7fe5-49f5-405f-9424-da4ff0205c99

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/projects.py
+++ b/dataloom-backend/app/api/endpoints/projects.py
@@ -26,6 +26,7 @@ from app.services.project_service import (
     get_last_change_log,
     get_recent_projects,
     rename_project,
+    search_projects,
 )
 from app.services.transformation_service import apply_logged_transformation
 from app.utils.file_formats import TableWriteOptions, get_format, get_format_for_extension
@@ -431,3 +432,24 @@ async def undo_last_transformation(
         "total_pages": 1,
         **resp,
     }
+
+
+@router.get("/search", response_model=list[schemas.LastResponse])
+def search_user_projects(
+    q: str,
+    db: Session = Depends(database.get_db),
+    current_user: models.User = Depends(get_current_user),
+):
+    """Search the current user's projects by name or description."""
+    if not q or not q.strip():
+        return []
+    projects = search_projects(db, owner_id=current_user.id, query=q.strip(), limit=20)
+    return [
+        schemas.LastResponse(
+            project_id=p.project_id,
+            name=p.name,
+            description=p.description,
+            last_modified=p.last_modified,
+        )
+        for p in projects
+    ]
--- a/dataloom-backend/app/services/project_service.py
+++ b/dataloom-backend/app/services/project_service.py
@@ -3,6 +3,7 @@
 import uuid
 from datetime import UTC, datetime
 
+import sqlalchemy as sa
 from fastapi import HTTPException
 from sqlalchemy.exc import SQLAlchemyError
 from sqlmodel import Session
@@ -282,3 +283,31 @@ def rename_project(db: Session, project: models.Project, new_name: str) -> model
     db.refresh(project)
     logger.info("Renamed project: id=%s, new_name=%s", project.project_id, trimmed_name)
     return project
+
+
+def search_projects(db: Session, owner_id: uuid.UUID, query: str, limit: int = 20) -> list[models.Project]:
+    """Search a user's projects by name or description, case-insensitive.
+
+    Args:
+        db: Database session.
+        owner_id: Restrict results to projects owned by this user.
+        query: Search string to match against name/description.
+        limit: Maximum number of results to return.
+
+    Returns:
+        List of matching Project model instances ordered by last_modified desc.
+    """
+    pattern = f"%{query}%"
+    return (
+        db.query(models.Project)
+        .filter(
+            models.Project.owner_id == owner_id,
+            sa.or_(
+                models.Project.name.ilike(pattern),
+                models.Project.description.ilike(pattern),
+            ),
+        )
+        .order_by(models.Project.last_modified.desc())
+        .limit(limit)
+        .all()
+    )
--- a/dataloom-backend/tests/test_search_project.py
+++ b/dataloom-backend/tests/test_search_project.py
@@ -1,0 +1,116 @@
+"""Tests for project search functionality."""
+
+from app import models
+from app.services.project_service import create_project, search_projects
+
+
+class TestSearchProjects:
+    def test_search_matches_project_by_name(self, db, test_user):
+        create_project(db, name="Sales Analysis Q1", file_path="/tmp/test.csv", description="", owner_id=test_user.id)
+        create_project(db, name="Marketing Report", file_path="/tmp/test.csv", description="", owner_id=test_user.id)
+
+        results = search_projects(db, owner_id=test_user.id, query="sales")
+
+        assert len(results) == 1
+        assert results[0].name == "Sales Analysis Q1"
+
+    def test_search_matches_project_by_description(self, db, test_user):
+        create_project(
+            db,
+            name="Q1 Data",
+            file_path="/tmp/test.csv",
+            description="Quarterly revenue breakdown",
+            owner_id=test_user.id,
+        )
+        create_project(
+            db, name="Q2 Data", file_path="/tmp/test.csv", description="Customer churn analysis", owner_id=test_user.id
+        )
+
+        results = search_projects(db, owner_id=test_user.id, query="revenue")
+
+        assert len(results) == 1
+        assert results[0].name == "Q1 Data"
+
+    def test_search_is_case_insensitive(self, db, test_user):
+        create_project(db, name="Sales Analysis", file_path="/tmp/test.csv", description="", owner_id=test_user.id)
+
+        results = search_projects(db, owner_id=test_user.id, query="SALES")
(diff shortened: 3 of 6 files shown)


hanzalahwaheed:
question: do we fetch all the projects on the homescreen? if yes, the projects search can be in-memory, not a db call.

SHIVAM-KUMAR-59:
@hanzalahwaheed  No, the Home Screen only fetches the 10 most recent projects via `/projects/recent`, it never loads the user's full project set into memory. That's actually the root problem this PR fixes: with only 10 projects in memory, any in-memory search can only ever search those same 10 — which is exactly the limitation of the original #280 approach.
Doing an in-memory search would require fetching all of a user's projects upfront on every Home Screen load, which doesn't scale (a user with hundreds of projects would pay that cost on every visit even if they never search) and duplicates work the database is already well-suited to do via an indexed `ILIKE` query. A debounced server-side search keeps the initial load light and only queries the full project set when the user actually searches.

ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `dataloom-frontend/src/api/projects.js:139` The PR description explicitly states it adds a backend service function `search_projects()` and a `GET /projects/search?q=<query>` FastAPI endpoint, but the diff contains zero backend changes. The three changed files are all under `dataloom-frontend/`. As a result, the frontend will call `/projects/search` and receive a 404 (or route-collision error), making the feature completely non-functional.
- **[major/quality]** `dataloom-frontend/src/Components/Homescreen.jsx:76` `fetchRecentProjects` is a plain `const` arrow function (not wrapped in `useCallback`) but is referenced inside `useEffect(() => { fetchRecentProjects(); }, [])` with an empty dependency array. The react-hooks/exhaustive-deps rule requires it in the deps array. Because it is recreated every render, the effect would re-fire on every render if the deps array were correct. The function should be wrapped in `useCallback` (with its own stable deps) and then listed in the effect's deps array.
- **[major/bug]** `dataloom-frontend/src/Components/Homescreen.jsx:291` After a successful delete, the code re-fetches recent projects AND immediately re-queries search with no debounce, making two sequential API calls in the delete handler. If the search query is long or the user deletes rapidly, this can fire many un-debounced search requests. The post-delete search refresh should go through the existing debounce mechanism (e.g. by nudging `searchQuery` state) or be guarded with its own abort/loading flag.
- **[minor/quality]** `dataloom-frontend/src/Components/Homescreen.jsx:114` The PR description states the debounce delay is 500 ms, but the implementation uses 300 ms. While 300 ms is a reasonable value, the mismatch between the description and code is a documentation inconsistency that can confuse future reviewers.
- **[minor/quality]** `dataloom-frontend/src/Components/Homescreen.jsx:392` The search bar is conditionally rendered only when `recentProjects.length > 0`, but search operates over all projects. If a user's oldest projects fall outside the recent-10 window but the user has no projects in the recent list (e.g. immediately after all projects are deleted), the search bar disappears even though projects still exist. The visibility guard should be based on total project count (from the backend) rather than the length of the recent-projects array.

**Conventions**
- _Hygiene is good overall: one clean commit, a Conventional-Commits-style title, and a thorough description with a screen recording and a manual-testing checklist._

This PR adds frontend plumbing for a project search feature but is missing its backend half entirely: no FastAPI route, no service function, and no database query are present in the diff, so the feature will return a 404 on every search request. The PR cannot be merged as-is. Once the backend is added, two secondary issues should be addressed: `fetchRecentProjects` needs `useCallback` so the `useEffect` dependency array is correct, and the post-delete search re-fetch should use the existing debounce path instead of making a raw unguarded API call.

SHIVAM-KUMAR-59:
Thanks for catching this — the backend changes (`search_projects()` service function and `GET /projects/search` endpoint) existed locally but were never actually committed/pushed in the original PR. Apologies for the confusion. Addressed all findings:
 
  - **[blocking/bug] missing backend** — fixed. Pushed `search_projects()` in `project_service.py` and the `GET /projects/search` endpoint in `routers/projects.py`, along with a full test suite (`test_search_projects.py`) covering name/description matching, case-insensitivity, ownership isolation, null descriptions, limits, and the beyond-recent-10 case the feature was built for.
 
  - **[major/quality] `fetchRecentProjects` missing `useCallback`** — fixed. Wrapped in `useCallback` with an empty dependency array and added to the `useEffect` deps array.
 
  - **[major/bug] un-debounced search re-fetch on delete** — fixed. Removed the extra API call entirely; the deleted project is now filtered out of `searchResults` locally, since we already know its ID from the delete action.
 
  - **[minor/quality] debounce description mismatch** — the implementation has always used 300ms; updating the PR description to match rather than the code, since 300ms is the intended value.
 
  - **[minor/quality] search bar disappearing when `recentProjects` is empty** — fixed. The search bar is now always rendered, and the empty-state vs. grid logic now checks `recentProjects.length === 0 && !isSearching`, so searching still works even when the recent list is empty.
 
  Ready for re-review. @ivantha

ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `dataloom-backend/app/services/project_service.py:305` LIKE metacharacter injection: user-supplied query is interpolated directly into the LIKE pattern via an f-string (`pattern = f"%{query}%"`). SQLAlchemy parameterizes the bound value so there is no SQL injection, but a query containing a literal `%` or `_` will be interpreted as a LIKE wildcard rather than a literal character. A user searching for "100%" or "col_name" will get unintended wildcard matches. The pattern should escape those characters before building the string, e.g. `query.replace('%', r'\%').replace('_', r'\_')`.
- **[nit/quality]** `dataloom-frontend/src/Components/Homescreen.jsx:153` Debounce value mismatch between PR description and code. The description states 500 ms debounce; the implementation uses 300 ms. Minor but worth keeping in sync to avoid confusion during future tuning.
- **[nit/quality]** `dataloom-frontend/src/api/projects.js:132` The JSDoc return type annotation on `searchProjects` says `@returns {Promise<Object>}` but the actual return value is an array of project objects. Should be `{Promise<Array<Object>>}` (or the schema type) to match the backend `list[schemas.LastResponse]` response.
- **[nit/test]** `dataloom-backend/tests/test_search_project.py:66` The test `test_search_handles_project_with_no_description` creates a project named "No Description Project" with `description=None` and searches for the term "description". The search matches the word in the project *name*, not through the null-description path, so the test does not actually exercise the null-description code branch (the `ilike` on a NULL column). Renaming to something like `test_search_with_null_description_does_not_crash` and asserting `len(results) == 0` (or using a query that would only match the description) would make the intent clearer.

**Conventions**

This PR adds a well-structured backend-powered project search: a `search_projects` service function, a `GET /projects/search?q=` endpoint scoped to the authenticated user, debounced frontend integration, and a solid test suite covering ownership isolation, case-insensitivity, and edge cases. The main correctness concern is LIKE metacharacter injection -- queries containing `%` or `_` produce unintended wildcard matches because those characters are not escaped before being embedded in the pattern string; this should be fixed before merge. The other findings are minor: a 300 ms vs 500 ms debounce discrepancy with the PR description, an inaccurate JSDoc return type, and a test whose intent does not match what it actually exercises.

PR #389, opened 2026-06-21
Feat/select component

### LINKED ISSUE #388: [Feature] Reusable Select component
Lots of places use browsers native select, disrupting the UI. We need something cleaner and better.

A resuable select component for DRY

hanzalahwaheed:
## Description

- add Select, a generic styled single-select sharing ColumnSelect's look, animation, and keyboard navigation but taking a plain {value,label}[] list
- replace native <select> elements in the filter, sort, pivot, group by, cast, drop duplicate, fill empty, and export forms
- add tests covering placeholder, selection, keyboard navigation, and escape-to-close

Fixes #388

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [X] New tests added
- [X] Manual testing

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/ExportModal.tsx
+++ b/dataloom-frontend/src/Components/ExportModal.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useState, type CSSProperties, type FormEvent } from "react";
 import Modal from "./common/Modal";
 import Button from "./common/Button";
+import Select from "./common/Select";
 import { exportProject } from "../api";
 
 type FormatDef = {
@@ -247,17 +248,7 @@ export default function ExportModal({
               {/* Encoding (secondary — quiet dropdown) */}
               <label className="block">
                 <span className="mb-1.5 block text-xs text-gray-500">Encoding</span>
-                <select
-                  value={encoding}
-                  onChange={(e) => setEncoding(e.target.value)}
-                  className="w-full rounded-lg border border-gray-300 bg-white px-3 py-2 text-sm text-gray-800 transition-colors focus:border-blue-500 focus:outline-none focus:ring-2 focus:ring-blue-500/25"
-                >
-                  {ENCODINGS.map((enc) => (
-                    <option key={enc.value} value={enc.value}>
-                      {enc.label}
-                    </option>
-                  ))}
-                </select>
+                <Select value={encoding} onChange={setEncoding} options={ENCODINGS} />
               </label>
             </div>
           ) : (
--- a/dataloom-frontend/src/Components/common/ColumnSelect.tsx
+++ b/dataloom-frontend/src/Components/common/ColumnSelect.tsx
@@ -149,6 +149,7 @@ const ColumnSelect = ({
         aria-expanded={open}
         aria-required={required}
         onClick={toggle}
+        onKeyDown={handleKeyDown}
         className="flex items-center justify-between gap-2 border border-gray-300 rounded-md px-3 py-2 w-full bg-white text-left text-gray-900 focus:ring-2 focus:ring-blue-500 focus:border-blue-500 focus:outline-none disabled:opacity-50 disabled:cursor-not-allowed"
       >
         <span className="flex items-center min-w-0">
--- a/dataloom-frontend/src/Components/common/Select.tsx
+++ b/dataloom-frontend/src/Components/common/Select.tsx
@@ -1,0 +1,170 @@
+import { useEffect, useMemo, useRef, useState } from "react";
+
+export interface SelectOption {
+  value: string;
+  label: string;
+}
+
+interface SelectProps {
+  value: string;
+  onChange: (value: string) => void;
+  options: SelectOption[];
+  placeholder?: string;
+  id?: string;
+  disabled?: boolean;
+  className?: string;
+  "data-testid"?: string;
+}
+
+/**
+ * Generic styled dropdown. A trigger button opens an animated popover with a
+ * keyboard-navigable list of options. Shares the look and motion of ColumnSelect
+ * but takes a plain {value,label}[] list — used for the app's short, curated
+ * dropdowns (aggregation functions, sort order, target types, etc.).
+ */
+const Select = ({
+  value,
+  onChange,
+  options,
+  placeholder = "Select...",
+  id,
+  disabled = false,
+  className = "",
+  "data-testid": dataTestId,
+}: SelectProps) => {
+  const [open, setOpen] = useState(false);
+  // Kept mounted while the close animation plays, then unmounted on animationend.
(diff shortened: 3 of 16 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/forms/TrimWhitespaceForm.jsx:1527` TrimWhitespaceForm passes the synthetic 'All string columns' sentinel as the first element of the options array to ColumnSelect (which expects column names from the project). When the user selects it, the string 'All string columns' is sent to the API as the column parameter instead of an empty string / dedicated sentinel. This is a behaviour change from the old code, where 'All string columns' was an <option value='All string columns'> — the backend must already handle that literal string — but mixing a display label into a column-name list is semantically fragile and bypasses the dtype-badge logic safely only by accident.
- **[major/bug]** `dataloom-frontend/src/Components/common/Select.tsx:574` The Select trigger button has onKeyDown wired to the button itself, but ColumnSelect wires onKeyDown to the search input inside the popover. In Select there is no search box, so ArrowDown/Up/Enter are handled on the trigger. However, once the popover is open the focus stays on the button (there is no explicit focus transfer to a list item or search field), which means keyboard navigation only works reliably while the button remains focused. If the user tabs away after opening the popover the keyboard handler is lost and the popover cannot be closed or navigated via keyboard without clicking back.
- **[minor/quality]** `dataloom-frontend/package.json:53` tailwindcss-animate is listed under devDependencies in package-lock.json but under devDependencies in package.json as well, yet the lock file entry has '"dev": true'. The tailwind.config.js imports it as a plugin, which runs at build time only, so devDependency is correct — but the package.json diff shows it placed in the devDependencies block without an explicit 'devDependencies' key (it appears mixed into the existing devDependencies list). Worth confirming no production bundle includes it.
- **[minor/quality]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:1114` AGG_FUNCTIONS is defined independently in both GroupByForm.jsx and PivotTableForm.jsx with identical entries. Extracting it to a shared constants file would avoid duplication and future drift.
- **[minor/quality]** `dataloom-frontend/src/Components/common/ColumnSelect.tsx:392` In ColumnSelect, the trigger button's display renders the dtype badge inline inside the <span> that also shows the selected value text (line 392: '{value || placeholder}{displayDtype && <DtypeBadge .../>}'). When a column is selected, the badge appears flush after the text inside the truncation span, which can cause it to be clipped. The badge should likely be outside the truncating span, matching how list items render it (gap-1 flex row).
- **[nit/quality]** `dataloom-frontend/src/Components/common/ColumnSelect.tsx:272` useProjectContext is cast as 'unknown as { columns: string[]; dtypes: Record<string,string> }' in both ColumnSelect.tsx and ColumnMultiSelect.tsx. If ProjectContext already exports a typed hook or the type is known, the double cast should be removed; at minimum the inline type should be extracted to a shared interface to avoid the same cast in two places.

**Conventions**
- **PR title:** Title uses a branch-name slug ('Feat/select component') instead of a Conventional Commits imperative sentence; prefer something like 'feat: add reusable Select dropdown and replace native selects across transform forms'.
- _PR title is the main hygiene gap; description and commit messages are solid._

This PR introduces a reusable styled Select component and a searchable, dtype-aware ColumnSelect / ColumnMultiSelect pair, then wires them into all transform forms as a drop-in replacement for native selects and free-text column inputs. The implementation is well-structured and ships unit and E2E test coverage. Two issues need attention before merge: (1) TrimWhitespaceForm passes the display string 'All string columns' directly as a column name in the options array, conflating UI labelling with backend contract — consider an includeEmptyOption + emptyLabel pattern as used in FillEmptyForm instead; (2) the Select component's keyboard handler on the trigger button has no focus-trap mechanism, so ArrowDown/Up/Enter only work reliably while the button itself is focused. The duplicate AGG_FUNCTIONS definitions across GroupByForm and PivotTableForm are a minor cleanup opportunity.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/forms/TrimWhitespaceForm.jsx:1551` TrimWhitespaceForm passes 'All string columns' as a literal column name in the options array rather than using the includeEmptyOption/emptyLabel mechanism that ColumnSelect provides. When a user selects that entry, the string 'All string columns' is sent to the backend as the column value instead of an empty string or a recognised sentinel. The backend almost certainly does not handle this string.
- **[minor/quality]** `dataloom-frontend/src/Components/forms/StringReplaceForm.jsx:1504` FormErrorAlert is rendered after the closing </form> tag in both StringReplaceForm and TrimWhitespaceForm. The error is visually present but lives outside the form element, breaking the semantic relationship and potentially confusing assistive technology that associates alerts with the form.
- **[minor/quality]** `dataloom-frontend/src/Components/common/ColumnSelect.tsx:382` The ColumnSelect trigger button does not receive an onKeyDown handler; keyboard navigation (ArrowDown/Enter/Escape) only activates after the search input auto-focuses on open. This diverges from Select, where the trigger button itself handles key events, and will silently fail if autofocus is ever suppressed.
- **[nit/quality]** `dataloom-frontend/package.json:53` tailwindcss-animate is listed under runtime dependencies in package.json but the package-lock.json entry marks it dev:true. It is a build-time Tailwind plugin and belongs in devDependencies in both files.
- **[nit/quality]** `dataloom-frontend/src/Components/common/ColumnSelect.tsx:405` The search input inside the ColumnSelect popover has no aria-label; only a placeholder attribute is present. Placeholder text disappears when the user types and is not a substitute for a persistent accessible label.

**Conventions**
- **PR title:** Title 'Feat/select component' uses a slash-separated branch-name style, not Conventional Commits format; should be something like 'feat(ui): add reusable Select dropdown and replace native selects'
- **Commits (3):** clean, conventional history
- _PR title is the only hygiene gap; commit messages and description are clean._

This PR introduces a generic `Select` component and a rewritten `ColumnSelect`/`ColumnMultiSelect` pair (with keyboard navigation, search, and dtype badges), then migrates all eight transform forms and the export modal from native `<select>` elements to these components. The implementation is well-structured, tested, and accompanied by E2E helper updates. The critical issue to fix before merging is in `TrimWhitespaceForm`: the sentinel option "All string columns" is injected as a plain entry in the `options` array instead of using the existing `includeEmptyOption` mechanism, so that string will be sent to the backend as a column name. There are also two minor issues: `FormErrorAlert` is rendered outside the `<form>` tag in `StringReplaceForm` and `TrimWhitespaceForm`, and `tailwindcss-animate` is misclassified as a runtime dependency in `package.json` when it is a build-time Tailwind plugin.

PR #391, opened 2026-06-21
feat(pagination): add first/last navigation and page jump input

### LINKED ISSUE #390: [Feature] Extend pagination buttons
- Lets add First/Last in the pagination buttons as well. Just keep previous/next is a little restrictive.
- Also, we should add an input for the user to enter a page no. and directly navigate to that page.

This is where I got the idea to add similar page navigation into our project. A tool called OpenRefine.

Image

SHIVAM-KUMAR-59:
## Description
Adds First/Last page navigation buttons and a direct page-jump input to `TablePagination`, modeled after OpenRefine-style pagination. Previously only Previous/Next were available, which made navigating large result sets (e.g. hundreds of pages) tedious. Also makes the pagination bar responsive for smaller viewports.

Fixes #390 

## Type of Change
- [x] New feature

## How Has This Been Tested?
- [x] Manual testing

Manually verified:
- First/Last buttons navigate correctly and disable at boundaries (page 1 / last page)
- Page jump input accepts valid page numbers and navigates via both the **Go** button and `Enter` key
- Invalid input (`0`, negative numbers, values greater than total pages, non-integers) triggers the native browser validation warning via `reportValidity()` and does not call `onPageChange`
- Input clears after a successful jump
- Layout checked at mobile (`<640px`), tablet, and desktop widths — info row and nav row stack correctly on small screens and sit inline from `sm:` up

## Screen Recording

https://github.com/user-attachments/assets/5507150c-3e3f-45ea-b96c-d60b2a14bbc6

## Checklist
- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/Table.jsx
+++ b/dataloom-frontend/src/Components/Table.jsx
@@ -16,6 +16,7 @@ import InputDialog from "./common/InputDialog";
 import Toast from "./common/Toast";
 import DtypeBadge from "./common/DtypeBadge";
 import PropTypes from "prop-types";
+import { useToast } from "../context/ToastContext";
 
 const MenuButton = ({ children, onClick }) => (
   <button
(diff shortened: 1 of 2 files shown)


hanzalahwaheed:
Lets move `Page X of Y` towards the right. And we can make the X editable.

SHIVAM-KUMAR-59:
Hey @hanzalahwaheed I have moved the Page number to the right as asked, here's the updated demo:

https://github.com/user-attachments/assets/22766ac4-c312-4ce7-bad5-774a66ff5dd5

hanzalahwaheed:
LGTM

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/Table.jsx:497` The page-synchronization pattern sets state unconditionally during render (the prevPageRef comparison block executes during the component body, not inside useEffect). Calling setPageInput during render is a side effect that can cause React to issue warnings in Strict Mode and can trigger a double-render. This should be replaced with a useEffect([page]) that calls setPageInput(String(page)).
- **[major/bug]** `dataloom-frontend/src/Components/Table.jsx:519` commitPageInput relies on browser native constraint validation via input.reportValidity(). For the 'page jump' flow triggered by Enter or blur, the input value at that moment is whatever the user typed but has not yet been reflected in the DOM's validity state if the number spinner was bypassed by typing directly. More critically, a user can type '0.5' or '1e2' — both are valid Numbers but violate the integer semantics expected by onPageChange. The check should use Number.isInteger(parsed) and explicit range guards (parsed >= 1 && parsed <= totalPages) instead of relying solely on reportValidity.
- **[minor/quality]** `dataloom-frontend/src/test/Table.columnOrder.test.jsx:46` The PR description says invalid input triggers a 'Go button', but the actual implementation uses Enter or blur, with no 'Go' button present. This is a minor documentation inconsistency in the PR body, but there is no test coverage verifying the commitPageInput validation path (out-of-range, non-integer, empty string), so regressions may go undetected.
- **[nit/style]** `dataloom-frontend/src/Components/Table.jsx:597` The aria-label on the page-jump input is 'Current page', which is more appropriate for a read-only display. A label like 'Jump to page' better communicates that the field is interactive and accepts user input.

**Conventions**

The PR adds first/last navigation buttons and a page-jump input to TablePagination, which is a useful UX improvement for large datasets. Two issues need attention before merging: the prevPageRef synchronization block runs as a render-time side effect and should be converted to a useEffect, and the commitPageInput validation relies on native browser constraint validation without guarding against non-integer values like 0.5 or 1e2 that pass Number() but are semantically invalid for page navigation. The test file was updated to wrap the existing column-order tests in ToastProvider, but no tests were added for the new pagination behaviors, so consider adding at least one test covering the page-jump validation path.

PR #394, opened 2026-06-23
feat(table): migrate to TypeScript and add spreadsheet layout

### LINKED ISSUE #393: [Feature] Make tables UI boxy and cellular
- Tables currently are rounded, take up a lot of space and arent optimised.
- Need to migrate to TS
- The dtypes in table header are inline or below. Need to make this consistent.

- Clean, sharp boxy tables similar to excel

hanzalahwaheed:
## Description

Fixes #393 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass
- [X] New tests added
- [X] Manual testing

## Screenshots

image

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/DataScreen.jsx
+++ b/dataloom-frontend/src/Components/DataScreen.jsx
@@ -21,7 +21,7 @@ export default function DataScreen() {
   };
 
   return (
-    <div className="flex flex-col h-full overflow-hidden">
+    <div className="flex flex-col flex-1 min-h-0 overflow-hidden">
       <MenuNavbar onTransform={handleTransform} projectId={projectId} />
       <Table projectId={projectId} data={tableData} />
     </div>
--- a/dataloom-frontend/src/Components/Table.tsx
+++ b/dataloom-frontend/src/Components/Table.tsx
@@ -158,6 +158,8 @@ const Table = ({ projectId, data: externalData }: TableProps) => {
     const { columns, rows, dtypes: newDtypes } = response;
     setColumns(["S.No.", ...columns]);
     setData(rows.map((row, index) => [index + 1, ...Object.values(row)]));
+    // updateData resets the saved column order when the column count changes,
+    // which covers add/delete column; rename and row ops keep the order.
     updateData(columns, normalizeRows(rows), { dtypes: newDtypes });
   };
 
@@ -168,7 +170,6 @@ const Table = ({ projectId, data: externalData }: TableProps) => {
         row_params: { index },
       });
       updateTableData(response);
-      updateData(response.columns, normalizeRows(response.rows), { resetColumnOrder: false });
       refreshProject(projectId, 1, pageSize);
     } catch {
       setToast({
@@ -202,7 +203,6 @@ const Table = ({ projectId, data: externalData }: TableProps) => {
             add_col_params: { index: backendIndex, name: newColumnName },
           });
           updateTableData(response);
-          updateData(response.columns, normalizeRows(response.rows), { resetColumnOrder: true });
           refreshProject(projectId, 1, pageSize);
         } catch {
           setToast({
@@ -223,7 +223,6 @@ const Table = ({ projectId, data: externalData }: TableProps) => {
         row_params: { index },
       });
       updateTableData(response);
-      updateData(response.columns, normalizeRows(response.rows), { resetColumnOrder: false });
       refreshProject(projectId, 1, pageSize);
     } catch {
       setToast({
@@ -259,7 +258,6 @@ const Table = ({ projectId, data: externalData }: TableProps) => {
             rename_col_params: { col_index: backendIndex, new_name: newName },
           });
           updateTableData(response);
-          updateData(response.columns, normalizeRows(response.rows), { resetColumnOrder: false });
           refreshProject(projectId, 1, pageSize);
         } catch {
           setToast({
@@ -290,7 +288,6 @@ const Table = ({ projectId, data: externalData }: TableProps) => {
         del_col_params: { index: backendIndex },
       });
       updateTableData(response);
-      updateData(response.columns, normalizeRows(response.rows), { resetColumnOrder: true });
       refreshProject(projectId, 1, pageSize);
     } catch {
       setToast({
--- a/dataloom-frontend/src/Components/Table.tsx
+++ b/dataloom-frontend/src/Components/Table.tsx
@@ -94,6 +94,11 @@ const MenuButton = ({ children, onClick }: MenuButtonProps) => (
   </button>
 );
 
+/** Normalize backend rows (arrays or column-keyed objects) to value arrays. */
+function normalizeRows(rows: Array<Cell[] | Record<string, Cell>>): Cell[][] {
+  return rows.map((row) => (Array.isArray(row) ? row : Object.values(row)));
+}
+
 interface TableProps {
   projectId: string;
   data?: ExternalData;
@@ -157,7 +162,7 @@ const Table = ({ projectId, data: externalData }: TableProps) => {
   const updateTableData = (response: TransformResponse) => {
     const { columns, rows, dtypes: newDtypes } = response;
     setColumns(["S.No.", ...columns]);
-    setData(rows.map((row, index) => [index + 1, ...Object.values(row)]));
+    setData(normalizeRows(rows).map((row, index) => [index + 1, ...row]));
     // updateData resets the saved column order when the column count changes,
     // which covers add/delete column; rename and row ops keep the order.
     updateData(columns, normalizeRows(rows), { dtypes: newDtypes });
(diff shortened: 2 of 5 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/Table.tsx:177` updateTableData calls updateData(columns, normalizeRows(rows), ...) and then the callers (handleAddRow, handleAddColumn, handleDeleteRow, handleRenameColumn, handleDeleteColumn) call updateData a second time with normalizeRows(response.rows). That means updateData is invoked twice per operation with the same normalized rows — once inside updateTableData and once again in the caller. The first call (inside updateTableData) always uses the full column list and normalized rows, while the second call in the caller passes the same data. This double-call was present before the PR for add-row / delete-row / delete-column operations but the refactor preserved and extended it to rename-column as well. On a successful rename, updateData is called twice: first by updateTableData (correct columns/rows) and immediately again by the caller with the same columns/rows. The second call triggers a redundant context update and a redundant refreshProject. This is a latent correctness issue that should be resolved by removing the duplicate updateData calls from the callers and relying solely on updateTableData, or removing it from updateTableData and keeping it only in the callers.
- **[minor/quality]** `dataloom-frontend/src/Components/Table.tsx:135` useProjectContext() is cast as unknown as ProjectContextValue on line 135. This silences TypeScript rather than fixing it. Since ProjectContext is still a JS file the proper fix is to add a .d.ts or convert ProjectContext to TypeScript so the cast becomes unnecessary. The comment acknowledges this ("the context itself is still JS"), but leaving an as unknown as cast in a freshly migrated TypeScript file defeats part of the type-safety gain. At minimum, document this as a follow-up issue.
- **[minor/quality]** `dataloom-frontend/src/Components/Table.tsx:448` The onContextMenu handlers in the thead and tbody cast the React SyntheticEvent to MouseEvent via (e as unknown as MouseEvent). useContextMenu's open() should accept React.MouseEvent<Element> or the hook signature should be broadened. The double cast is a code smell that will hide type mismatches if the hook is refactored later.
- **[nit/quality]** `dataloom-frontend/src/Components/Table.tsx:557` DtypeBadge.jsx now accepts a className prop but the component file remains .jsx while Table is now .tsx. The className prop on DtypeBadge at line 557 of Table.tsx is passed as an empty string (className=""), which overrides the default "ml-1.5" but produces no visible class. This is intentional (to strip the margin in the dtype row), but an empty string as a prop value is confusing; a more explicit design would be to pass a dedicated prop like margin={false} or use undefined to fall back to the default.
- **[nit/style]** `dataloom-frontend/src/Components/Table.tsx:658` TablePagination.propTypes block is removed but the export comment "// TablePagination Component" on line 658 is still present and now refers to dead code artefact (no PropTypes). Minor cleanup: either remove the comment or fold it into a JSDoc block on the interface.

**Conventions**
- **PR title:** "Feat/tables UI" uses a slash instead of a colon and is vague — a Conventional Commits title would be "feat(table): spreadsheet-style frozen header, sticky serial column, and TS migration"
- **Description:** The description body is largely an unfilled template ("Brief summary of the changes") with no prose explanation of what changed or why; the screenshot is helpful but does not substitute for a written summary.
- **Commits (3):** clean, conventional history
- _Squash the three commits into one before merge (the three commits are logically a single coherent change), and replace the PR title with a proper Conventional Commits summary._

This PR migrates Table.jsx to TypeScript, adds a frozen/sticky header with a dedicated dtype sub-row, pins the serial-number column, and fixes the flex/overflow layout so the table fills the viewport correctly -- all solid UI improvements. The main correctness concern is that updateData is called twice per mutation operation (once inside updateTableData and again in each caller), causing a redundant context update and an extra refreshProject call on every add/delete/rename; this should be resolved by removing the duplicate calls. The TypeScript migration is mostly clean, but two as unknown as casts (on the context value and on the SyntheticEvent in onContextMenu) should be tracked as follow-up work rather than left indefinitely. PR title and description need to be updated to meet Conventional Commits standards before merging.

hanzalahwaheed:
@ivantha addressed the review suggestions. please check :)

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/Table.tsx:175` The refactored `updateTableData` no longer passes `resetColumnOrder: true` for add-column and delete-column operations. The original `handleAddColumn` and `handleDeleteColumn` each called `updateData` a second time with `{ resetColumnOrder: true }` after calling `updateTableData`. The new code consolidates into a single `updateTableData` call that only passes `{ dtypes: newDtypes }`, with a comment asserting that `updateData` auto-detects column-count changes to reset the order. If `ProjectContext.updateData` does not implement that implicit logic, column-reorder state will persist after structural add/delete operations, breaking the display.
- **[minor/quality]** `dataloom-frontend/src/Components/Table.tsx:135` `useProjectContext() as unknown as ProjectContextValue` uses a double cast to bypass TypeScript's type checker. The manually-declared `ProjectContextValue` interface can silently diverge from the actual JS context shape — any mismatch will only surface at runtime, not at compile time. Prefer exporting a typed interface from `ProjectContext` and importing it here.
- **[nit/style]** `dataloom-frontend/src/Components/Table.tsx:560` The `normalizeRows` helper is declared with the `function` keyword after `export default Table`. It is used inside the component above, so hoisting makes it work, but placing utilities before the component they serve is the conventional order and improves readability.

**Conventions**
- **PR title:** Title `Feat/tables UI` uses a capitalized word and slash separator instead of Conventional Commits form; should be `feat(table): ...` with a lower-case type and colon.
- **Description:** The body retains the template placeholder line `Brief summary of the changes. Reference any related issues.` verbatim; it links #393 and includes a screenshot but never explains what changed or why.
- **Commits (5):** clean, conventional history
- _Fix the PR title to conventional-commit format and replace the template text in the description body with an actual summary._

The PR migrates Table from JSX to TypeScript, adds a frozen serial-number column, moves dtype badges into a dedicated sub-header row, and eliminates duplicate `updateData` calls in all transform handlers. The main risk is in `updateTableData`: the explicit `resetColumnOrder: true` that `handleAddColumn` and `handleDeleteColumn` previously passed has been removed; confirm that `ProjectContext.updateData` automatically resets column order when the column count changes, or restore the flag explicitly. The double cast `as unknown as ProjectContextValue` is a workable transitional pattern but should be replaced by exporting a typed interface from the context module once that file is migrated. Five commit messages are well-formed; the PR title and description body need cleanup before merge.

hanzalahwaheed:
reg the findings:
1. the column reset logic is a bit messy atm. i will address this in a separate PR
2. migrating ProjectContext to TS will fix the issue. will address in a separate PR

ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `dataloom-frontend/src/Components/Table.tsx:160` updateTableData calls setData with Object.values(row) directly, but one line later passes the same rows to updateData via normalizeRows(rows). Object.values on an array is technically correct (returns index-keyed values), but using a different code path for the two callers of the same data is fragile and inconsistent. Replace the setData line with: setData(normalizeRows(rows).map((row, index) => [index + 1, ...row])) so both callers go through the same normalization.
- **[nit/quality]** `dataloom-frontend/src/Components/Table.tsx:135` useProjectContext() is cast with as unknown as ProjectContextValue to paper over the fact that the context module is still JavaScript. The comment acknowledges this, but the cast silently widens the escape hatch to the entire context object. A lighter mitigation is to type ProjectContext itself with a JSDoc @type annotation in context/ProjectContext.jsx, which would remove the need for the double cast and give the compiler real information.
- **[nit/style]** `dataloom-frontend/src/Components/Table.tsx:562` The normalizeRows helper is declared after export default Table. Even though function declarations are hoisted, placing a helper that is called inside the component body below the export default makes the file harder to read top-to-bottom. Move it above the Table component definition.

**Conventions**
- **PR title:** Title uses branch-name format ('Feat/tables UI') rather than Conventional Commits style; a title like 'feat(table): migrate to TypeScript and add spreadsheet layout' would be correct.
- **Description:** The body's opening line is the unedited template text 'Brief summary of the changes. Reference any related issues.' — no prose description of what changed or why is provided.
- **Commits (5):** The fifth commit ('chore: drop redundant casts in migrated Table component') is a direct cleanup of the first commit in the same PR and should have been squashed into it before opening the PR.
- _Squash the cast-cleanup commit, replace the template description with a short prose summary, and fix the PR title before merging._

The TypeScript migration is solid: all interfaces are well-documented, the type narrowing for context and API responses is precise, and dropping PropTypes in favour of TS interfaces is the right call. The main code-quality gap is in `updateTableData`, where `Object.values(row)` should be replaced with `normalizeRows` so the local `setData` path and the context `updateData` path use identical row normalization. Move `normalizeRows` above the component for readability, and address the double-cast on `useProjectContext` by adding JSDoc types to the context module. Fix the PR title, add a prose description, and squash the trailing cast-cleanup commit before merge.

hanzalahwaheed:
Addressed all changes except `useProjectContext() casting` whIch I will fix in a separate PR migrating the `ProjectContext.jsx` to Typescript.

PR #397, opened 2026-06-27
feat: unified /settings page consolidating account, preferences, and project settings

### LINKED ISSUE #392: [Feature] Unified Settings Experience for Account, Preferences, and Projects
## Is your feature request related to a problem?

Settings and configuration in DataLoom are currently scattered across multiple areas of the application with no single destination where users can manage everything.

Currently:

1. Project renaming is accessible through the project name in the workspace header/navbar.
2. Account management (email, password, account deletion) is available through a separate profile page accessed via the user avatar.
3. Project descriptions cannot be edited.
4. There is no dedicated home for future user preferences such as default page size, export defaults, themes, or other application-wide settings.

This makes the experience unintuitive, as users have no obvious answer to:

> "Where do I go to change my settings?"

As DataLoom grows and more configurable options are introduced, this fragmentation will become increasingly difficult to navigate and maintain.

## Describe the solution you'd like

Introduce a unified Settings experience that serves as the central location for all user, application, and project configuration.

Suggested structure:

```text
Settings
├── Account
├── Preferences
└── Projects
```

### Account

Contains all user account management functionality:

- Profile details
- Email management
- Password management
- Account deletion

### Preferences

Contains user-level preferences and defaults:

- Default page size
- Default export format
- Future user preferences

### Projects

Contains project-specific configuration:

- Project name
- Project description
- Future project-level preferences
- Project deletion / danger zone actions

Projects should be searchable/filterable to ensure the experience remains manageable for users with many projects.

A sidebar-based navigation pattern similar to Vercel would allow users to move between sections while maintaining a consistent mental model:

> If I want to configure something, I go to Settings.

## Describe alternatives you've considered

Keeping settings in their current locations and simply adding links between pages.

While this requires less initial work, it does not solve the discoverability problem, as users would still need to remember multiple locations for related configuration tasks.

## Additional context

- Existing account management functionality can be migrated into the new settings experience rather than rewritten.
- Project renaming already exists and can be moved into project settings.
- Project descriptions currently have no editable surface and would naturally belong in project settings.
- This creates a scalable foundation for future preferences and project configuration options while improving overall UX consistency.

SHIVAM-KUMAR-59:
## Description
Settings and configuration in DataLoom were scattered across multiple locations with no single destination for users to manage everything. This PR introduces a unified `/settings` experience inspired by Vercel's settings UX, organized into three sections accessible via a sidebar:

- **Account** — email, password, account deletion (migrated from `/profile`, which now redirects to `/settings/account`)
- **Preferences** — default page size and default export format, persisted to localStorage
- **Projects** — searchable project list with per-project settings (rename, description edit, delete)

Closes #392 

## Type of Change
- [x] New feature

## How Has This Been Tested?
- Navigated to `/settings` — redirects to `/settings/account` correctly
- Navigated to `/profile` — redirects to `/settings/account` correctly
- Updated email, password, deleted account — all work as before
- Saved preferences — page size and export format persist across reloads
- Browsed projects list, searched by name and description — works correctly
- Renamed project, updated description, deleted project from settings — all work
- Sidebar navigation between sections works correctly
- [x] Existing tests pass
- [x] Manual testing
- [x] Added new test cases 

## Screenshots
Screenshot 2026-06-27 at 14 51 45

Screenshot 2026-06-27 at 14 52 04

Screenshot 2026-06-27 at 14 52 29

Screenshot 2026-06-27 at 14 52 42

## Checklist
- [x] Tests pass locally


hanzalahwaheed:
hi @SHIVAM-KUMAR-59, fantastic work here! really appreciate it. I think we can polish this even more. here are some suggestions:

- new files should be written in TS as we are undergoing a migration.
- lets have a vertical list for `all projects` on the settings page rather than cards. I believe this UI would scale better.
- individual project settings can be routed to `/settings/projects/{projectId}`. I think this would be better in the longer run.
- I don't see a way to navigate back to the main dashboard here. Maybe we should add a back button somewhere.
- i think we can use a `Gear Icon` for settings instead of the product logo.

SHIVAM-KUMAR-59:
Thanks @hanzalahwaheed  for the detailed feedback! Addressing all points:

- TypeScript — will convert all new settings files to `.tsx`
- Vertical list for projects — agreed, will use a vertical project list in the sidebar itself
- Project settings route — already routed to `/settings/projects/:projectId`, matches your suggestion
- Back to dashboard — will add a "Back to Dashboard" link in the settings sidebar
- Gear icon — will swap the DataLoom logo for a Settings gear icon in the sidebar header

Will push the updated changes shortly.

ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `dataloom-frontend/src/App.jsx:40` The /profile route is deleted without adding a redirect to /settings/account. Any user visiting /profile via a bookmark, external link, or direct navigation lands on NotFoundPage. The PR description claims 'Navigated to /profile — redirects to /settings/account correctly' but no such redirect exists in the new route tree.
- **[major/bug]** `dataloom-frontend/src/pages/settings/SettingsProjectPage.tsx:88` After deleteProject succeeds, handleDeleteProject calls navigate('/settings/projects'), which is not a registered route (the router only knows /settings/projects/:projectId). The router falls through to NotFoundPage. Should navigate to /settings or /projects instead.
- **[minor/bug]** `dataloom-frontend/src/pages/settings/SettingsProjectPage.tsx:57` handleSave always passes both name: name.trim() and description: description.trim() to updateProject regardless of which field triggered the save. Clicking 'Save name' silently patches the description to whatever is currently in the textarea, and vice versa. Each save button should only send its own field (pass undefined or omit the other key).
- **[minor/bug]** `dataloom-frontend/src/pages/settings/SettingsPreferencesPage.tsx:17` EXPORT_FORMAT_OPTIONS includes tsv, json, xlsx, and parquet. If the ExportModal and backend do not currently support these formats, persisting one of them as defaultExportFormat in localStorage will cause ExportModal to open with a format value not present in its own option list, resulting in the browser silently falling back to the first option. Only formats the existing export pipeline actually handles should be offered here.
- **[nit/quality]** `dataloom-frontend/src/pages/settings/SettingsProjectPage.tsx:87` Commented-out call '// deleteProjectOrder(projectId);' is dead code left over from development and should be removed before merging.

**Conventions**
- **Commits (4):** The 4th commit mixes a TypeScript migration with review-feedback fixes and uses 'addressing the PR chan...' in its message; these changes should have been squashed into the relevant earlier commits for a clean history.
- _Squash the TS-migration/feedback commit into the earlier feature commits before merging._

The PR delivers a well-structured settings hub and a clean PATCH /projects/{id} backend endpoint, with a thorough description and screenshots. Three bugs need to be fixed before merging: the /profile route is silently dropped rather than redirected (users with existing bookmarks hit the 404 page despite the PR description claiming a redirect exists), post-deletion navigation targets the unregistered /settings/projects route (also landing on 404), and handleSave always patches both name and description together regardless of which save button was clicked. Remove the commented-out deleteProjectOrder call on line 87 as well.

SHIVAM-KUMAR-59:
**Addressed all findings:**
 
  - **[blocking/bug] `/profile` missing redirect** — added `<Route path="/profile" element={<Navigate to="/settings/account" replace / } / ` back into the route tree as a safety net for users with existing bookmarks.
 
  - **[major/bug] post-deletion navigating to unregistered `/settings/projects`** — fixed. Now navigates to `/settings` after project deletion.
 
  - **[minor/bug] `handleSave` patching both fields** — fixed. Now uses a conditional spread so clicking "Save name" only sends `{ name }` and "Save description" only sends `{ description }`.
 
  - **[minor/bug] unsupported export formats in preferences** — all 5 formats (`csv`, `tsv`, `json`, `xlsx`, `parquet`) are fully supported by both `ExportModal` and the backend export endpoint, so no change needed here.
 
  - **[nit/quality] commented-out `deleteProjectOrder` call** — removed.
 
  - **Commit history** — squashed 5 commits into 2 clean meaningful commits.
 
  Ready for re-review. @ivantha

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/App.jsx:41` The /profile backward-compatibility redirect is broken. In React Router v6, a child route with an absolute path (/profile) nested inside a parent route (/settings) is only evaluated after the parent path matches the URL prefix. Since /profile does not start with /settings, navigating to /profile bypasses the nested redirect entirely and falls through to the catch-all NotFoundPage. The redirect should be moved as a sibling of the /settings route, not a child.
- **[minor/quality]** `dataloom-backend/app/api/endpoints/projects.py:455` The new PATCH endpoint reuses RenameProjectResponse (fields: project_id, filename, file_path) even though it also updates the description. A description-only update returns a response body that contains no confirmation of the new description value, forcing the client to infer success solely from the HTTP 200 status. A dedicated response schema or adding description to RenameProjectResponse would make the contract explicit.
- **[nit/quality]** `dataloom-backend/app/services/project_service.py:338` UpdateProjectRequest allows both name and description to be None (empty body {}). The update_project service then calls db.add, db.commit, and db.refresh with no field changes, wasting a DB round-trip. A model_validator that raises if both fields are None would prevent no-op commits at the schema layer.
- **[minor/quality]** `dataloom-frontend/src/pages/settings/SettingsProjectPage.tsx:31` SettingsProjectPage calls getProjectDetails(projectId, 1, 50), which fetches paginated row data, column names, dtypes, and pagination metadata, just to populate a name and description form. This is a heavy call for a lightweight need. Using a project-metadata-only endpoint (or deriving it from the sidebar project list already fetched in SettingsPage) would avoid fetching irrelevant row data.

**Conventions**
- **Commits (2):** The second commit message ('feat: migrate settings files to typescript and addressing the PR chan…') mixes two unrelated concerns (TS migration and PR feedback), uses non-imperative mood ('addressing'), and encodes iteration noise that should be squashed before merge.
- _Squash the two commits into one clean, imperative-mood commit before merging._

The unified settings page is well-structured and the PR description is thorough with screenshots and a complete manual-testing checklist. However, the /profile backward-compatibility redirect is broken: nesting an absolute-path child route (/profile) inside the /settings parent in React Router v6 means the router never evaluates it when the URL is /profile, sending users with old bookmarks to the 404 page instead. The new PATCH endpoint reuses RenameProjectResponse, which has no description field, so a description-only update returns no confirmation of the new value in the response body. Two smaller items: SettingsProjectPage fetches full paginated row data via getProjectDetails just to display a name/description form, and UpdateProjectRequest should require at least one non-None field to prevent no-op DB commits.

SHIVAM-KUMAR-59:
**Addressed all findings:**

 - **[major/bug] `/profile` redirect broken in React Router v6** — fixed. Moved the redirect to be a sibling of `/settings` rather than a nested child, so it's evaluated independently for the `/profile` URL.

 - **[minor/quality] `RenameProjectResponse` missing `description`** — fixed. Added a new `UpdateProjectResponse` schema with `description` included, so the client gets explicit confirmation of the updated value.

 - **[nit/quality] `UpdateProjectRequest` allowing empty-body no-op commits** — fixed. Added a `model_validator` that rejects requests where both `name` and `description` are `None`.

 - **[minor/quality] `SettingsProjectPage` fetching full paginated row data** — confirmed, leaving as-is for this PR and will file a follow-up issue for a dedicated project-metadata-only endpoint. Not addressing here to keep this PR focused on the settings consolidation itself rather than introducing a new endpoint.
 
 - **Commit history** — squashed both commits into one clean, imperative-mood commit.

 Ready for re-review. @ivantha

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/pages/settings/SettingsPreferencesPage.tsx:20` Saving the 'Default page size' preference only writes to localStorage. ProjectContext wraps BrowserRouter and stays mounted for the whole SPA session; it reads that localStorage key just once at provider-mount time and otherwise keeps pageSize in its own state (which is further overwritten by data.page_size on every refreshProject response), so a newly saved default has no effect until the user does a full page reload, contrary to what a live 'Save preferences' action implies. _(unverified)_
- **[major/quality]** `dataloom-backend/app/services/project_service.py:339` update_project() re-implements the exact same trim/empty-check/commit logic already in the existing rename_project() (used by the pre-existing PATCH /projects/{id}/rename endpoint), giving the API two independently-maintained code paths for renaming a project. The name-handling branch here should delegate to rename_project() instead of duplicating its validation, or the two endpoints should be consolidated.
- **[major/test]** `dataloom-backend/app/api/endpoints/projects.py:459` The new PATCH /projects/{project_id} endpoint and its update_project() service function ship with no automated test coverage, unlike sibling project-mutation endpoints (rename, save, search) which each have a dedicated tests/test_*.py file. Edge cases such as the 'at least one field required' validator and the empty-name-after-trim check are untested.
- **[minor/quality]** `dataloom-frontend/src/pages/settings/SettingsProjectPage.tsx:71` handleSave and handleDeleteProject use bare `catch {}` blocks that always show a fixed generic toast, discarding the backend's actual error detail (e.g. a 422 from name/description validation) — inconsistent with the getErrorMessage() pattern used in SettingsAccountPage.tsx added in the same PR. The project-name input also has no maxLength matching the backend's 255-char cap, so hitting that limit just produces an unhelpful generic failure toast.
- **[minor/quality]** `dataloom-frontend/src/pages/settings/SettingsPage.tsx:36` The sidebar's 'Projects' list calls getRecentProjects(), which the backend route hard-caps at 10 results. Without typing a matching search query, users can only reach their 10 most-recently-modified projects from Settings, not the full project list that a general per-project settings browser implies.
- **[nit/style]** `dataloom-frontend/package-lock.json:1192` This lockfile diff removes `libc` metadata from roughly 20 optional-dependency entries throughout the file, unrelated to the settings feature — looks like incidental churn from a different local npm version rather than an intended change. Regenerating with the project's pinned npm version would keep the diff focused on the feature.

**Conventions**

This PR consolidates account, preference, and per-project settings into a single /settings experience with sidebar navigation, replacing the old /profile page (which now redirects) and adding a new PATCH /projects/{id} endpoint for name/description updates. The backend addition correctly reuses the existing ownership-scoped get_project_or_404 dependency, but it duplicates validation logic already present in the existing rename endpoint and ships without any tests. On the frontend, the new 'default page size' preference is saved to localStorage but isn't actually picked up by the already-running ProjectContext until a full reload, and the Settings project browser is implicitly capped at the 10 most-recently-modified projects unless the user searches. These points are worth a revision pass before merge; the rest of the PR (routing, account/preferences UI, and the PR description itself) is solid and well-documented.

SHIVAM-KUMAR-59:
Addressed all findings:

- **[major/bug] src/pages/settings/SettingsPreferencesPage.tsx** — fixed. Saving preferences now updates the in-memory page size state exposed by ProjectContext in addition to persisting it to localStorage, so the new default page size takes effect immediately without requiring a full page reload.

- **[major/quality] app/services/project_service.py** — fixed. Removed the duplicated project-name validation logic by extracting the shared trim/empty-name validation into a single helper used by both `rename_project()` and `update_project()`, keeping the behavior consistent across both endpoints.

- **[major/test] app/api/endpoints/projects.py** — fixed. Added dedicated service and endpoint tests covering the new `PATCH /projects/{project_id}` endpoint, including the "at least one field required" validator, empty-name-after-trim validation, name-only updates, description-only updates, combined updates, and authentication checks.

- **Formatting**: While running the project's formatter, it automatically reformatted the following files:
    - `src/hooks/__tests__/useDatasetSummary.test.ts`
    - `src/context/WorkspaceTabsContext.tsx`

These contain formatter-only changes; I did not make any functional modifications to them.

Commit history — squashed into a single clean commit.

Ready for re-review. @ivantha

ivantha:
### Automated review summary

**Findings**
- **[major/quality]** `dataloom-frontend/src/pages/settings/SettingsProjectPage.tsx:30` fetchProject calls getProjectDetails(projectId, 1, 50), which hits GET /projects/get/{id} and reads the entire underlying file into a DataFrame via read_table_safe before pagination is applied in memory — just to populate two text inputs (name, description). For large datasets this makes opening a project's settings page unnecessarily slow/expensive, when the existing lightweight shape already used by /projects/recent and /projects/search (project_id, name, description, last_modified) has everything this page actually needs.
- **[major/test]** `dataloom-frontend/src/pages/settings/SettingsPage.tsx:1` None of the new Settings components (SettingsPage's project search/sidebar, SettingsPreferencesPage, SettingsProjectPage's inline name/description editors) ship with any frontend tests, even though the repo has Vitest + Testing Library and tests similarly-sized components elsewhere. Only the backend PATCH endpoint got test coverage, despite the PR checklist claiming new test cases were added.
- **[minor/bug]** `dataloom-frontend/src/pages/settings/SettingsPage.tsx:58` The debounced project-search effect has no guard against out-of-order responses: setSearchResults(response) applies whichever request resolves last. If an earlier (slower) keystroke's search resolves after a newer (faster) one, the sidebar can end up showing results that no longer match the current search box text.
- **[minor/quality]** `dataloom-frontend/src/pages/settings/SettingsPage.tsx:36` The Projects sidebar's default (non-search) view is populated via getRecentProjects(), which the backend hardcodes to the 10 most-recently-modified projects with no pagination or 'view all' option, so a user with more than 10 projects can't browse to older ones without already knowing a search term that matches.
- **[nit/quality]** `dataloom-frontend/src/context/ProjectContext.jsx:86` updatePageSizePreference manually calls localStorage.setItem('pageSize', ...), but the existing useEffect keyed on [pageSize] immediately above it already persists every pageSize change (including this one, since the function calls setPageSize first) — the manual write is redundant.
- **[nit/quality]** `dataloom-frontend/src/pages/settings/SettingsPreferencesPage.tsx:22` localStorage.getItem/setItem calls here (and in the exportFormat initializer) aren't wrapped in try/catch, unlike every other localStorage access in this codebase (ProjectContext.jsx), which deliberately fails silently on storage errors. If storage throws, handleSave throws uncaught and 'Preferences saved' silently never shows.
- **[nit/quality]** `dataloom-frontend/src/pages/settings/SettingsAccountPage.tsx:134` handleDeleteAccount's success path no longer shows a confirmation toast or resets the delete-modal/password state before navigating away, unlike the ProfilePage.jsx version it replaces. Harmless since navigate() unmounts the page, but it's a silent loss of explicit success feedback.
- **[nit/style]** `dataloom-frontend/src/Components/workspace/features/transforms.tsx:53` A large portion of this diff (this file, ColumnProfileCard.tsx, DatasetSummaryPanel.tsx, history.tsx, profiling.tsx, WorkspaceTabsContext.tsx, FillEmptyForm.tsx, two pre-existing test files, plus ~100 lines of package-lock.json libc-metadata churn) is pure reformatting with no behavior change and no relation to the settings feature, inflating the diff and adding avoidable merge-conflict surface.

**Conventions**
- _Hygiene is good overall; the one improvement is splitting the unrelated Prettier reformatting and package-lock.json churn out of this commit so the feature diff stays focused and easy to review._

This is a well-scoped feature that migrates account management into a proper Settings section and adds project metadata editing behind a correctly ownership-checked PATCH endpoint, backed by decent backend tests and a clear PR description. Two things are worth addressing before merge: SettingsProjectPage loads a project's full paginated dataset just to show its name and description, which is unnecessarily expensive for large files, and the new frontend surface (search, preferences, metadata editing) has no test coverage despite the PR checklist claiming otherwise. The diff also carries a sizable amount of unrelated Prettier reformatting and package-lock.json churn that should be split out to keep the change easy to review. A few smaller polish items (a stale-response race in the debounced search, a redundant localStorage write, a dropped success toast on account deletion) are worth quick follow-ups but aren't blocking.

SHIVAM-KUMAR-59:
Fixed:

 - **[major/quality] heavy fetch for name/description** — new `GET /projects/{id}/meta` endpoint returns only `project_id`, `name`, `description`, `last_modified`. `SettingsProjectPage` now calls this instead of `getProjectDetails`.

 - **[major/test] no frontend tests** — added Vitest tests for `SettingsPreferencesPage` (localStorage read/write, save button) and `SettingsProjectPage` (API called with correct partial payload per save button).

 - **[minor/bug] out-of-order search responses** — added `cancelled` flag, all state updates guarded, cleanup cancels both timeout and in-flight result application.

 - **[minor/quality] sidebar capped at 10** — deferred, same as prior two rounds. Filing separate issue.

 Ready for re-review. @ivantha

ivantha:
### Automated review summary

**Findings**
- **[major/quality]** `dataloom-frontend/src/pages/settings/SettingsPage.tsx:36` The Settings sidebar's 'Projects' list (non-search state) is populated via getRecentProjects(), backed by GET /projects/recent which is hardcoded server-side to the 10 most-recently-modified projects (projects.py line 127, limit=10). Despite the PR description advertising a 'searchable project list,' a user with more than 10 projects can only reach an older project's settings by already knowing its name/description well enough to search for it — there is no complete or paginated project listing wired into this view.
- **[minor/quality]** `dataloom-frontend/src/constants/routes.js:28` ROUTES.settingsAccount, settingsPreferences, and settingsProject are added but never imported/used anywhere; App.jsx (Route path/index elements) and SettingsPage.tsx (NAV_ITEMS, project NavLink) hardcode the identical '/settings/account', '/settings/preferences', and '/settings/projects/:id' strings as literals instead. This breaks the ROUTES.xxx convention used everywhere else in the app (e.g. ROUTES.home in SignInPage/SignUpPage/NotFoundPage) and leaves two sources of truth for the same paths.
- **[minor/test]** `dataloom-backend/app/api/endpoints/profiling.py:97` The new GET /{project_id}/meta endpoint has zero automated test coverage. Manually verified it works correctly (200 with the right fields for the owner, 401 when unauthenticated), but no test in the diff exercises it — SettingsProjectPage.test.tsx only mocks the frontend API wrapper. Contrast with the sibling PATCH /{project_id} endpoint, which got 9 passing tests in test_update_project.py.
- **[minor/quality]** `dataloom-backend/app/api/endpoints/profiling.py:97` The new /meta route is added to profiling.py, whose module docstring scopes the file to exactly three descriptive, working-copy-reading routes ('quality assessment is out of scope here'); /meta doesn't touch the dataframe at all and is plain project-metadata CRUD that fits better beside update_project_endpoint in projects.py. It also reuses schemas.LastResponse, whose own docstring says 'Response for recently modified projects' — the fields happen to line up but the name is misleading for a single-project fetch in the generated API docs.
- **[nit/quality]** `dataloom-frontend/src/context/ProjectContext.jsx:86` updatePageSizePreference manually writes pageSize to localStorage, but the existing effect a few lines above already persists pageSize to localStorage on every change via setPageSize — the manual write is redundant (harmless, just dead duplication).
- **[nit/quality]** `dataloom-frontend/src/pages/settings/types.ts:21` The exported Toast interface is never imported or used anywhere in the codebase (showToast/useToast come from the untyped ToastContext.jsx) — dead type.

**Conventions**

This PR replaces the old /profile page with a unified /settings experience (account, preferences, per-project settings) and adds a matching PATCH /{project_id} endpoint with solid, passing test coverage; backend lint/format and the full new test suite (verified locally) pass, and the /profile route migration leaves no dangling references. A few gaps worth closing before merge: the new GET /{project_id}/meta endpoint (used to load a project's settings) has no test coverage and is arguably misplaced in profiling.py under a reused, misleadingly-named schema; the new ROUTES.settingsAccount/settingsPreferences/settingsProject constants go unused because App.jsx and SettingsPage.tsx hardcode the same paths as literals instead; and the Settings sidebar's project list is capped at the 10 most-recently-modified projects (via the existing /projects/recent endpoint), so browsing to an older project's settings without already knowing its name won't work once a user has more than 10 projects.

SHIVAM-KUMAR-59:
Fixed:

* **[major/quality]** Settings sidebar no longer relies on `GET /projects/recent`. Added a paginated `GET /projects` endpoint and updated the sidebar to use it, allowing access to the complete project list instead of only the 10 most recently modified projects.
* **[minor/quality]** Replaced hardcoded settings route strings with the existing `ROUTES` constants to remove duplicate sources of truth.
* **[nit/quality]** Removed the unused `Toast` interface from `types.ts`.

Ready for re-review. @ivantha

PR #400, opened 2026-06-29
feat: add structured JSON logging with request correlation

shivv23:
## Summary

Replace the plain-text root-logger handler with a JSON formatter on the `app` logger so that all `app.*` module loggers emit machine-parseable JSON lines. Add a `request_id` context variable and a FastAPI middleware that correlates every log entry emitted during an HTTP request under a single UUID.

## Key changes

- **`JSONFormatter`** — stdlib-`logging` `Formatter` subclass that serialises each record as a JSON object with `timestamp`, `level`, `logger`, and `message` keys, plus an optional `request_id` from the execution context and an `exception` key for tracebacks.
- **`request_id_var`** — a `contextvars.ContextVar` that carries the correlation ID across async boundaries within the same request.
- **`setup_logging()`** — now configures the `app` logger with the JSON handler and sets `propagate=False` to avoid duplicate output when uvicorn has already installed a root-level handler.
- **`add_request_id` middleware** — reads or generates a request ID, sets it on the context var, measures request duration, and echoes the ID back in the `X-Request-ID` response header.
- **Zero new runtime dependencies** — uses only Python stdlib `logging`, `contextvars`, `json`, `uuid`, and `time`.

## Verification

- All 458 existing backend tests pass (no regressions).
- Ruff lint + format pass clean.
- Zero new dependencies added.
- Two files changed (+82, -17).
- Startup/shutdown logs (outside any HTTP request) correctly omit the `request_id` field.

### DIFF
--- a/dataloom-backend/app/main.py
+++ b/dataloom-backend/app/main.py
@@ -79,11 +79,13 @@ async def add_request_id(request: Request, call_next):
     response.headers["X-Request-ID"] = rid
 
     logger.info(
-        "request complete (method=%s path=%s status=%d duration_ms=%.2f)",
-        request.method,
-        request.url.path,
-        response.status_code,
-        duration_ms,
+        "request complete",
+        extra={
+            "method": request.method,
+            "path": request.url.path,
+            "status_code": response.status_code,
+            "duration_ms": duration_ms,
+        },
     )
     return response
 
--- a/dataloom-backend/app/main.py
+++ b/dataloom-backend/app/main.py
@@ -3,6 +3,8 @@
 Configures middleware, exception handlers, and mounts all API routers.
 """
 
+import time
+import uuid
 from contextlib import asynccontextmanager
 from pathlib import Path
 
@@ -15,7 +17,7 @@ from app.config import get_settings
 from app.database import verify_database_connection
 from app.exceptions import AppException, app_exception_handler
 from app.services.transformation_service import TransformationError
-from app.utils.logging import get_logger, setup_logging
+from app.utils.logging import get_logger, request_id_var, setup_logging
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- **[minor/security]** `dataloom-backend/app/main.py:72` Client-supplied X-Request-ID is accepted without a length cap, then echoed into the response header and written to the log on every request. A caller can send a value up to the HTTP header limit (~8 KB) to inflate log volume on every request.
- **[minor/quality]** `dataloom-backend/app/main.py:76` No try/finally around call_next. If the inner call raises (e.g., async cancellation or a bug in a deeper middleware), the 'request complete' log entry is never emitted, silently dropping the access-log record for that request.
- **[nit/quality]** `dataloom-backend/app/utils/logging.py:17` Python 3.12 adds 'taskName' automatically to every LogRecord.__dict__. Because 'taskName' is absent from _STANDARD_ATTRS, it will appear as an unexpected top-level key in every JSON log line on Python 3.12+.

**Conventions**
- **Commits (2):** The second commit ('fix(logging): pass structured request fields via Logger.extra') corrects the first commit within the same PR; it should be squashed into the feat commit before merge so history is one logical unit.
- _Squash the two commits before merging._

The PR is a clean, zero-dependency infrastructure addition that upgrades the backend from plain-text root-logger output to scoped JSON logging with per-request correlation IDs. The formatter, ContextVar usage, and middleware logic are all idiomatic. Three issues need attention before merge: the client-controlled X-Request-ID value should be length-capped (e.g., 128 chars) before being echoed into logs and the response header; the middleware should wrap call_next in a try/finally so that failed requests still produce an access-log entry; and 'taskName' must be added to _STANDARD_ATTRS to avoid leaking an unexpected field into every JSON log line on Python 3.12+. The two-commit history should be squashed into a single feat commit.

PR #402, opened 2026-06-30
refactor: extract transformation registry into dedicated module

shivv23:
## Summary

Split the 1016-line `transformation_service.py` god-module into two focused modules:

- **`app/services/transformation_service.py`** -- pure DataFrame transformation functions only (filter, sort, pivot, melt, groupby, etc.)
- **`app/services/transforms_registry.py`** (new) -- `TransformationError`, `TransformationSpec`, `TRANSFORMATION_REGISTRY`, `resolve_transformation`, and `apply_logged_transformation`

## Design decisions

- **Zero public API breakage.** `transformation_service.py` re-exports `TransformationError`, `TRANSFORMATION_REGISTRY`, `apply_logged_transformation`, and `resolve_transformation` so all existing import paths continue to work unchanged.
- **No circular imports.** `resolve_transformation` uses a lazy function-level `import` to look up functions on `transformation_service`. This avoids a module-level circular dependency while preserving the ability for tests to patch functions on `transformation_service` (both `globals()` lookup and `getattr(module, name)` see the same patched module object).
- **Import-time integrity check preserved.** The `_missing` guard that ensures every `OperationType` enum member has a `TRANSFORMATION_REGISTRY` entry still runs at import time.
- **Only 2 files changed.** No tests, config, or unrelated code touched.

## Verification

- All 458 backend tests pass (no regressions).
- `ruff check` passes clean.
- `ruff format --check` passes clean.
- +287, -261 net -- the new module absorbs the registry weight and the original file shrinks by 25%.


hanzalahwaheed:
@ivantha I think the service in intentionally kept together right now. please check.

shivv23:
@hanzalahwaheed Hi! You are right about this, thanks for pointing it out.
I just ignored the design intent. Closing this PR. Thanks!

PR #404, opened 2026-07-01
feat: add rate limiting to auth endpoints with sliding-window counter

shivv23:
## Summary

Adds an in-memory sliding-window rate limiter to the four auth endpoints that accept credentials or tokens: `/auth/signin`, `/auth/signup`, `/auth/forgot-password`, and `/auth/reset-password`.

## Implementation

- **`app/utils/rate_limiter.py`** — `RateLimiter` class using `collections.deque` + `threading.Lock`. Default: 20 requests per 5-minute window per client IP. Time function is injectable for deterministic testing.
- **`app/api/dependencies.py`** — `rate_limit()` FastAPI dependency. Uses `request.client.host` (direct TCP connection, never client-supplied headers). Returns 429 with `Retry-After` header.
- **`app/config.py`** — Three new settings: `rate_limit_enabled`, `rate_limit_max_requests`, `rate_limit_window_seconds`.
- **`app/api/endpoints/auth.py`** — `Depends(rate_limit)` added to the four rate-limited endpoints.

## Testing

- 12 unit tests for the `RateLimiter` class: under-limit, at-limit, window expiry, partial slide, multiple keys, reset, invalid construction, and thread-safety (200 concurrent requests).
- 3 sweep tests verifying periodic key expiry prevents unbounded store growth.
- 1 integration test exercising the FastAPI dependency end-to-end (429 + Retry-After header).
- Existing test suite (474 tests) passes with `RATE_LIMIT_ENABLED=false` set in conftest.py.

## Design decisions

- Sliding-window counter (not fixed window) to avoid boundary bursts.
- Thread-safe via `threading.Lock` — single-process in-memory; production deployments behind multiple workers should substitute a shared store (Redis).
- 20 req / 5 min is conservative for interactive usage; tune via env vars.

### DIFF
--- a/dataloom-backend/app/api/dependencies.py
+++ b/dataloom-backend/app/api/dependencies.py
@@ -1,15 +1,56 @@
 """Shared FastAPI dependencies for endpoint functions."""
 
+import math
+import threading
 import uuid
 
-from fastapi import Cookie, Depends, HTTPException
+from fastapi import Cookie, Depends, HTTPException, Request
 from sqlmodel import Session
 
 from app import database, models
+from app.config import get_settings
 from app.services import auth_service
 from app.utils.logging import get_logger
+from app.utils.rate_limiter import RateLimiter
 
 logger = get_logger(__name__)
+
+_limiter: RateLimiter | None = None
+_limiter_lock = threading.Lock()
+
+
+def rate_limit(
+    request: Request,
+) -> None:
+    """FastAPI dependency that rate-limits requests by the direct TCP client IP.
+
+    The client IP is always taken from ``request.client.host`` (the direct TCP
+    connection) and never from client-supplied headers.  Operators behind a
+    reverse proxy must strip or overwrite the peer IP at the proxy layer.
+
+    Returns 429 with a ``Retry-After`` header when the limit is exceeded.
+    """
+    settings = get_settings()
+    if not settings.rate_limit_enabled:
+        return
+
+    global _limiter
+    if _limiter is None:
+        with _limiter_lock:
+            if _limiter is None:
+                _limiter = RateLimiter(
+                    settings.rate_limit_max_requests,
+                    settings.rate_limit_window_seconds,
+                )
+
+    client_ip = request.client.host if request.client is not None else "unknown"
+    allowed, retry_after = _limiter.check(client_ip)
+    if not allowed:
+        raise HTTPException(
+            status_code=429,
+            detail=f"Too many requests. Try again in {math.ceil(retry_after)} seconds.",
+            headers={"Retry-After": str(math.ceil(retry_after))},
+        )
 
 
 def get_current_user(
--- a/dataloom-backend/app/api/endpoints/auth.py
+++ b/dataloom-backend/app/api/endpoints/auth.py
@@ -5,7 +5,7 @@ from sqlalchemy.exc import SQLAlchemyError
 from sqlmodel import Session
 
 from app import database, models, schemas
-from app.api.dependencies import get_current_user
+from app.api.dependencies import get_current_user, rate_limit
 from app.config import get_settings
 from app.services import auth_service
 from app.utils.logging import get_logger
@@ -32,7 +32,12 @@ def _set_auth_cookie(response: Response, token: str) -> None:
 
 
 @router.post("/signup", response_model=schemas.UserResponse, status_code=201)
-def signup(payload: schemas.SignupRequest, response: Response, db: Session = Depends(database.get_db)):
+def signup(
+    payload: schemas.SignupRequest,
+    response: Response,
+    db: Session = Depends(database.get_db),
+    _rate_limit: None = Depends(rate_limit),
+):
     """Register a new user and issue an auth cookie."""
     if auth_service.get_user_by_email(db, payload.email) is not None:
         raise HTTPException(status_code=409, detail="An account with this email already exists")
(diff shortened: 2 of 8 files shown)


ivantha:
### Automated review summary

**Findings**
- **[blocking/security]** `dataloom-backend/app/api/dependencies.py:39` rate_limit() derives the rate-limit key from the client-supplied X-Forwarded-For (or X-Real-IP) header unconditionally, with no check that the request actually passed through a trusted proxy that sets/overwrites it. Any caller can bypass the limiter entirely by sending a different X-Forwarded-For value on every request, which defeats the brute-force/credential-stuffing protection this PR exists to add to signup/signin/forgot-password/reset-password.
- **[major/security]** `dataloom-backend/app/utils/rate_limiter.py:34` RateLimiter._store entries are only ever removed via the unused reset() method (never called from request-handling code), so once a key is seen it lives forever. Combined with the spoofable IP key above, an attacker who varies X-Forwarded-For per request can grow _store without bound over time — there is no TTL, LRU cap, or periodic sweep.
- **[major/test]** `dataloom-backend/tests/test_auth.py:1` No test exercises the rate_limit() dependency itself: the header-trust/IP-derivation logic and a live 429 + Retry-After response are completely uncovered. test_auth.py (which already uses a TestClient against the real endpoints) wasn't touched, the whole suite runs with RATE_LIMIT_ENABLED=false (tests/conftest.py), and the new tests only unit-test the standalone RateLimiter class in isolation.
- **[minor/bug]** `dataloom-backend/app/api/dependencies.py:50` int(retry_after) truncates toward zero instead of rounding up. When the true wait is sub-one-second (e.g. 0.4s), this yields `Retry-After: 0` and "Try again in 0 seconds" while the caller is still inside the rate-limit window, so an immediate retry following that header is rejected again.
- **[minor/bug]** `dataloom-backend/app/api/dependencies.py:33` Lazy singleton init of the module-level `_limiter` (`if _limiter is None: _limiter = RateLimiter(...)`) is not guarded by a lock, unlike RateLimiter's own carefully-locked internal state. Concurrent first requests can race to construct two separate limiter instances; whichever loses is discarded along with any counts it already recorded.
- **[minor/quality]** `dataloom-backend/app/main.py:72` The client-supplied X-Request-ID header is trusted verbatim (only whitespace-stripped) with no length cap or character validation before being used as the correlation ID stamped into every log line for the request and echoed back in the response header.

**Conventions**
- **PR title:** Names only the rate-limiting feature; doesn't reflect that the diff also rewrites logging.py (new JSONFormatter, request-id ContextVar) and adds request-id middleware to main.py — a second, unrelated feature bundled into the same PR.
- **Description:** No linked issue (no 'Closes #...'), and every section (Summary/Implementation/Testing/Design decisions) covers only the rate limiter — the JSONFormatter/request-id-middleware changes aren't mentioned anywhere.
- **Commits (3):** 2 of the 3 commits ('add structured JSON logging with request correlation', 'fix(logging): pass structured request fields via Logger.extra') implement an unrelated logging/observability feature not referenced in the PR title or description; only the third commit matches what the PR claims to do.
- _Split this into two PRs (auth rate limiting vs. structured logging/request correlation) and update the description to link the tracking issue and cover the full diff._

PR #404 bundles two unrelated changes — a sliding-window rate limiter on the four public auth endpoints, and an undisclosed structured-JSON-logging/request-correlation rewrite of logging.py and main.py — neither the title nor description mentions the latter. The rate limiter's client-IP derivation trusts the client-supplied X-Forwarded-For/X-Real-IP headers unconditionally, so it can be bypassed by rotating that header per request, undermining the feature's stated purpose and also letting an attacker grow the limiter's in-memory store without bound. The RateLimiter class itself is well unit-tested (12 tests, including concurrency), but nothing tests the actual FastAPI dependency's header-trust logic or asserts a live 429/Retry-After response end-to-end. Recommend fixing (or explicitly documenting/enforcing) the trusted-proxy assumption, adding an integration test for the dependency, fixing the Retry-After truncation, and splitting the logging changes into their own PR.

ivantha:
### Automated review summary

**Findings**
- **[minor/test]** `dataloom-backend/tests/test_auth.py:353` test_rate_limit_http_429 only sends requests to POST /auth/signup. Depends(rate_limit) was also added to /auth/signin, /auth/forgot-password, and /auth/reset-password in auth.py, but no test exercises the 429 path on those three routes, so a broken or missing dependency wire-up there would not be caught by CI.
- **[minor/other]** `dataloom-frontend/package-lock.json:1190` This is a backend-only feature (auth rate limiting) but the PR also rewrites large parts of dataloom-frontend/package-lock.json, stripping 'libc' metadata from dozens of optional-dependency entries. This looks like incidental lockfile churn from a different local npm version rather than an intentional change, and adds unrelated diff noise and merge-conflict risk to an otherwise focused backend PR.
- **[minor/quality]** `dataloom-backend/.env.example:44` The three new settings added in app/config.py (rate_limit_enabled, rate_limit_max_requests, rate_limit_window_seconds at lines 49-51) are not added to .env.example, breaking this file's existing pattern of documenting every Settings field for operators.
- **[minor/quality]** `dataloom-backend/app/api/dependencies.py:47` _limiter.check(client_ip) keys only on client IP with no per-endpoint dimension, so signup, signin, forgot-password, and reset-password all draw from one shared 20-requests-per-5-minutes budget per IP instead of independent budgets. Worth confirming this pooled behavior (versus per-endpoint limits) is intentional.
- **[nit/quality]** `dataloom-backend/app/utils/rate_limiter.py:66` self._store.pop(key, None) immediately before self._store[key] = deque([now]) on the next line is dead code: the unconditional reassignment on the following line makes the pop a no-op.

**Conventions**

This PR adds a thread-safe, stdlib-only sliding-window rate limiter to the four unauthenticated auth endpoints (signup, signin, forgot-password, reset-password), backed by solid unit tests including a concurrency test that pins exact behavior under 200 simultaneous requests. The main gap is integration coverage: the new end-to-end 429 test only exercises /auth/signup, so dependency wiring on the other three routes is unverified by CI. Two small cleanups are worth doing before merge: document the three new RATE_LIMIT_* settings in .env.example, and drop the unrelated dataloom-frontend/package-lock.json diff, which looks like incidental lockfile churn. Core logic is otherwise sound (bounded memory via periodic sweep, no trust in spoofable headers), making this a solid, low-risk security improvement.

PR #401, opened 2026-06-29
feat: revamp layout - add tabs and sidepanel

hanzalahwaheed:
## Description

Revamps the workspace into a **tabbed layout with docked side panels**, and adds a
**feature registry** so new features slot in through one seam instead of edits
scattered across the codebase.

- **Tabs** — a VS Code-style workspace tab strip; every tab type (DataSet, Logs,
  Checkpoints, Summary) resolves through a tab registry.
- **Side panel** — transform forms and other feature panels dock in a right-side
  panel (`RightPanel`, via `PanelContext`) instead of floating.
- **Data profiling** re-homed into the workspace: a **Summary** tab and an inline
  **Column Profiles** toggle under a new Profiling menu.
- **History sync** — Logs/Checkpoints refresh via `HistoryRefreshContext`, split
  into stable actions vs. changing tokens so action-only callers don't re-render;
  refresh is dispatched from the component layer, keeping the API layer
  side-effect free.
- **Feature registry** — `registerFeature({ tabs, panels, menu })` lets a feature
  declare its tabs, docked panels, and ribbon menu in one self-registering adapter
  (`workspace/features/*`). Adding a feature is now one file + one import rather
  than edits across the tab registry, a panel map, the menu, and DataScreen — this
  is what delivers the "easy to add features" goal.

Fixes #398

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass (129 Vitest tests, lint clean)
- [X] Tests updated (tab-registry test repointed to the feature adapters)
- [X] Manual testing

## Screenshots
image

## Checklist

- [X] My code follows the project's style guidelines
- [X] I have performed a self-review
- [X] I have added/updated documentation as needed
- [X] My changes generate no new warnings
- [X] Tests pass locally

PR #407, opened 2026-07-01
feat: add data visualization (Charts feature)

### LINKED ISSUE #399: [Feature] Graph based Visualisations
- Add charts/graphs for better dataset representation.

hanzalahwaheed:
## Description

Adds **data visualization** as a Charts feature. Opening **Profiling ▸ Charts**
gives a Charts tab (rendered charts + auto-suggestions) with a docked no-code
**chart builder** in the right side panel. Six chart types — histogram, bar,
line, area, scatter, pie — plus dtype-driven suggestions and an embedded
correlation heatmap. The backend computes aggregated chart data (so payloads stay
bounded regardless of dataset size); the frontend renders with Recharts
(lazy-loaded into its own chunk).

Built on the tabbed-workspace **feature registry** from #401: the whole feature is
a single `workspace/features/charts.tsx` adapter declaring its tab, its docked
panel, and its menu entry.

> **Depends on #401.** This branch is stacked on `feat/tabs-layout`, so the diff
> currently includes that PR's registry commit. Once #401 merges I'll rebase, and
> this reduces to just the Charts commit.

Fixes #399

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [X] Existing tests pass (backend 498, frontend 153)
- [X] New tests added (visualization service + endpoints; chart-builder panel, chart-suggestion and correlation hooks, correlation heatmap)
- [X] Manual testing

_Charts tab with the docked chart builder — to be attached._

## Checklist

- [X] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/dependencies.py
+++ b/dataloom-backend/app/api/dependencies.py
@@ -4,6 +4,7 @@ import math
 import threading
 import uuid
 
+import pandas as pd
 from fastapi import Cookie, Depends, HTTPException, Request
 from sqlmodel import Session
 
@@ -11,6 +12,7 @@ from app import database, models
 from app.config import get_settings
 from app.services import auth_service
 from app.utils.logging import get_logger
+from app.utils.pandas_helpers import read_table_safe
 from app.utils.rate_limiter import RateLimiter
 
 logger = get_logger(__name__)
@@ -107,3 +109,19 @@ def get_project_or_404(
     if project is None or project.owner_id != current_user.id:
         raise HTTPException(status_code=404, detail=f"Project with ID {project_id} not found")
     return project
+
+
+def load_project_df(project: models.Project) -> pd.DataFrame:
+    """Read a project's working copy, redacting any path from error details.
+
+    ``read_table_safe`` embeds the absolute file path in its 404/500 details;
+    surface a generic message to the client instead, matching the transform
+    endpoint's handling. Shared by the profiling and visualization endpoints.
+    """
+    try:
+        return read_table_safe(project.file_path)
+    except HTTPException as e:
+        if e.status_code == 404:
+            raise HTTPException(status_code=404, detail="Project data file not found") from e
+        logger.warning("Failed to read project file project_id=%s status=%s", project.project_id, e.status_code)
+        raise HTTPException(status_code=e.status_code, detail="Could not read project data") from e
--- a/dataloom-backend/app/api/endpoints/profiling.py
+++ b/dataloom-backend/app/api/endpoints/profiling.py
@@ -7,36 +7,13 @@ descriptive — quality assessment is out of scope here.
 
 import uuid
 
-import pandas as pd
 from fastapi import APIRouter, Depends, HTTPException, Query
 
 from app import models, schemas
-from app.api.dependencies import get_project_or_404
+from app.api.dependencies import get_project_or_404, load_project_df
 from app.services import profiling_service
-from app.utils.logging import get_logger
-from app.utils.pandas_helpers import read_table_safe
-
-logger = get_logger(__name__)
 
 router = APIRouter()
-
-
-def _load_project_df(project: models.Project) -> pd.DataFrame:
-    """Read the project's working copy, redacting any path from error details.
-
-    ``read_table_safe`` embeds the absolute file path in its 404/500 details;
-    surface a generic message to the client instead, matching the transform
-    endpoint's handling.
-    """
-    try:
-        return read_table_safe(project.file_path)
-    except HTTPException as e:
-        if e.status_code == 404:
-            raise HTTPException(status_code=404, detail="Project data file not found") from e
-        logger.warning(
-            "Failed to read project file for profiling project_id=%s status=%s", project.project_id, e.status_code
-        )
-        raise HTTPException(status_code=e.status_code, detail="Could not read project data") from e
 
 
 @router.get("/{project_id}/profile/summary", response_model=schemas.DatasetSummaryResponse)
@@ -45,7 +22,7 @@ async def get_dataset_summary(
     project: models.Project = Depends(get_project_or_404),
 ):
     """Return a top-level summary of the project's dataset."""
-    df = _load_project_df(project)
+    df = load_project_df(project)
     return profiling_service.dataset_summary(df)
 
 
@@ -62,7 +39,7 @@ async def get_column_profile(
     across servers and reverse proxies, which handle ``%2F`` in paths
     inconsistently.
     """
-    df = _load_project_df(project)
+    df = load_project_df(project)
     try:
         return profiling_service.column_profile(df, column_name)
     except KeyError as e:
@@ -80,7 +57,7 @@ async def get_all_column_profiles(
     client needs all columns (e.g. the column-profiles table view) so the
     working copy is read once rather than once per column.
     """
-    df = _load_project_df(project)
+    df = load_project_df(project)
     return profiling_service.all_column_profiles(df)
 
 
@@ -90,5 +67,5 @@ async def get_correlation_matrix(
     project: models.Project = Depends(get_project_or_404),
 ):
     """Return the pairwise Pearson correlation over numeric columns."""
-    df = _load_project_df(project)
+    df = load_project_df(project)
     return profiling_service.correlation_matrix(df)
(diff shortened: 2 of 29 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/visualizations/ChartRenderer.tsx:57` ChartRenderer's mergeByX (used for line/area charts) joins all series by point.x into a Map keyed on that raw x value with no de-duplication. When the chosen x column repeats a value across rows (e.g. several rows sharing the same date, or a low-cardinality/categorical x picked for a line chart), each later point silently overwrites the earlier row object at that key (row[series.name] = point.y), so all but the last data point for a repeated x are dropped from the rendered chart with no error, warning, or meta flag — verified by tracing the Map.set/get logic; there is also no ChartRenderer test in this PR to catch it.
- **[minor/bug]** `dataloom-backend/app/services/visualization_service.py:213` visualization_service._xy_series (build_line_chart/build_area_chart) and build_scatter_plot never check that x_col/y_cols/color_col are distinct before selecting df[[x_col, *y_cols]] or grouping by color_col. Picking the same column twice (e.g. the same column for a line chart's X and Y, or a scatter's color equal to its X or Y) makes pandas raise 'The column label ... is not unique' or 'Grouper for ... not 1-dimensional' (reproduced locally), which the endpoint's blanket `except ValueError` turns into an HTTP 400 that leaks the raw pandas message instead of a clear validation error; ChartBuilderPanel's field options don't filter out the already-chosen column either, so this is reachable from the UI, not just a crafted request.
- **[nit/test]** `dataloom-backend/tests/test_visualizations.py:282` test_requires_auth only calls GET /projects/{id}/charts/suggest; the GET /projects/{id}/charts route (the one that actually touches user-supplied column names) has no matching 401 test, even though both share the same auth dependency.

**Conventions**
- _One of the two commits belongs to the not-yet-merged base PR #401 (explicitly called out by the author); once that merges and this branch rebases, the PR reduces to its own single, clean Charts commit._

PR #407 adds a Charts feature (histogram, bar, line, area, scatter, and pie charts, plus chart auto-suggestions and a correlation heatmap) via a shared backend ChartSpec/aggregation design and a lazy-loaded Recharts frontend; the PR's own commit is clean, thoroughly tested, and clearly described. Two issues are worth fixing before merge: ChartRenderer's mergeByX silently drops/overwrites data points that share the same x value in a line or area chart (a realistic case for repeated dates or categorical x-axes), and the backend's line/area/scatter builders don't validate that x/y/color columns are distinct, so picking the same column twice surfaces a raw pandas error message through the API instead of a clean validation error. Note the diff is inflated by an unmerged dependency (#401) that will disappear once this branch rebases, and there's a small test-coverage gap where only one of the two new endpoints has an explicit auth test.

ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `dataloom-backend/app/services/visualization_service.py:306` build_pie_chart appends a synthetic {"x": "Other", ...} entry whenever truncated is true, without checking if a real category value already equals the literal string "Other". If the label column has a genuine "Other" category that ranks within the top max_slices, and the column has more than max_slices distinct values (so truncation also fires), the response contains two series entries both labeled "Other" (the real category's count and the synthetic overflow sum) that are indistinguishable to any renderer/tooltip. This is a plausible real-world case since "Other" is a very common category label in survey/categorical data.
- **[nit/quality]** `dataloom-frontend/src/Components/visualizations/ChartBuilderPanel.tsx:8` usesBins is imported from ./chartFields but never invoked; the histogram bins control is shown via a direct chartType === "histogram" check further down instead, leaving usesBins as dead code. This isn't caught in CI because package.json's lint script is `eslint . --ext js,jsx`, which doesn't cover .ts/.tsx files at all (no unused-import/no-unused-vars check runs over this file).
- **[nit/quality]** `dataloom-frontend/src/Components/workspace/features/charts.tsx:23` The Charts ribbon menu item's action never sets activePanel (e.g. activePanel: "ChartBuilder"), unlike every transform menu item (Filter, Sort, GroupBy, etc.) which pairs its togglePanel action with a matching activePanel so the ribbon button gets a pressed/active highlight while its panel is open. The Charts button will never show that active state while the Chart Builder panel is docked open.

**Conventions**

This PR adds a complete data-visualization feature: six chart types (histogram, bar, line, area, scatter, pie) built on aggregated, size-bounded backend payloads, plus chart auto-suggestions and a correlation heatmap, rendered through a lazy-loaded Recharts chunk behind a docked no-code builder panel. Verification backs up the quality bar: all 498 backend tests and 153 frontend tests pass on the branch, Ruff and ESLint are clean, and a strict `tsc --noEmit` run (project uses `strict: true` and `noUncheckedIndexedAccess`) turns up zero type errors across the new TypeScript files. One genuine edge-case bug is worth fixing before merge: `build_pie_chart` can silently emit two indistinguishable "Other" slices when a real category is literally named "Other" and the column also exceeds the slice cap. A couple of trivial loose ends (an unused `usesBins` import, a missing active-state highlight on the Charts menu button) are worth a quick cleanup pass but aren't blocking.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/visualization_service.py:256` build_scatter_plot never checks x_col != y_col, and the ChartBuilderPanel UI lets a user pick the same numeric column for both scatter axes (X and Y both draw from the same numericCols list with no mutual exclusion). When x_col==y_col, `cols = [x_col, y_col]` makes df[cols] a frame with two identically-named columns, so `frame[x_col]`/`frame[y_col]` (line 272) resolve to a DataFrame instead of a Series; the zip() then iterates column-label strings, not row values, and _round() swallows the resulting float() failure into None. Verified by reproducing the exact pattern with pandas 3.0.1: the response comes back HTTP 200 with every point's x/y set to null instead of a real (if degenerate) diagonal scatter, and there is no test covering this input.
- **[minor/bug]** `dataloom-backend/app/services/visualization_service.py:213` Same missing guard in _xy_series (shared by line/area): df[[x_col, *y_cols]] with x_col repeated in y_cols raises pandas' own `ValueError: The column label '...' is not unique` from the following .sort_values(x_col) call (reproduced directly with pandas 3.0.1). The /charts endpoint's generic `except ValueError` handler forwards that raw internal pandas message as the 400 response detail for line/area charts instead of a clear validation error such as "x and y must be different columns".
- **[minor/bug]** `dataloom-frontend/src/Components/visualizations/ChartRenderer.tsx:42` Mismatch between the backend's pie-chart overflow-label collision handling and the frontend's color logic. build_pie_chart appends a space to the *overflow* slice's label ("Other ") only when a real category is already literally named "Other" (lines 305-312) so the two don't collide. But ChartRenderer's pieColor() (line 42) special-cases the muted gray color via `label === "Other"` (exact match). So in exactly the edge case the backend comment calls out ("common in survey data"), the real "Other" category keeps the unmodified label "Other" and gets colored gray (wrong — it should get a normal ramp color), while the actual miscellaneous overflow bucket is labeled "Other " and gets a normal ramp color instead of the intended gray. Neither the backend nor frontend test suite exercises a dataset with a real "Other" category alongside truncation, so this inversion isn't caught.
- **[minor/test]** `dataloom-frontend/src/Components/visualizations/ChartRenderer.tsx:1` ChartRenderer.tsx — the component that maps a ChartSpec to one of six recharts renderings (and contains the pie/color logic above) — has no test file, unlike its sibling new components in the same feature (ChartBuilderPanel.test.tsx, CorrelationHeatmap.test.tsx both exist). A snapshot/behavioral test per chart_type branch would have caught the "Other" color inversion above.

**Conventions**
- _Clean single-commit PR with a Conventional-Commits title and an unusually thorough description; the large diff is explained by a disclosed stacking dependency on #401 and should shrink once that merges._

This PR adds a full Charts feature: a new /charts API and visualization_service compute six recharts chart types (histogram, bar, line, area, scatter, pie) plus a correlation heatmap, surfaced through a docked no-code chart builder and auto-suggested starter charts, with solid unit-test coverage across both layers overall. However, the API never validates that a chart's axis columns are actually distinct: picking the same numeric column for both axes of a scatter chart (reachable directly from the builder UI, since neither dropdown excludes the other's current value) silently returns an all-null series instead of an error, and the equivalent case for line/area charts leaks a raw pandas \"column label is not unique\" message through the 400 response. There's also a latent mismatch between the backend's collision-avoiding label for the pie chart's \"Other\" overflow slice and the frontend's color-matching logic, which would mis-color slices for any dataset that has a real category literally named \"Other\". Recommend adding an explicit x != y validation with a friendly error message and a test file for ChartRenderer before merge.

PR #409, opened 2026-07-03
fix: GroupBy Cancel does not revert data to original state (fixes #405)

### LINKED ISSUE #405: [Bug] "Cancel" button in GroupBy tool fails to revert data to original state
**Description**
When using the GroupBy tool, clicking the "Cancel" button after applying a grouping does not revert the data back to its original state. The main table permanently retains the grouped data (dropping the unselected columns and altering row counts) even after the tool panel is closed, acting as if the changes were intentionally saved. 

**To Reproduce**
1. Go to the DataLoom workspace with an active dataset.
2. Click on the 'GroupBy' tool in the top navigation bar.
3. Select columns to group by (e.g., `Transaction_ID` and `Date`).
4. Select an Aggregation Column (e.g., `Qty`) and a Function (e.g., `Sum`).
5. Click the blue 'Apply GroupBy' button and observe the table updating to the grouped view.
6. Click the 'Cancel' button on the right side of the tool panel.
7. See error: The tool panel closes, but the main data table remains stuck in the grouped state instead of reverting to the original, un-grouped dataset.

**Dataset**

[test_dataset.csv](https://github.com/user-attachments/files/29561489/test_dataset.csv)

**Expected behavior**
Clicking the 'Cancel' button should immediately discard the applied transformations and revert the main data table to the exact state it was in before the GroupBy tool was initiated. 

**Additional context**
This is a critical UX issue. If a user is just exploring data and clicks 'Cancel' to back out, their working dataset is unintentionally overwritten. They are left with no way to revert unless there is a separate "undo" feature, making data loss highly likely during routine use.

shivv23:
## Description

Fixes #405. The GroupBy form now correctly reverts the data table when Cancel is used, even when the panel's header X button dismisses the form instead of the in-form Cancel button.

### Root cause

The original GroupBy implementation had persist=True in TRANSFORMATION_REGISTRY (the only reshaping-style transform with that setting), which caused the working file to be overwritten on every Apply — making Cancel non-functional.

### What this PR does

The upstream already adopted a universal preview-before-persist pattern (PR #410). This PR adds the remaining guardrails on top:

| Scenario | Before | After |
|---|---|---|
| Click **Cancel** in preview mode | Already worked (upstream code) | Works (unchanged) |
| Click panel header **X** while in preview mode | Bypassed cancelPreview — data stayed grouped | useEffect cleanup calls cancelPreview() on unmount |
| **Cancel/close while Apply in flight** | Pending response could call enterPreviewMode after unmount | cancelledRef guards async continuation; Cancel button disabled during loading |
| No error handling in revert path | N/A — upstream has no async revert in handleCancel | N/A (cancelPreview is synchronous) |

### Changes

**dataloom-frontend/src/Components/forms/GroupByForm.jsx**
- Add cancelledRef + useEffect cleanup: sets ref on unmount and calls cancelPreview() if in preview mode (fixes X button bypass)
- Check cancelledRef.current in handleSubmit before calling enterPreviewMode (fixes race condition)
- Disable Cancel button while loading is true
- Wrap form in data-testid="groupby-form", add data-testid="groupby-agg-column" to ColumnSelect

**dataloom-frontend/src/test/GroupByForm.test.jsx** (new)
- 6 tests covering: apply success, apply failure, close-during-loading, Cancel in preview mode, Cancel outside preview mode, unmount cleanup in preview mode

### Testing
- npx vitest run — 162 tests pass across 31 test files (6 new)
- npx eslint — clean

### DIFF
--- a/dataloom-frontend/src/Components/forms/GroupByForm.jsx
+++ b/dataloom-frontend/src/Components/forms/GroupByForm.jsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useRef, useEffect } from "react";
 import PropTypes from "prop-types";
 import { transformProject } from "../../api";
 import { GROUPBY } from "../../constants/operationTypes";
@@ -25,6 +25,20 @@ const GroupByForm = ({ projectId, onClose }) => {
   const [loading, setLoading] = useState(false);
   const { error, clearError, handleError } = useError();
   const { saving, handleSave } = usePreviewSave({ clearError, handleError, onClose });
+  const cancelledRef = useRef(false);
+  const cancelPreviewRef = useRef(cancelPreview);
+  const isPreviewModeRef = useRef(isPreviewMode);
+  cancelPreviewRef.current = cancelPreview;
+  isPreviewModeRef.current = isPreviewMode;
+
+  useEffect(() => {
+    return () => {
+      cancelledRef.current = true;
+      if (isPreviewModeRef.current) {
+        cancelPreviewRef.current();
+      }
+    };
+  }, []);
 
   const handleSubmit = async (e) => {
     e.preventDefault();
@@ -43,6 +57,7 @@ const GroupByForm = ({ projectId, onClose }) => {
         },
       };
       const response = await transformProject(projectId, payload, { preview: true });
+      if (cancelledRef.current) return;
       enterPreviewMode(response.columns, response.rows, response.dtypes, { projectId, payload });
     } catch (err) {
       handleError(err);
@@ -60,7 +75,7 @@ const GroupByForm = ({ projectId, onClose }) => {
   };
 
   return (
-    <div>
+    <div data-testid="groupby-form">
       <form onSubmit={handleSubmit}>
         <div className="mb-3">
           <label className="block mb-1 text-sm font-medium text-gray-700">Group By Columns:</label>
@@ -80,6 +95,7 @@ const GroupByForm = ({ projectId, onClose }) => {
             onChange={setAggColumn}
             options={availableColumns.filter((col) => !selectedColumns.includes(col))}
             required
+            data-testid="groupby-agg-column"
           />
         </div>
         <div className="mb-4">
@@ -102,7 +118,12 @@ const GroupByForm = ({ projectId, onClose }) => {
               </Button>
             )}
           </div>
-          <Button type="button" variant="secondary" onClick={handleCancel}>
+          <Button
+            type="button"
+            variant="secondary"
+            onClick={handleCancel}
+            disabled={loading || saving}
+          >
             Cancel
           </Button>
         </div>
--- a/dataloom-frontend/src/test/GroupByForm.test.jsx
+++ b/dataloom-frontend/src/test/GroupByForm.test.jsx
@@ -1,0 +1,197 @@
+import { render, within } from "@testing-library/react";
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import GroupByForm from "../Components/forms/GroupByForm";
+import { transformProject } from "../api";
+
+vi.mock("../api", () => ({
+  transformProject: vi.fn(),
+}));
+
+const mockContext = {
+  columns: ["City", "Amount", "Date"],
+  dtypes: { City: "string", Amount: "float", Date: "date" },
+  columnOrder: [0, 1, 2],
+  updateData: vi.fn(),
+  enterPreviewMode: vi.fn(),
+  cancelPreview: vi.fn(),
+  confirmPreview: vi.fn(),
+  refreshProject: vi.fn(),
+  pageSize: 50,
+  pendingTransform: null,
+  isPreviewMode: false,
+};
+
+vi.mock("../context/ProjectContext", () => ({
+  useProjectContext: () => mockContext,
+}));
+
+vi.mock("../context/HistoryRefreshContext", () => ({
+  useHistoryRefresh: () => ({ refreshLogs: vi.fn(), refreshCheckpoints: vi.fn() }),
+}));
+
+beforeEach(() => {
+  vi.resetAllMocks();
+  mockContext.isPreviewMode = false;
+  mockContext.pendingTransform = null;
+});
+
+describe("GroupByForm", () => {
+  it("calls enterPreviewMode after successful apply", async () => {
+    transformProject.mockResolvedValueOnce({
+      columns: ["City", "Amount"],
+      rows: [["Paris", "300"]],
+      dtypes: { City: "string", Amount: "float" },
+    });
+
+    const { getByText, getByTestId, getByRole } = render(
+      <GroupByForm projectId="proj-1" onClose={vi.fn()} />,
+    );
+    const { fireEvent } = await import("@testing-library/react");
+
+    fireEvent.click(getByText("City"));
+    fireEvent.click(getByTestId("groupby-agg-column"));
+    fireEvent.click(within(getByRole("listbox")).getByText("Amount"));
+    fireEvent.click(getByText("Apply GroupBy"));
+
+    await vi.waitFor(() => {
+      expect(mockContext.enterPreviewMode).toHaveBeenCalledWith(
+        ["City", "Amount"],
+        [["Paris", "300"]],
+        { City: "string", Amount: "float" },
+        {
+          payload: {
+            operation_type: "groupby",
+            groupby_params: {
+              columns: ["City"],
+              agg_column: "Amount",
+              agg_function: "sum",
+            },
+          },
+          projectId: "proj-1",
+        },
+      );
+    });
+  });
+
+  it("does not call enterPreviewMode when transform fails", async () => {
+    transformProject.mockRejectedValueOnce(new Error("Server error"));
+
+    const { getByText, getByTestId, getByRole } = render(
+      <GroupByForm projectId="proj-1" onClose={vi.fn()} />,
+    );
+    const { fireEvent } = await import("@testing-library/react");
+
+    fireEvent.click(getByText("City"));
+    fireEvent.click(getByTestId("groupby-agg-column"));
+    fireEvent.click(within(getByRole("listbox")).getByText("Amount"));
+    fireEvent.click(getByText("Apply GroupBy"));
+
+    await vi.waitFor(() => {
+      expect(mockContext.enterPreviewMode).not.toHaveBeenCalled();
+    });
+  });
+
+  it("does not call enterPreviewMode when user closes during loading", async () => {
+    let resolveTransform;
+    transformProject.mockImplementationOnce(
+      () =>
+        new Promise((resolve) => {
+          resolveTransform = resolve;
+        }),
+    );
+
+    const { getByText, getByTestId, getByRole, unmount } = render(
+      <GroupByForm projectId="proj-1" onClose={vi.fn()} />,
+    );
+    const { fireEvent } = await import("@testing-library/react");
+
+    fireEvent.click(getByText("City"));
+    fireEvent.click(getByTestId("groupby-agg-column"));
+    fireEvent.click(within(getByRole("listbox")).getByText("Amount"));
+    fireEvent.click(getByText("Apply GroupBy"));
+    unmount();
+
+    resolveTransform({
+      columns: ["City", "Amount"],
+      rows: [["Paris", "300"]],
+      dtypes: { City: "string", Amount: "float" },
+    });
+
+    await vi.waitFor(() => {
+      expect(mockContext.enterPreviewMode).not.toHaveBeenCalled();
+    });
+  });
+
+  it("allows re-submit after a cancelled in-flight request", async () => {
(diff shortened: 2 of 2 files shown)


hanzalahwaheed:
Can we instead have an `Apply` button for all transformations. We don't update the dataframe unless the Apply button is clicked.

YADUNANDAN-SINGH:
Thanks for jumping on this @shivv23, and good catch on the `persist=True` flag being the root cause.

As the issue author, I'd scoped a slightly broader approach in my comment above: rather than just flipping GroupBy to `persist=False`, having Apply *preview* the transform with an explicit Cancel/Save step — which lines up closely with what @hanzalahwaheed is describing (not updating the dataframe until Apply is clicked, consistently across all transformations). I've done similar preview-before-persist work in this repo before, so I'm happy to take that broader Apply-button refactor as a follow-up PR so the behavior is consistent across GroupBy, Pivot, etc.

@ivantha does that direction work for you? Happy to open a draft and scope it out.

YADUNANDAN-SINGH:
Hey @shivv23 — wanted to flag this here rather than let it surface on its own.

Following up on @hanzalahwaheed's comment above about a consistent Apply-button 
workflow, I opened #410, which implements that as a broader refactor: a universal 
preview-before-persist step (usePreviewSave) across all structural transforms 
(GroupBy, PivotTable, Melt, Sort, etc.), not just the persist=True flag flip 
on GroupBy.

Not trying to step on this PR — your fix is correct and much smaller in scope, 
so it might make sense to merge this first as the immediate fix and treat #410 
as the follow-up architecture change. Deferring to @ivantha on how they'd like 
to sequence these, happy to adjust either way.

ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:54` The revert-on-cancel fix is wired only to GroupByForm's own in-form 'Cancel' button via handleClose. The side panel's header close ('X') button is rendered by SidePanel.tsx and bound directly to the same onClose/closePanel reference in RightPanel.jsx (`<SidePanel onClose={closePanel}><Component onClose={closePanel} /></SidePanel>`), completely bypassing handleClose. If a user applies a GroupBy and dismisses the panel via that X button instead of the Cancel button, ProjectContext's columns/rows remain stuck on the grouped preview (updateData already mutated the shared context state in handleSubmit) because refreshProject is never invoked on that path -- reproducing the exact stuck-grouped-view bug this PR claims to fix, through an equally standard UI affordance.
- **[major/test]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:46` The new conditional revert logic (hasAppliedRef set on a successful submit, consumed in handleClose to decide whether to call refreshProject) ships with no test at all: there is no existing or newly-added test file for GroupByForm, and none of the e2e specs under e2e/ cover a GroupBy apply/cancel flow. A component test asserting refreshProject fires only when a groupby was applied before cancel (and not otherwise) would also have caught the header-close-button bypass in the finding above.
- **[minor/bug]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:56` handleClose awaits refreshProject and then unconditionally calls setResult(null) and onClose(), even if the refresh failed. refreshProject swallows its own errors internally (catches and stores them as a context-level error state rather than rethrowing), so on a transient failure during cancel the panel still closes as though the revert succeeded, potentially leaving the grouped preview on screen with no signal to the user that their data was not actually reverted.
- **[nit/quality]** `dataloom-backend/app/services/transformation_service.py:782` The TransformationSpec.persist docstring's example list of read-only-preview operations ('advanced query, pivot, melt') is now stale, since this PR adds group by to that same persist=False category. Worth updating the comment so future readers of the registry understand groupby is intentionally grouped with the other reshaping/preview ops.

**Conventions**

This PR correctly diagnoses the root cause (GroupBy was the only reshaping-style transform left with persist=True) and the persist=False flip plus the hasAppliedRef-gated refresh-on-cancel is a sound approach, backed by an unusually thorough PR description. However, the revert logic is wired only to GroupByForm's own Cancel button; the side panel's header 'X' close button calls the same onClose callback directly and bypasses that logic entirely, so a user who applies a GroupBy and closes the panel that way will still hit the exact stuck-grouped-view bug this PR is meant to fix. The fix also ships with zero new tests (no GroupByForm unit test exists, no e2e coverage) despite fixing a data-loss-adjacent bug. Recommend covering the panel's header-close path (e.g., reset on unmount or route the revert through the parent panel) and adding a regression test before merging.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:42` handleSubmit's transformProject call is not cancellation-aware, and nothing disables Cancel or the SidePanel header's X button while `loading` is true (SidePanel.tsx wires its close button straight to `onClose`, and GroupByForm's own Cancel button has no `disabled={loading}`). So a user can Cancel/close the panel while an Apply request is still in flight: handleClose/the unmount cleanup then correctly see `hasAppliedRef.current === false` and skip refreshProject, but the pending request can resolve afterward and still run `updateData(...)` plus `hasAppliedRef.current = true` against the shared project context after the panel is already gone. Nothing is left to trigger a corrective refreshProject, so the main table silently flips to the grouped view with no way back -- reproducing the exact #405 bug this PR sets out to fix, just under a timing window instead of the original always-on case.
- **[major/test]** `dataloom-frontend/src/test/GroupByForm.test.jsx:61` The second new test never actually clicks the Cancel button after a successful Apply -- it calls testing-library's raw `unmount()` instead, which only exercises the separate unmount-cleanup effect (lines 73-79), not `handleClose` (lines 63-71). That means the literal repro from issue #405 (click Apply, then click the Cancel button) has no direct test: a regression that broke `handleClose`'s `if (wasApplied)` branch specifically, while leaving the unmount-cleanup effect intact, would pass this suite.
- **[minor/quality]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:67` `await refreshProject(projectId, 1, pageSize)` here has no try/catch, unlike handleSubmit which routes failures through `handleError`. If the revert fetch rejects (e.g. a transient network error), handleClose throws before reaching `setResult(null)`/`onClose()`, so clicking Cancel after a successful Apply would leave the panel stuck open on the grouped preview with no visible error and an apparently-dead Cancel button.

**Conventions**
- **Commits (2):** Second commit's subject bundles four unrelated fixup items into one run-on line and names the reviewer ('address ivantha review') rather than describing the change itself; better squashed into the first commit or rewritten before merge.
- _Title and description are exemplary (clear root-cause table, explicit follow-up note); just tighten or squash the second commit's message._

This PR correctly traces #405 to GroupBy being the lone aggregation-style transform with persist=True and fixes it by aligning it with the other preview operations (pivot, melt, advQueryFilter), while reworking GroupByForm's cancel/unmount handling to restore the original view. I cloned the branch and confirmed locally that the full frontend suite (131 tests/25 files) and the relevant backend tests pass, with clean ESLint/Ruff output, matching the PR's own testing notes. The remaining gap is a real race: neither the Cancel button nor the SidePanel's header X button is disabled while the Apply request is in flight, so a fast cancel/close can be followed by the pending response landing anyway and silently reapplying the grouped view with nothing left to revert it. The new tests also never actually click Cancel after a successful Apply (the literal #405 repro) -- only the "no apply" and "raw unmount" paths are covered -- and handleClose's revert fetch has no error handling, so a failed revert would leave the panel stuck open.

ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:32` This effect's cleanup (dependency array `[isPreviewMode, cancelPreview]` at line 37) runs on every preview-mode transition, not only on real unmount, because `cancelPreview` is a `useCallback` in ProjectContext keyed on `previewSnapshot` and changes identity the instant preview mode is entered or exited (and `isPreviewMode` itself is also a listed dependency). Since `cancelledRef.current = true` here is set unconditionally and is never reset back to false anywhere in the file, it becomes permanently true immediately after the first successful Apply (the moment isPreviewMode flips to true), so the guard at line 56 silently drops every later Apply in the same mounted session, e.g. Apply, then Cancel, then change parameters and Apply again shows "Applying..." and then reverts with no preview, no Save button, and no error shown.
- **[major/test]** `dataloom-frontend/src/test/GroupByForm.test.jsx:95` `transformProject.mockReturnValueOnce(new Promise(() => {}))` never resolves, so this test passes trivially: enterPreviewMode was never going to be called regardless of whether the cancelledRef guard actually works, since the awaited promise never settles before or after `unmount()` on line 106. No test resolves the transform promise after an unmount/cancel and then re-submits, so the async guard this test is named for is never actually exercised, which is how the cancelledRef permanent-poisoning bug ships through a fully green suite.
- **[minor/bug]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:117` The Cancel button is now disabled only while `loading` (Apply in flight); it is not disabled while `saving` is true (Save Changes in flight, from usePreviewSave). Because `handleCancel` calls `cancelPreview()` unconditionally whenever `isPreviewMode` is true, and isPreviewMode only flips to false once the in-flight save's `confirmPreview()` runs, a user can click Cancel mid-save and trigger a client-side revert that races the in-flight persist call, the same class of async race this PR otherwise guards against for Apply but leaves unaddressed for Save.

**Conventions**

This PR's unmount-cleanup effect in GroupByForm.jsx (lines 30 to 37) fires on every preview-mode transition rather than only on real unmount, because its dependency array includes isPreviewMode and cancelPreview, both of which change identity whenever preview mode toggles; since it unconditionally sets cancelledRef.current = true and nothing ever resets that ref, the very first successful Apply permanently poisons it, silently breaking every later Apply in the same session (line 56). The added test for the close-during-loading race (test file, line 94) does not actually exercise this path because its mocked promise never resolves, so the regression slips through a fully green suite. The Cancel button's new disabled state also only accounts for loading (Apply in flight), leaving the same class of race unaddressed while a Save is in flight (line 117). PR title, description, and the single commit are all clean and Conventional-Commits compliant; recommend reworking the cancellation tracking (for example an unmount-only effect with an empty dependency array, or resetting cancelledRef.current at the top of handleSubmit) and adding a test that resolves the transform promise after unmount/cancel and then re-submits, before merging.

ivantha:
### Automated review summary

**Findings**
- **[minor/test]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:37` The unmount cleanup's cancelPreview() call is only correct because isPreviewMode has already flipped to false (via confirmPreview() in usePreviewSave.handleSave, which runs before onClose()) by the time a successful Save unmounts the form. I traced this against usePreviewSave.js and it holds today under React 18's batching across the intervening 'await refreshProject' call, but it is an untested, implicit cross-file ordering assumption: a future reorder in usePreviewSave.js (e.g. calling onClose() before confirmPreview(), or dropping the intermediate await) would make this cleanup wrongly re-fire cancelPreview() after a successful save and silently revert data the user just persisted.
- **[minor/test]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:125` The Cancel button's new disabled={loading || saving} attribute has no dedicated test; none of the 6 new tests assert the button is actually disabled while an Apply/Save is in flight, or that a click on it is a no-op in that state.
- **[nit/test]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:78` The new data-testid="groupby-form" follows the filter-form/sort-form convention already used in e2e/transformations.spec.js, but nothing consumes it: GroupBy still has no Playwright e2e coverage for the Cancel-revert scenario that produced issue #405, unlike Filter and Sort which are covered in that spec file.
- **[nit/style]** `dataloom-frontend/src/Components/forms/GroupByForm.jsx:31` cancelPreviewRef.current and isPreviewModeRef.current are reassigned directly in the render body instead of being synced inside a useEffect. It works here because refs don't trigger re-renders and the values are only read from the unmount cleanup, but mutating refs during render is a non-idiomatic React pattern that a short comment would help justify for future readers.

**Conventions**
- _The title's "(fixes #405)" is stale since that issue was already closed by the upstream preview-workflow PR #410 before this branch reached its current form, though the PR body is transparent that its real contribution is the two edge-case guardrails layered on top._

This PR closes a real gap in the GroupBy preview/cancel flow: closing the panel via the header X button (rather than the in-form Cancel button) previously bypassed cancelPreview() and left the table stuck showing grouped data, and a slow Apply request could resolve after the form was dismissed and silently flip the table into preview mode with no form left to cancel it. The fix (an unmount effect that calls cancelPreview() when leaving while still in preview mode, a cancelledRef guard around the async response, and disabling Cancel while a request is in flight) is sound; I traced its ordering assumptions against ProjectContext.jsx and usePreviewSave.js and found no double-cancel or lost-save scenario in the current code. The new unit tests cover the core logic well, but the newly added disabled state on the Cancel button has no direct test, and there is no regression test for the interaction between a successful Save and the new unmount cleanup, an implicit cross-file ordering assumption worth locking down. Neither gap blocks merge, but both are cheap to add and would harden this against future refactors of usePreviewSave or the panel's close flow.

PR #412, opened 2026-07-05
fix(security): bump pyjwt and python-multipart to patched releases

anxkhn:
## Description

The backend lockfile pinned two request-path dependencies at versions with published advisories, and the `pyproject.toml` floors sat well below the fixed releases (`python-multipart>=0.0.6`, `pyjwt>=2.9`). There is no Dependabot config in the repo, so these were never surfaced.

- `python-multipart` parses every multipart upload body (`File(...)` / `Form(...)` in `dataloom-backend/app/api/endpoints/projects.py`). Releases before `0.0.31` carry denial-of-service advisories: large preamble/epilogue, unbounded part headers, quadratic querystring parsing, and negative `Content-Length` buffering. The lock resolved to `0.0.22`.
- `pyjwt` signs and verifies auth tokens (`jwt.encode` / `jwt.decode` in `dataloom-backend/app/services/auth_service.py`). Releases before `2.13.0` carry published advisories. The lock resolved to `2.12.1`.

This raises the two `pyproject.toml` floors to the first patched releases and refreshes `uv.lock` accordingly (`python-multipart` 0.0.22 -> 0.0.32, `pyjwt` 2.12.1 -> 2.13.0). No source changes are required; both upgrades are drop-in for the APIs this project uses (a symmetric-secret `encode`/`decode`; standard `File`/`Form` upload parsing). Only these two packages and their `requires-dist` specifiers change in the lockfile; no other dependency drifts.

A small regression guard (`dataloom-backend/tests/test_dependency_security.py`) asserts the resolved environment stays at or above the patched minimums, so a future stale lockfile or accidental downgrade fails fast in CI.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Backend suite run from `dataloom-backend` with `DATABASE_URL=sqlite:///./test.db` (Python 3.12):

- `uv sync --frozen` resolves consistently against the refreshed lock; the environment reports `python-multipart` 0.0.32 and `pyjwt` 2.13.0.
- `uv run pytest -q` -> all tests pass (baseline plus the new guard). The auth and upload suites (`test_auth.py`, `test_upload.py`, `test_multiformat_upload.py`) pass, confirming the upgrades are drop-in.
- `uv run ruff check .` and `uv run ruff format --check .` -> clean.
- `uv lock --check` -> passes (pyproject and lockfile stay consistent, matching the CI `uv sync --frozen` gate).
- Red/green confirmed: the new guard fails against the previously pinned versions (0.0.22 / 2.12.1) and passes against the patched ones (0.0.32 / 2.13.0), so it is a genuine regression guard rather than a tautology.

- [x] Existing tests pass
- [x] New tests added
- [ ] Manual testing

## Checklist

- [x] Tests pass locally


YADUNANDAN-SINGH:
The regression guard is a nice touch — pinning the floors and refreshing the lock together is the right call.

One small thing: the guard imports `from packaging.version import Version`, but `packaging` isn't declared in `pyproject.toml`. It resolves today because `pytest` depends on it, so nothing's broken — but since the test imports it directly, adding `packaging` to the `dev` group makes that explicit instead of leaning on pytest's tree.

Separately: good call flagging the missing Dependabot config. Might be worth its own PR so it can cover both the backend (pip) and the frontend (npm) in one file — happy to open that if it's useful.

ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `dataloom-backend/tests/test_dependency_security.py:11` The new test imports `packaging` directly, but `packaging` is not declared anywhere in pyproject.toml (neither main deps nor the dev group) — it only appears in uv.lock as a transitive dependency of pytest, so this direct import relies on an undeclared package.

**Conventions**

Clean, well-scoped security fix: raises the pyproject.toml floors for pyjwt (>=2.13.0) and python-multipart (>=0.0.31) to close DoS and auth-token advisories the lockfile had resolved down to (2.12.1 / 0.0.22), refreshes uv.lock accordingly, and adds a regression test that fails CI if the environment ever drops below those minimums. I spot-checked two of the cited GHSA advisories and both correspond correctly to the claimed packages and fix versions. Only nitpick: the new test imports `packaging` without declaring it anywhere in pyproject.toml, relying on pytest's transitive dependency to provide it — works today but is worth an explicit dev-dependency entry for robustness. Otherwise safe to merge as-is.

PR #426, opened 2026-07-06
fix: disable Save/Export/Undo toolbar buttons while previewing a transformation

shivv23:
### Problem

When a user previews a transformation (via Filter, Sort, GroupBy, etc.), the data on screen reflects the \*preview\* state — an in-memory copy that has **not** been persisted to the working copy on disk. However, the **Save**, **Export**, and **Undo** toolbar buttons remain active. Clicking any of these while in preview mode silently acts on the **unmodified** working copy on disk, not the visible preview data:

- **Save** creates a checkpoint of the stale disk file, discarding the preview
- **Export** downloads the stale disk file, not what the user sees
- **Undo** reverts the last persisted transform while preview is active, creating a stale undo target

This is a silent data integrity bug — the user has no indication that their action did not apply to the data they are looking at.

### Fix

Add \disabled: isPreviewMode\ to the Save, Export, and Undo core ribbon items in \MenuNavbar.jsx\, following the exact pattern already used by feature-contributed items (line 151 in the original: \disabled: item.disabledInPreview ? isPreviewMode : false\). Also add a \	itle\ attribute to the rendered button to surface the hover/tooltip text explaining why the button is disabled.

### Changes

\dataloom-frontend/src/Components/MenuNavbar.jsx\ (2 files — not a mistake, it's one file with two logical changes):

1. **coreItems** — Save, Export, Undo each get \disabled: inPreview\ + conditional \hover\ text that explains the disabled state
2. **button renderer** — added \	itle={item.hover}\ so the tooltip text is visible to users (hover text was already declared on some items but never wired to the DOM)

### Testing

- \
pm run lint\ — clean (0 warnings)
- \
pm run test\ — 162/162 tests pass
- Existing disabled-button styling (opacity-50, cursor-not-allowed, disabled:hover:bg-transparent) applies automatically via the already-present CSS classes

### DIFF
--- a/dataloom-frontend/src/Components/MenuNavbar.jsx
+++ b/dataloom-frontend/src/Components/MenuNavbar.jsx
@@ -77,6 +77,8 @@ const MenuNavbar = ({ projectId }) => {
     }
   };
 
+  const inPreview = isPreviewMode;
+
   // Core ribbon items — the ones that need component-local state/handlers and so
   // can't be declared as (declarative) feature menu items.
   const coreItems = [
@@ -87,7 +89,10 @@ const MenuNavbar = ({ projectId }) => {
       label: "Save",
       icon: LuSave,
       onClick: handleSave,
-      hover: "Save the current state of the project as a new checkpoint.",
+      disabled: inPreview,
+      hover: inPreview
+        ? "Save is unavailable while previewing a transformation."
+        : "Save the current state of the project as a new checkpoint.",
     },
     {
       ribbon: "File",
@@ -96,7 +101,10 @@ const MenuNavbar = ({ projectId }) => {
       label: "Export",
       icon: LuDownload,
       onClick: () => setShowExportModal(true),
-      hover: "Export the data to a file.",
+      disabled: inPreview,
+      hover: inPreview
+        ? "Export is unavailable while previewing a transformation."
+        : "Export the data to a file.",
     },
     {
       ribbon: "File",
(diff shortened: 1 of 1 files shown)


ivantha:
### Automated review summary

**Findings**
- **[minor/test]** `dataloom-frontend/src/Components/MenuNavbar.jsx:92` No unit or e2e test asserts the new disabled-while-previewing behavior for Save/Export/Undo. Existing coverage (the unit suite, e2e/checkpoints.spec.js, e2e/export.spec.js) clicks these toolbar buttons via their data-testid but never does so with isPreviewMode true, so a later edit to coreItems could silently reintroduce the stale-data bug this PR fixes.
- **[nit/quality]** `dataloom-frontend/src/Components/MenuNavbar.jsx:80` `const inPreview = isPreviewMode;` is a pure rename with no derived logic, and it is inconsistent with the existing pattern a few lines below (line 146, `disabled: item.disabledInPreview ? isPreviewMode : false`) which reads `isPreviewMode` directly from context. Referencing `isPreviewMode` directly in the three new fields would avoid the redundant alias.

**Conventions**
- **Description:** Does not link a closing issue, and claims a change ('button renderer, added title={item.hover}') that is not in the diff, since that attribute already existed on main before this PR (verified against the pre-PR file).
- _Single clean commit and a solid problem/fix write-up otherwise; just correct the button-renderer claim and link the underlying bug report before merge._

This is a small, well-scoped fix: it wires the existing isPreviewMode flag into the Save, Export, and Undo core toolbar items so they disable, with an explanatory tooltip, while a transformation preview is active, closing a real data-integrity gap where these actions previously operated on stale, unpreviewed disk data. The change is a single clean commit and matches the repo's existing disabled-in-preview pattern already used by feature-contributed menu items, so it should be low-risk to merge. Two minor points to clean up: no test (unit or e2e) was added to guard the new disabled behavior, and the description claims a button-renderer change that is not actually in the diff since that title attribute already existed on main.

PR #435, opened 2026-07-12
fix: translate page-local row index to global DataFrame index for add/delete/edit row

shivv23:
## Bug

On any page > 1, row-level operations (cell edit, add row, delete row) send a **page-local** row index (0-based within the visible page) to the backend, which treats it as a global DataFrame index. This silently corrupts the wrong row.

### Affected operations

- **Cell edit** (\handleEditCell\, line 321) — writes \
ewValue\ to the wrong cell
- **Add row** (\handleAddRow\, line 178) — inserts blank row at the wrong position
- **Delete row** (\handleDeleteRow\, line 231) — deletes the wrong row

The corrupted index self-propagates through undo/save/revert because the change log stores the page-local index.

## Fix

Add \(page - 1) * pageSize\ offset to the row index before sending to the backend, converting page-local to global DataFrame index. Column index translation (via \columnOrder\) was already correct — only row indices were missing the offset.

## Verification

- Lint: clean
- Tests: 156/156 passing

### DIFF
--- a/dataloom-frontend/src/Components/Table.tsx
+++ b/dataloom-frontend/src/Components/Table.tsx
@@ -173,9 +173,10 @@ const Table = ({ projectId, showColumnProfiles = false }: TableProps) => {
 
   const handleAddRow = async (index: number) => {
     try {
+      const globalIndex = (page - 1) * pageSize + index;
       const response = await applyTransform({
         operation_type: ADD_ROW,
-        row_params: { index },
+        row_params: { index: globalIndex },
       });
       updateTableData(response);
       refreshProject(projectId, 1, pageSize);
@@ -226,9 +227,10 @@ const Table = ({ projectId, showColumnProfiles = false }: TableProps) => {
 
   const handleDeleteRow = async (index: number) => {
     try {
+      const globalIndex = (page - 1) * pageSize + index;
       const response = await applyTransform({
         operation_type: DELETE_ROW,
-        row_params: { index },
+        row_params: { index: globalIndex },
       });
       updateTableData(response);
       refreshProject(projectId, 1, pageSize);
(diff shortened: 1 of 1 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/test]** `dataloom-frontend/src/Components/Table.tsx:319` The page-local-to-global row-index translation (handleAddRow, handleDeleteRow, and this handleEditCell line) ships with no regression test. The existing dataloom-frontend/src/test/Table.columnOrder.test.jsx already mocks page, pageSize, and transformProject, so asserting the translated index in the transformProject call args when page > 1 would be a small addition and would catch this exact bug if it's ever reintroduced by a future refactor.
- **[minor/quality]** `dataloom-frontend/src/Components/Table.tsx:176` The `(page - 1) * pageSize` global-index formula is now duplicated three times (handleAddRow, handleDeleteRow, handleEditCell), joining a fourth pre-existing copy used for the S.No. column at line 156. Extracting a single toGlobalIndex(localIndex) helper would remove the duplication and give the translation one place to fix/test if pagination semantics ever change.
- **[nit/style]** `dataloom-frontend/src/Components/Table.tsx:665` This hunk re-indents the TablePagination wrapper <div> (2 spaces -> 6 spaces), unrelated to the stated row-index fix — it looks like a drive-by correction of an indentation slip left by a prior commit. Harmless, but a formatting-only change like this is easier to review split out from a bug-fix diff.

**Conventions**

PR #435 correctly fixes a real bug: the Table component's add-row, delete-row, and cell-edit handlers were sending page-local (0-based within the visible page) row indices to the backend, which silently mutated the wrong row on any page beyond the first. The fix adds the standard `(page - 1) * pageSize` offset, matching both the pre-existing S.No.-column display logic and the backend's documented 0-based global-index contract for add_row/delete_row/change_cell_value, so the core logic checks out. Before merging, it's worth asking the author for a regression test — the existing Table.columnOrder.test.jsx already mocks page/pageSize/transformProject, so a page>1 assertion on the outgoing index would be cheap to add and would lock the fix in. There's also a small unrelated indentation fix riding along in the diff and some formula duplication worth a follow-up refactor, but neither blocks merging.

PR #436, opened 2026-07-12
feat: Implement dedicated multi-dataset comparison view

### LINKED ISSUE #429: [Feature] Cross-dataset comparison view — grouped/multi-line charts across two or more datasets
## Summary

The current Profiling tab turns a single dataset into visual summaries (charts, column profiles). I'd like to propose extending this so users can **compare two or more datasets against each other** in a single view — e.g. grouped bar charts or multi-line charts where each dataset is a differently-colored series sharing one axis.

The common use case: a user has "Coffee shop dataset" and "Super market dataset" (or two time periods of the same data) and wants to see them side by side rather than flipping between projects.

## Motivation

Profiling today answers *"what does this dataset look like?"* — comparison answers *"how do these datasets differ?"*, which is a distinct and frequent analytical need. It reuses the existing charting engine, so most of the visualization layer already exists; the new work is dataset selection and axis alignment.

## The core design question

The interesting part isn't the rendering — it's **how two datasets get aligned for comparison**. Datasets rarely share an identical schema, so we need a shared *dimension* (the axis, e.g. `Item_Ordered` or `Date`) and a shared *measure* (the value, e.g. `Total_Amount`), plus an aggregation (sum/avg/count). I see two ways to structure this, and I'd like your input on which fits DataLoom's direction.

## Approach A — Inline, within an open dataset

Add a "Compare" control inside the existing Profiling tab of an open dataset. The user picks one *other* existing dataset to overlay against the current one, maps the comparable columns, and the charts render as overlaid series.

- **Pros:** Small, contained change; lives next to existing profiling UI; minimal new navigation.
- **Cons:** Anchored to "current dataset vs. one other" framing; awkward for comparing 3+ datasets or datasets from different projects; comparison feels secondary rather than first-class.

## Approach B — Dedicated comparison view from the home page *(my preference)*

Add a "Compare Datasets" entry point on the home/projects page. The user explicitly selects two or more datasets (from existing projects, and later optionally uploads a new one), maps a shared dimension + measure + aggregation, and gets a dedicated comparison workspace supporting grouped bar and multi-line charts with a per-dataset legend.

- **Pros:** Comparison is inherently multi-dataset, so a dedicated entry point models it more honestly; naturally scales to 3+ datasets and cross-project comparison; cleaner separation from single-dataset profiling; room to grow (upload-on-the-fly, saved comparisons).
- **Cons:** Larger surface area; introduces a new view and navigation; needs a column-mapping UX since schemas won't match automatically.

## Suggested MVP (to keep initial scope reviewable)

Regardless of A or B:

1. Select exactly 2 datasets.
2. Auto-detect shared/mappable columns; let the user pick one dimension + one measure + an aggregation.
3. Render **grouped bar** and **multi-line** charts, one color per dataset, with a legend.
4. Defer to later: 3+ datasets, on-the-fly upload, pie/100%-stacked, saved comparison configs.

(I'd suggest skipping side-by-side pie charts for comparison entirely — they're hard to read comparatively; a 100%-stacked bar does the job better.)

## Question for maintainers

Which approach fits DataLoom's intended architecture better — the **inline (A)** or the **dedicated view (B)**? I lean toward **B** because comparison is fundamentally a multi-dataset operation and a dedicated view scales better, but A is meaningfully less work and stays closer to the existing profiling flow. Happy to align on scope before starting.

**I'd like to work on this** once we've settled on the direction.

YADUNANDAN-SINGH:
## Description

This PR introduces a new **Compare** feature, allowing users to compare two datasets side-by-side in a dedicated workspace tab. It answers the analytical need of *"how do these datasets differ?"* without requiring users to constantly flip between open projects. 

Following **Option B** from the original issue proposal, this is structured as a dedicated comparison view to scale better for multi-dataset operations in the future. 

Key details:
- Integrated via the existing `featureRegistry.ts`.
- Two-pane layout with a custom resizable split divider.
- Mode switcher supports Data (Table), Summary, Columns, and Charts.
- Reuses existing components (`PreviewTable`, `ColumnProfileCard`, `ChartRenderer`) and synchronizes dual API calls seamlessly.

Fixes #429

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Verified that the `featureRegistry` seamlessly integrates the new tab without modifying core routing logic. Tested the custom flexbox properties to ensure horizontal scrolling works correctly in the `PreviewTable` within the flex layout constraint. Ran the full Vite build and test suite successfully to ensure no regressions.

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Screenshots 

image

image

image

image

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/DataScreen.jsx
+++ b/dataloom-frontend/src/Components/DataScreen.jsx
@@ -17,6 +17,7 @@ import "./workspace/features/profiling";
 import "./workspace/features/charts";
 import "./workspace/features/quality";
 import "./workspace/features/addFile";
+import "./workspace/features/compare";
 import WorkspaceTabBar from "./workspace/WorkspaceTabBar";
 import RightPanel from "./workspace/RightPanel";
 import MenuNavbar from "./MenuNavbar";
--- a/dataloom-frontend/src/Components/workspace/CompareTab.tsx
+++ b/dataloom-frontend/src/Components/workspace/CompareTab.tsx
@@ -1,0 +1,782 @@
+import {
+  useState,
+  useEffect,
+  useRef,
+  useCallback,
+  useMemo,
+  Suspense,
+  lazy,
+  type KeyboardEvent as ReactKeyboardEvent,
+} from "react";
+import type { WorkspaceTab } from "../../context/WorkspaceTabsContext";
+import { useProjectContext } from "../../context/ProjectContext";
+import { getProjects, getProjectDetails } from "../../api/projects";
+import { getChart } from "../../api/visualizations";
+import type { AggFunc, ChartParams, ChartSpec, ChartType } from "../../api/visualizations";
+import useDatasetSummary from "../../hooks/useDatasetSummary";
+import useColumnProfiles from "../../hooks/useColumnProfiles";
+import { isDatetime, isNumeric, usesAgg } from "../visualizations/chartFields";
+import { LuTable, LuLayoutDashboard, LuColumns3, LuChartColumnBig } from "react-icons/lu";
+import DtypeBadge from "../common/DtypeBadge";
+import ColumnProfileCard from "../profiling/ColumnProfileCard";
+import type { DatasetSummary } from "../../api/profiling";
+
+// Lazy import ChartRenderer to avoid bundling recharts in main chunk
+const ChartRenderer = lazy(() => import("../visualizations/ChartRenderer"));
+
+// Module-scoped so every compare selection across the whole session gets a
+// version the shared profiling cache (utils/profilingCache.ts) has never
+// seen, rather than a hardcoded constant that could collide with another
+// consumer's cached entry for the same project.
+let compareVersionCounter = 0;
+
+// eslint-disable-next-line react-refresh/only-export-components
+export const COMPARE_TAB: WorkspaceTab = {
+  id: "compare",
+  title: "Compare",
+  type: "compare",
+  closeable: true,
+};
+
+type Mode = "table" | "summary" | "columns" | "charts";
+type Cell = string | number | null | undefined;
+
+// SplitDivider component for resizing panes. Focusable and arrow-key
+// resizable per the WAI-ARIA APG separator pattern, not just mouse-draggable.
+function SplitDivider({
+  leftPct,
+  onDrag,
+  onResizeBy,
+}: {
+  leftPct: number;
+  onDrag: (clientX: number) => void;
+  onResizeBy: (deltaPct: number) => void;
+}) {
+  const dragging = useRef(false);
+
+  useEffect(() => {
+    const handleMove = (e: MouseEvent) => {
+      if (dragging.current) onDrag(e.clientX);
+    };
+    const handleUp = () => {
+      dragging.current = false;
+      document.body.style.cursor = "";
+      document.body.style.userSelect = "";
+    };
+    window.addEventListener("mousemove", handleMove);
+    window.addEventListener("mouseup", handleUp);
+    return () => {
+      window.removeEventListener("mousemove", handleMove);
+      window.removeEventListener("mouseup", handleUp);
+    };
+  }, [onDrag]);
+
+  const handleKeyDown = (e: ReactKeyboardEvent<HTMLDivElement>) => {
+    if (e.key === "ArrowLeft") {
+      e.preventDefault();
+      onResizeBy(-2);
+    } else if (e.key === "ArrowRight") {
+      e.preventDefault();
+      onResizeBy(2);
+    }
+  };
+
+  return (
+    <div
+      onMouseDown={() => {
+        dragging.current = true;
+        document.body.style.cursor = "col-resize";
+        document.body.style.userSelect = "none";
+      }}
+      onKeyDown={handleKeyDown}
+      tabIndex={0}
+      className="w-1.5 shrink-0 cursor-col-resize bg-gray-200 hover:bg-blue-300 focus:outline-none focus:ring-2 focus:ring-blue-400"
+      role="separator"
+      aria-orientation="vertical"
+      aria-valuenow={Math.round(leftPct)}
+      aria-valuemin={20}
+      aria-valuemax={80}
+      aria-label="Resize panes"
+    />
+  );
+}
+
+// ModeSwitcher Component
+function ModeSwitcher({ mode, setMode }: { mode: Mode; setMode: (m: Mode) => void }) {
+  const modes = [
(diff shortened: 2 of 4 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/workspace/CompareTab.tsx:251` The 'fetch right pane preview data' effect (useEffect at lines 246-256, fetch call `getProjectDetails(selectedProjectId, 1, 50)` on line 251) has no guard against out-of-order responses. If the user changes the compared dataset quickly (selects A, then B before A's request resolves), and A's response arrives after B's, `setPreview` on line 254 overwrites the correct B data with stale A data — silently showing the wrong dataset's table/columns/summary while the dropdown says B is selected. Needs an ignore-flag or AbortController tied to the effect cleanup.
- **[minor/quality]** `dataloom-frontend/src/Components/workspace/CompareTab.tsx:256` Both data-fetching effects swallow failures with `.catch(console.error)` only (line 242 for `getProjects`, line 256 for `getProjectDetails`) and never set an error state, so a failed request leaves the 'Compare with' dropdown silently empty or the right pane blank with zero feedback to the user. This is inconsistent with `handleRenderCharts` in the same file, which does surface failures via `leftChartResult.error`/`rightChartResult.error`.
- **[minor/test]** `dataloom-frontend/src/Components/workspace/CompareTab.tsx:1` The new ~480-line CompareTab component (two-pane layout, four render modes, two parallel data-fetching effects, chart comparison flow) ships with no accompanying test file, and the PR's own checklist leaves 'New tests added' unchecked.
- **[nit/style]** `dataloom-frontend/src/Components/Table.tsx:662` This is a pure reindentation (2-space to 6-space) unrelated to the Compare feature and is the only change to this file in the PR. It should be split into its own commit/PR to keep the diff focused on the stated change.

**Conventions**

PR #436 adds a new Compare workspace tab (CompareTab.tsx + a featureRegistry registration) that lets users view two datasets side-by-side across Table/Summary/Columns/Charts modes, wired cleanly through the existing feature-registry pattern and reusing existing profiling and chart components rather than duplicating them. The main functional risk is a race condition in the preview-fetch effect (CompareTab.tsx:251): switching the compared dataset quickly can leave stale data on screen because responses aren't guarded against arriving out of order, and both data-fetching effects fail silently with no user-facing error state. The new component is fairly large (~480 lines) and ships with no tests. Table.tsx also carries one unrelated indentation-only line that should be split into its own commit. None of this is disqualifying, but the race condition and silent error handling are worth fixing before merge.

YADUNANDAN-SINGH:
@ivantha Thanks for the review — all addressed:

- **Race condition** (`CompareTab.tsx:251`): both data-fetching effects now use an `ignore` flag cleared on cleanup, so a stale response from a previous selection can no longer overwrite the current one if requests resolve out of order.
- **Silent failures**: `getProjects` and `getProjectDetails` failures now set an error state that's surfaced in the UI (dropdown and right pane), consistent with how `handleRenderCharts` already reports chart errors.
- **Missing tests**: added `CompareTab.test.tsx` with 3 tests, including a regression test that reproduces the out-of-order-response race and asserts the stale data never appears.
- **Unrelated `Table.tsx` change**: removed. It was a merge artifact from rebasing against a stale branch, not an intentional change — there's now a zero diff to that file.

Also squashed to a single commit and rebased cleanly onto `main` per the CI bot's request.

Build, lint, and the full test suite (171 tests) all pass locally.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/workspace/CompareTab.tsx:299` The comparison pane's useDatasetSummary/useColumnProfiles calls (lines 299-303 and 310-314) pass a hardcoded dataVersion of 0 instead of the compared project's real dataVersion. profilingCache.ts stores a single entry per projectId keyed by whatever version was last written, using exact-equality matching, so two Compare-tab views of the same project both requesting the literal 0 collide on the same cache slot. If that project is edited directly (bumping its real dataVersion) between two Compare visits, the second visit silently serves the pre-edit cached summary/column profiles instead of refetching. Verified by reproducing against the real hook and cache modules: a second render passing the same hardcoded 0 returned the first render's stale value and made zero additional network calls, even though the mocked backend response had changed in between.
- **[minor/quality]** `dataloom-frontend/src/Components/workspace/CompareTab.tsx:431` `<ColumnProfileCard key={col} profile={profiles[col]} />` omits the `loading` prop, which ColumnProfileCardProps (profiling/ColumnProfileCard.tsx) declares as required (`loading: boolean`, not optional). The existing call site in Table.tsx always passes both `profile` and `loading`. This violates the component's typed contract under the repo's strict:true tsconfig, though it is currently harmless at runtime because profilesLoading is already checked before any card renders here, so the component's own loading branch would be unreachable regardless.
- **[minor/quality]** `dataloom-frontend/src/Components/workspace/CompareTab.tsx:50` SplitDivider is a mouse-only drag-to-resize control (role="separator", aria-orientation="vertical") with no tabIndex and no keyboard handler, so it cannot be focused or resized via keyboard. Per the WAI-ARIA APG, an interactive separator should be focusable and support arrow-key resizing. This is new code; there is no existing resizable-divider elsewhere in the codebase it could have inherited the gap from.
- **[minor/test]** `dataloom-frontend/src/test/CompareTab.test.tsx:59` The test suite (6 tests) only exercises the default Table mode and the Charts mode config panel. No test switches to Summary or Columns mode, so the stale-cache regression in useDatasetSummary/useColumnProfiles described above has no test that would have caught it.

**Conventions**

This PR adds a well-built Compare tab, a two-pane layout with a resizable divider and a Data/Summary/Columns/Charts mode switcher, that reuses existing profiling and charting infrastructure to let users inspect two datasets side by side. It ships as a single clean commit with a thorough description and screenshots, plus a solid targeted test suite that even covers an out-of-order-response race condition. The main issue to fix before merge is that the comparison pane hardcodes dataVersion=0 when fetching the compared project's cached summary/column profiles, which causes the shared profiling cache to silently serve stale data for that project once it has been edited elsewhere in the session (verified with a reproduction against the real hook and cache code). There is also a smaller compile-time nit, a required loading prop omitted on ColumnProfileCard, and the new resize divider has no keyboard alternative.

ivantha:
@YADUNANDAN-SINGH Let's hold on a bit before updating this PR.

YADUNANDAN-SINGH:
@ivantha 
 Okay 👍🏻, I'll hold off on any further changes and wait for your go-ahead before touching it again. Just let me know when you'd like me to pick it back up :)

PR #439, opened 2026-07-13
feat: auto-infer datetime columns on data load

### LINKED ISSUE #437: [Feature] Auto-detect datetime columns on CSV upload and data load
DataLoom already supports datetime as a dtype — `DtypeBadge` renders it in purple, `cast_data_type` can convert columns to datetime, and `DatasetSummaryResponse` has a `datetime_columns` field. However, columns containing date/time strings (e.g. `"2024-01-01"`, `"Jan 1 2024"`) are always read as `str` by pandas on CSV load, so they never get the `datetime` badge automatically. Users have to manually cast them via Cast Type every time.

After reading a DataFrame, attempt datetime inference on all `object`/`str` columns. If at least 80% of non-null values in a column parse successfully as datetime, convert the column to `datetime64` so `map_dtype` returns `"datetime"` and the purple badge appears automatically.

Implementation is a single helper `_infer_datetime_columns()` called inside `dataframe_to_response()` in `pandas_helpers.py`. No schema changes, no frontend changes, no migration — everything else is already wired.

Requiring users to manually cast every datetime column — rejected as poor UX, especially for datasets with many date columns.

Running inference on upload only — rejected since inference should apply consistently on every data load (upload, transform, revert, undo).

**Additional context**
- `map_dtype` in `pandas_helpers.py` already maps `dtype.kind == "M"` → `"datetime"`
- `DtypeBadge` already has the purple datetime style
- `DataType` enum already has `datetime`
- Only `pandas_helpers.py` needs changing — one new function, one line added to `dataframe_to_response`

SHIVAM-KUMAR-59:
## Description

Adds automatic datetime detection for uploaded datasets and improves datetime editing safety.

This change introduces automatic inference of datetime columns from string/object columns during file loading. A column is inferred as `datetime` when at least 80% of its non-null string values can be parsed successfully, ensuring consistent dtype detection across table rendering, profiling, and transformations.

Additionally, datetime values are now formatted for display without exposing unnecessary midnight timestamps:

- Date-only values display as `YYYY-MM-DD`
- Datetime values with time components display as `YYYY-MM-DD HH:MM:SS`
- The column dtype continues to be reported as `datetime`

Cell editing has also been improved by adding strict validation for datetime columns. Valid dates and timestamps are accepted, while incomplete or malformed values (e.g. `2024-05`, `2024-05.`, `hello`) are rejected instead of being silently normalized to unintended dates such as `2024-05-01`.

Fixes #437 

## Type of Change

- [x] New feature

## How Has This Been Tested?

**Added and updated tests covering:**

- Automatic datetime detection during file loading
- Datetime dtype mapping
- Date-only response formatting
- Datetime response formatting while preserving time components
- Preservation of null values in datetime columns
- Valid datetime and timestamp cell edits
- Rejection of incomplete and invalid datetime inputs
- Existing numeric, boolean, string, and null editing behavior

**Manual verification included:**

- Uploading datasets containing date columns and verifying automatic `datetime` detection
- Confirming date-only values render without `T00:00:00`
- Verifying timestamp values retain their time component
- Confirming invalid edits such as `2024-05` are rejected instead of being coerced to `2024-05-01`

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

## Screenshots

Screenshot 2026-07-14 at 03 09 05

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -4,11 +4,11 @@ Each function takes a DataFrame and parameters, returns a new DataFrame.
 No side effects -- saving to disk is handled by the caller.
 """
 
+import math
 from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 
-import numpy as np
 import pandas as pd
 
 from app.schemas import FillStrategy, MeltParams, OperationType
@@ -220,7 +220,12 @@ def delete_column(df: pd.DataFrame, index: int) -> pd.DataFrame:
     return df.drop(column_name, axis=1)
 
 
-def change_cell_value(df: pd.DataFrame, row_index: int, col_index: int, value) -> pd.DataFrame:
+def change_cell_value(
+    df: pd.DataFrame,
+    row_index: int,
+    col_index: int,
+    value,
+) -> pd.DataFrame:
     """Update a single cell value.
 
     Note: col_index is 1-based from the frontend (skipping S.No. column).
@@ -233,15 +238,20 @@ def change_cell_value(df: pd.DataFrame, row_index: int, col_index: int, value) -
 
     Returns:
         DataFrame with updated cell.
+
+    Raises:
+        TransformationError: If an index is out of bounds, a numeric value is
+            invalid or non-finite, a datetime value has an invalid format, or
+            pandas cannot store the converted value in the target column.
     """
     df = df.copy()
 
     # col_index is 1-based (1..len(columns)); row_index is 0-based. Guard both
     # ends so a negative index can't silently wrap to the wrong cell.
-    if row_index < 0 or row_index >= len(df) or col_index < 1 or col_index > len(df.columns):
+    if row_index < 0 or row_index >= len(df) or col_index < 1 or col_index >= len(df.columns) + 1:
         raise TransformationError("Row or column index out of bounds")
 
-    # col_index is 1-based from frontend (accounting for S.No. column).
+    # col_index is 1-based from frontend (accounting for S.No. column)
     column_name = df.columns[col_index - 1]
 
     # Cast value to match the column's existing dtype so pandas doesn't reject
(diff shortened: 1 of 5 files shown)


SHIVAM-KUMAR-59:

ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `dataloom-backend/app/utils/pandas_helpers.py:123` _infer_datetime_columns uses pd.to_datetime(..., format="mixed", errors="coerce") gated only by an 80% success-rate threshold, with no minimum-length or format sanity check. That means an ordinary non-date column of bare 4-digit years (e.g. a "grad_year" column with values "2024", "2025", "2023", ...) clears the 80% bar and gets silently converted to datetime64, with every value rewritten to January 1 of that year. I reproduced this by running the PR's own function against such a DataFrame: "2024" comes back as Timestamp("2024-01-01"). Because read_table_safe is the single ingestion path used by upload, every transform, save, revert, undo and export, and revert_to_checkpoint/export_project write the resulting DataFrame straight back to disk via save_table_safe -> df.to_csv(...), the fabricated value is persisted into the user's actual file on the next revert or export, not just shown differently in the UI.
- **[major/test]** `dataloom-backend/app/utils/pandas_helpers.py:84` No test in this PR exercises _infer_datetime_columns directly. The new tests in test_transformations.py only cover change_cell_value's per-cell validation, and the single change in test_dtypes.py just updates an existing formatting assertion. There is no test for the 80% threshold boundary, no test guarding against false positives on non-date columns (years, numeric IDs, short codes), and no test asserting that read_table_safe actually changes a column's dtype on load, which is exactly how the false-positive issue above went unnoticed.
- **[minor/bug]** `dataloom-backend/app/utils/pandas_helpers.py:160` _format_datetime_columns_for_response computes has_time once per column via .any() across all rows, then formats the entire column uniformly. If a column mixes date-only values with even one genuine timestamp (plausible when rows come from sources with different logging granularity), every row in that column, including the pure date-only ones, renders with a spurious "00:00:00" suffix, contradicting the PR's own stated goal that date-only values display as YYYY-MM-DD.
- **[minor/quality]** `dataloom-backend/app/utils/pandas_helpers.py:34` _infer_datetime_columns now runs on every read_table_safe call, which is the shared path used on essentially every request that touches a project (upload, each transform, save, revert, undo, export), not just initial upload. Columns that do convert are parsed twice with pd.to_datetime(..., format="mixed") (once for the success-rate probe, once for the real conversion), and format="mixed" disables pandas' fast vectorized parsing path, so this adds avoidable per-request latency proportional to the number of string columns instead of doing the inference once at upload time.
- **[nit/test]** `dataloom-backend/tests/test_transformations.py:428` The six new datetime cell-edit tests (test_datetime_cell_with_valid_date_preserves_dtype through test_clear_datetime_cell_stores_missing_value_and_preserves_dtype) are nested inside TestFillEmpty instead of TestChangeCellValue, even though all of them call change_cell_value and none call fill_empty. Purely organizational, but it will make these tests harder to find later.

**Conventions**

This PR adds automatic datetime dtype inference on data load, display formatting for datetime columns, and stricter validation on datetime cell edits, backed by a clear PR description and a single well-titled commit. The main risk is in _infer_datetime_columns: its 80% threshold with format="mixed" has no guard against short, ambiguous tokens, so I reproduced a case where an ordinary column of bare years (e.g. "2024") is silently coerced to datetime64 with a fabricated day and month, and that fabricated value is then actually persisted back to disk on the next revert or export via save_table_safe. None of the new tests exercise the inference function itself, only the cell-edit path, so this false-positive case has no coverage. I'd ask for a stronger inference heuristic (or an opt-out) plus direct tests for _infer_datetime_columns before merging.

SHIVAM-KUMAR-59:
Hi @ivantha ! I've addressed the review comments on this PR.

**Changes made:**

- Updated the datetime inference logic to avoid false positives by requiring values to match complete date/datetime patterns before attempting inference. This prevents columns containing bare years, numeric IDs, short codes, or partial dates from being incorrectly converted to `datetime64`.
- Added comprehensive tests covering:
 - valid datetime inference,
 - mixed datetime formats,
 - the 80% inference threshold,
 - false positives (bare years, numeric IDs, short codes, partial dates),
 - invalid date-shaped values,
 - immutability of the input DataFrame, and
 - `read_table_safe` integration to verify inference during file loading.

The PR should now be ready for another review. Thank you!

ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** needs-coordination

> Please coordinate with the maintainers before continuing. This is unlikely to be merged as-is.

**Findings**
- **[blocking/bug]** `dataloom-backend/app/utils/pandas_helpers.py:180` _looks_like_datetime's _TIME_SUFFIX regex (line 88) matches ISO timestamps carrying any UTC offset ('Z' or '+HH:MM'), so a column that mixes offsets across rows (e.g. most timestamps in 'Z', one in '+02:00' -- a realistic multi-region event-log export) gets passed to pd.to_datetime(..., format="mixed", errors="coerce"). pandas raises ValueError: "Mixed timezones detected. Pass utc=True..." for this, and errors="coerce" does not suppress it -- verified by running _infer_datetime_columns on such a column. read_table_safe calls this unconditionally and is the single read path used by upload, project view, transform, export, and checkpoint revert, so the exception surfaces as an HTTP 400 on every subsequent read of that project: the whole dataset becomes unreadable, not just the offending column.
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:290` The new datetime branch (elif at line 290) always builds a tz-naive pd.Timestamp, with no handling for tz-aware columns. _infer_datetime_columns in pandas_helpers.py can now produce a tz-aware column automatically (e.g. from uniformly 'Z'-suffixed timestamps), and the unconditional df.at[row_index, column_name] = value assignment at line 319 -- outside any try/except -- then raises an uncaught TypeError: "Invalid value '...' for dtype 'datetime64[us, UTC]'" for every edit on that column (verified directly with change_cell_value against a DataFrame carrying an inferred tz-aware column). It is not caught by the surrounding except (ValueError, TypeError, OverflowError) block, which has already exited by the time the assignment runs, so a valid edit crashes instead of succeeding or returning a clean TransformationError.
- **[major/bug]** `dataloom-backend/app/utils/pandas_helpers.py:182` format="mixed" (line 182) infers each row's day/month order independently instead of applying one convention to the whole column. For a genuinely day-first column, pd.to_datetime(pd.Series(['31/12/2024','01/02/2024','15/03/2024']), format='mixed', errors='coerce') returns [2024-12-31, 2024-01-02, 2024-03-15]: the middle value (meant as 1 Feb, consistent DD/MM/YYYY with its neighbors) is silently reinterpreted as 2 Jan. No NaT/error results, so parse_success_rate counts it as a success and the column converts with corrupted values for every ambiguous (day<=12) row.
- **[minor/quality]** `dataloom-backend/app/utils/pandas_helpers.py:161` _infer_datetime_columns runs three separate Python-level row-wise .map() passes per candidate string column (isinstance check at line 161, strip+regex-fullmatch at lines 172-173, a second strip at line 178) instead of vectorized pandas string ops. Harmless at the current small upload-size cap, but it's an O(rows) interpreter loop repeated per object/string column and will show up on wider or longer datasets.

**Conventions**

This PR adds automatic datetime-column detection on data load, cleaner date/timestamp display formatting, and stricter validation for datetime cell edits, backed by thorough new tests; the full backend test suite (631 tests) and ruff lint/format both pass cleanly on this branch. However, direct testing surfaced two real bugs in the core _infer_datetime_columns helper: a column mixing timezone offsets (e.g. 'Z' and '+02:00' rows in the same column) raises an uncaught ValueError that becomes an HTTP 400 on every future read of that project, and format="mixed" silently misinterprets ambiguous day-first dates (e.g. '01/02/2024') inconsistently within one column with no error or warning. A related bug in the new cell-edit validation crashes with an unhandled TypeError when editing any cell in a timezone-aware datetime column, which this same PR's inference logic can now produce. All three should be fixed and covered by tests before merge.

SHIVAM-KUMAR-59:
Thanks @ivantha for the review! I've addressed all three issues in the latest commit.

Changes made:

- Added consistent day/month inference for numeric date formats to avoid row-by-row ambiguity and skip inference when conflicting conventions are detected.
- Handled mixed UTC offsets during datetime inference by falling back to UTC normalization, preventing datasets from becoming unreadable.
- Updated `change_cell_value` to correctly support timezone-aware datetime columns by localizing edits to the column's timezone and converting assignment failures into `TransformationErrors` instead of uncaught pandas exceptions.

Regarding the scope, I had discussed these changes with @hanzalahwaheed before opening this PR and got his approval to include them here, since they were required to address the datetime inference behavior introduced by this feature.

Thanks again for the review!

PR #441, opened 2026-07-14
fix: prevent unchanged cell edits from being persisted

### LINKED ISSUE #440: [Bug] Prevent unchanged cell edits from creating change logs
**Describe the bug**

When a user clicks a table cell, enters edit mode, and presses `Enter` without changing the cell value, the frontend still sends a transformation request.

The backend then processes the request, rewrites the project file, creates a change log entry, and updates the project even though no actual data change occurred.

This results in unnecessary requests, misleading transformation history, and incorrect project modification state.

**To Reproduce**

1. Open a project containing tabular data.
2. Click on any editable cell.
3. Do not modify the existing value.
4. Press `Enter`.
5. Open the transformation logs or inspect the network request.
6. Observe that a transformation request is sent and a change log is created.

**Expected behavior**

Pressing `Enter` without changing the cell value should simply exit edit mode.

The frontend should not send a transformation request when the edited value is unchanged.

The backend should also detect no-op cell transformations and avoid:

- Rewriting the project file
- Creating a transformation log
- Updating the project's last-modified timestamp

**Screenshots**

Not applicable.

**Environment:**

- OS: any
- Browser: any
- Node.js version: any
- Python version: any

**Additional context**

The validation should be implemented on both sides:

- The frontend should compare the edited value with the original displayed value before sending the request.
- The backend should compare the original and transformed DataFrames before persisting or logging the transformation.

SHIVAM-KUMAR-59:
## Description

Prevents unchanged cell edits from being processed as transformations.

Previously, clicking a cell and pressing `Enter` without modifying its value still sent a request to the backend. The backend then rewrote the project file and created a transformation log even though the data remained unchanged.

This change adds safeguards at both levels:

- The frontend compares the edited value with the original cell value and skips the request when they are equal.
- The backend detects no-op cell transformations and skips file persistence and change-log creation.

This avoids unnecessary network requests, file writes, misleading transformation history, and incorrect project modification updates.

Fixes #440 

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Manually verified the following scenarios:

- Clicking a cell and pressing `Enter` without changing the value does not send a request.
- No change log is created for an unchanged cell value.
- The project file is not rewritten for a no-op edit.
- Changing a cell value still sends one transformation request.
- A valid cell edit still updates the table and creates a change log.
- Changing a value and then restoring the original value before pressing `Enter` is treated as a no-op.
- Pressing `Escape` exits edit mode without sending a request.

- [x] Existing tests pass
- [x] Manual testing

Not applicable.

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/api/endpoints/transformations.py
+++ b/dataloom-backend/app/api/endpoints/transformations.py
@@ -122,7 +122,13 @@ async def transform_project(
 
         result_df = _dispatch_transform(df, transformation_input)
 
-        if not preview:
+        # A cell edit that produces the same DataFrame is a no-op. Return the current
+        # data without rewriting the file or creating a transformation log.
+        is_noop_cell_edit = (
+            transformation_input.operation_type == schemas.OperationType.changeCellValue and result_df.equals(df)
+        )
+
+        if not preview and not is_noop_cell_edit:
             save_table_safe(result_df, project.file_path)
             try:
                 log_transformation(db, project_id, operation_type, transformation_input.dict())
--- a/dataloom-frontend/src/Components/Table.tsx
+++ b/dataloom-frontend/src/Components/Table.tsx
@@ -305,7 +305,21 @@ const Table = ({ projectId, showColumnProfiles = false }: TableProps) => {
     }
   };
 
+  const stopEditing = () => {
+    setEditingCell(null);
+    setEditValue("");
+  };
+
   const handleEditCell = async (rowIndex: number, cellIndex: number, newValue: string) => {
+    const currentValue = data[rowIndex]?.[cellIndex];
+    const normalizedCurrentValue = currentValue == null ? "" : String(currentValue);
+
+    // Do not send a transformation request when the value was not changed.
+    if (newValue === normalizedCurrentValue) {
+      stopEditing();
+      return;
+    }
+
     try {
       const backendColIndex =
         cellIndex === 0
@@ -322,10 +336,10 @@ const Table = ({ projectId, showColumnProfiles = false }: TableProps) => {
           fill_value: newValue,
         },
       });
+
       updateTableData(response);
       refreshProject(projectId, 1, pageSize);
-      setEditingCell(null);
-      setEditValue("");
+      stopEditing();
     } catch {
       setToast({
         message: "Failed to edit cell. Please try again.",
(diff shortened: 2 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- **[minor/test]** `dataloom-backend/app/api/endpoints/transformations.py:131` The new no-op-skip branch (bypassing save_table_safe/log_transformation when is_noop_cell_edit is true) has no accompanying automated test, even though this endpoint already has dedicated coverage in test_transformations.py and test_transform_http_semantics.py. The PR description lists only manual verification steps, so nothing in CI guards this behavior against future regression.
- **[nit/quality]** `dataloom-frontend/src/Components/Table.tsx:361` In the Escape-key branch, the pre-existing `setEditingCell(null); setEditValue("")` lines were left in place and a redundant `stopEditing()` call was appended right after them, instead of replacing the two lines with the call. The new helper just repeats what the lines above it already did, undercutting the point of extracting it.
- **[nit/bug]** `dataloom-backend/app/api/endpoints/transformations.py:128` `result_df.equals(df)` also compares dtypes, and change_cell_value (transformation_service.py) unconditionally upcasts a numeric column to `object` whenever the submitted value is empty/None, even if that cell was already null. A no-op empty-to-empty edit on a numeric column would therefore fail this equality check and still get persisted/logged. It's only reachable by calling the API directly, since the frontend guard already filters this exact case out of the UI path, but the backend safeguard alone isn't airtight for it.
- **[nit/style]** `dataloom-frontend/src/Components/Table.tsx:677` Re-indents the TablePagination wrapper `<div>` — an unrelated formatting fix bundled into a bug-fix PR that has nothing to do with cell editing, adding diff noise unrelated to the stated change.

**Conventions**

This PR fixes a genuine bug: unchanged cell edits were previously always sent to the backend, which rewrote the project file and added a spurious transformation-log entry every time. The fix is properly layered — a frontend value comparison avoids the network call, and a backend result_df.equals(df) check scoped to changeCellValue acts as defense-in-depth — and the value-coercion logic between the two guards lines up correctly for the common cases. Minor cleanup items: the Escape-key handler keeps a redundant stopEditing() call alongside the state resets it was meant to replace, an unrelated indentation fix in TablePagination is bundled into the diff, and the new skip-persistence branch has no accompanying automated test despite the backend's existing transformation test suite. None of these block the fix from working correctly for the reported issue.

PR #443, opened 2026-07-14
perf: run blocking data endpoints in a threadpool, not on the event loop

anxkhn:
The data-heavy routes are declared `async def` while doing synchronous, blocking
pandas file I/O (`read_table_safe` -> `pd.read_csv/read_excel/read_parquet`,
`save_table_safe` -> `df.to_csv/to_excel`) and CPU-bound pandas work. In
Starlette an `async def` path operation runs directly on the asyncio event loop,
so a single request parsing a CSV/XLSX (up to the 10MB upload limit, which can
take hundreds of milliseconds to seconds) blocks every other concurrent request
for that whole duration. `GET /projects/get/{id}` is the clearest case: it reads
the entire file on every paginated page view, then slices in memory.

FastAPI runs plain `def` path operations in an external threadpool for exactly
this reason, keeping the loop free. This PR moves the blocking work off the event
loop with the smallest change that preserves behavior:

- Convert the routes that perform blocking I/O and hold no `await` from
  `async def` to plain `def`, so FastAPI runs them in its threadpool:
  - `projects.py`: `get_project_details`, `save_project`, `revert_to_checkpoint`,
    `export_project`, `undo_last_transformation`
  - `transformations.py`: `transform_project`
  - `profiling.py`: `get_dataset_summary`, `get_column_profile`,
    `get_all_column_profiles`, `get_correlation_matrix`
- `upload_project` must stay `async def` because it `await`s upload validation,
  so wrap its blocking `store_upload` (disk write) and `read_table_safe` (parse)
  calls in `starlette.concurrency.run_in_threadpool` instead.

`rename_project_endpoint` and `delete_project_endpoint` are intentionally left
`async def`: they touch only DB metadata and do no pandas file I/O, so keeping
the diff off them stays focused on the actual problem.

This is a pure concurrency fix with no change to request/response behavior.

No existing issue tracks this, so there is no `Fixes #` reference.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Added `dataloom-backend/tests/test_endpoint_concurrency.py` with three tests:

- Structural: the ten blocking routes are plain `def` (not coroutines), so FastAPI
  offloads them to the threadpool.
- Structural: `upload_project` stays `async def` and its body offloads both
  `store_upload` and `read_table_safe` via `run_in_threadpool`.
- Behavioral: with `read_table_safe` patched to sleep, five overlapping
  `GET /projects/get/{id}` requests must finish well under the serialized floor
  (they overlap in the threadpool instead of running one after another on the loop).

Verified red -> green: on the pre-fix `async def` code the behavioral test
serializes five 0.4s reads to about 2.1s and all three tests fail; after the fix
they overlap and finish in about 0.5s and pass.

Commands (from `dataloom-backend/`, Python 3.12, `DATABASE_URL=sqlite:///./test.db`):

- `uv run ruff check .` -> All checks passed
- `uv run ruff format --check .` -> already formatted
- `uv run pytest` -> 482 passed (the pre-existing SQLModel `session.query`
  deprecation warnings are unrelated to this change)

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

N/A (backend-only concurrency change).

## Checklist

- [x] I have added/updated documentation as needed (no docs change needed; behavior is unchanged)
- [x] Tests pass locally

---

## Files changed (diff = origin/main..HEAD)

```
dataloom-backend/app/api/endpoints/profiling.py       |   8 +-
dataloom-backend/app/api/endpoints/projects.py        |  15 +--
dataloom-backend/app/api/endpoints/transformations.py |   2 +-
dataloom-backend/tests/test_endpoint_concurrency.py   | 106 +++++++++++++++++++
4 files changed, 119 insertions(+), 12 deletions(-)
```

### DIFF
--- a/dataloom-backend/app/api/dependencies.py
+++ b/dataloom-backend/app/api/dependencies.py
@@ -13,7 +13,7 @@ from app.config import get_settings
 from app.services import auth_service
 from app.utils.logging import get_logger
 from app.utils.pandas_helpers import read_table_safe
-from app.utils.project_locks import project_write_lock
+from app.utils.project_locks import project_read_lock
 from app.utils.rate_limiter import RateLimiter
 
 logger = get_logger(__name__)
@@ -159,18 +159,36 @@ def get_project_or_404(
     return fetch_owned_project(db, project_id, current_user)
 
 
-def load_project_df(project: models.Project) -> pd.DataFrame:
+def read_project_df(project: models.Project) -> pd.DataFrame:
     """Read a project's working copy, redacting any path from error details.
 
     ``read_table_safe`` embeds the absolute file path in its 404/500 details;
     surface a generic message to the client instead, matching the transform
-    endpoint's handling. Shared by the profiling and visualization endpoints.
+    endpoint's handling.
+
+    Takes no project lock, so it is safe to call from code already holding
+    :func:`app.utils.project_locks.project_write_lock` (which is not reentrant).
+    Unlocked callers should use :func:`load_project_df` instead.
     """
     try:
-        with project_write_lock(project.project_id):
-            return read_table_safe(project.file_path)
+        return read_table_safe(project.file_path)
     except HTTPException as e:
         if e.status_code == 404:
             raise HTTPException(status_code=404, detail="Project data file not found") from e
         logger.warning("Failed to read project file project_id=%s status=%s", project.project_id, e.status_code)
         raise HTTPException(status_code=e.status_code, detail="Could not read project data") from e
+
+
+def load_project_df(project: models.Project) -> pd.DataFrame:
+    """Read a project's working copy under the shared project read lock.
+
+    The lock keeps the read from landing mid-write while a save, transform,
+    revert, undo, append or pipeline run rewrites the file in place. It is
+    shared, so concurrent readers of the same project still overlap.
+
+    Blocks the calling thread, so callers must be plain ``def`` path operations
+    (run in FastAPI's threadpool) rather than ``async def`` ones. Shared by the
+    profiling, visualization, reporting, quality, pipeline and append endpoints.
+    """
+    with project_read_lock(project.project_id):
+        return read_project_df(project)
--- a/dataloom-backend/app/api/dependencies.py
+++ b/dataloom-backend/app/api/dependencies.py
@@ -13,6 +13,7 @@ from app.config import get_settings
 from app.services import auth_service
 from app.utils.logging import get_logger
 from app.utils.pandas_helpers import read_table_safe
+from app.utils.project_locks import project_write_lock
 from app.utils.rate_limiter import RateLimiter
 
 logger = get_logger(__name__)
@@ -166,7 +167,8 @@ def load_project_df(project: models.Project) -> pd.DataFrame:
     endpoint's handling. Shared by the profiling and visualization endpoints.
     """
     try:
-        return read_table_safe(project.file_path)
+        with project_write_lock(project.project_id):
+            return read_table_safe(project.file_path)
     except HTTPException as e:
         if e.status_code == 404:
             raise HTTPException(status_code=404, detail="Project data file not found") from e
--- a/dataloom-backend/app/api/endpoints/pipelines.py
+++ b/dataloom-backend/app/api/endpoints/pipelines.py
@@ -12,10 +12,12 @@ from app.api.dependencies import (
     fetch_owned_project,
     get_current_user,
     load_project_df,
+    read_project_df,
 )
 from app.services import pipeline_service
 from app.services.transformation_service import TransformationError
 from app.utils.pandas_helpers import dataframe_to_response
+from app.utils.project_locks import project_write_lock
 from app.utils.security import safe_transformation_error_detail
 
 router = APIRouter()
@@ -99,9 +101,14 @@ def apply_pipeline(
     pipeline = fetch_owned_pipeline(db, pipeline_id, current_user)
     project = fetch_owned_project(db, body.project_id, current_user)
 
-    df = load_project_df(project)
+    # Applying a pipeline is a read-modify-write of the working copy, so the read
+    # and the write share one exclusive lock; splitting them would let a
+    # concurrent transform slip in between and lose an update. The write lock is
+    # not reentrant, hence read_project_df rather than load_project_df.
     try:
-        result_df = pipeline_service.apply_pipeline_to_project(db, project, pipeline, df)
+        with project_write_lock(project.project_id):
+            df = read_project_df(project)
+            result_df = pipeline_service.apply_pipeline_to_project(db, project, pipeline, df)
     except TransformationError as e:
         raise HTTPException(status_code=400, detail=safe_transformation_error_detail(e)) from e
 
--- a/dataloom-backend/app/api/endpoints/profiling.py
+++ b/dataloom-backend/app/api/endpoints/profiling.py
@@ -17,7 +17,7 @@ router = APIRouter()
 
 
 @router.get("/{project_id}/profile/summary", response_model=schemas.DatasetSummaryResponse)
-async def get_dataset_summary(
+def get_dataset_summary(
     project_id: uuid.UUID,
     project: models.Project = Depends(get_project_or_404),
 ):
@@ -27,7 +27,7 @@ async def get_dataset_summary(
 
 
 @router.get("/{project_id}/profile/column", response_model=schemas.ColumnProfileResponse)
-async def get_column_profile(
+def get_column_profile(
     project_id: uuid.UUID,
     column_name: str = Query(..., description="Name of the column to profile"),
     project: models.Project = Depends(get_project_or_404),
@@ -47,7 +47,7 @@ async def get_column_profile(
 
 
 @router.get("/{project_id}/profile/columns", response_model=schemas.ColumnProfilesResponse)
-async def get_all_column_profiles(
+def get_all_column_profiles(
     project_id: uuid.UUID,
     project: models.Project = Depends(get_project_or_404),
 ):
@@ -62,7 +62,7 @@ async def get_all_column_profiles(
 
 
 @router.get("/{project_id}/profile/correlation", response_model=schemas.CorrelationResponse)
-async def get_correlation_matrix(
+def get_correlation_matrix(
     project_id: uuid.UUID,
     project: models.Project = Depends(get_project_or_404),
 ):
--- a/dataloom-backend/app/api/endpoints/project_files.py
+++ b/dataloom-backend/app/api/endpoints/project_files.py
@@ -19,9 +19,10 @@ from pathlib import Path
 
 from fastapi import APIRouter, Depends, File, HTTPException, UploadFile
 from sqlmodel import Session
+from starlette.concurrency import run_in_threadpool
 
 from app import database, models, schemas
-from app.api.dependencies import get_project_or_404, load_project_df
+from app.api.dependencies import get_project_or_404, load_project_df, read_project_df
 from app.services.append_service import analyze_append, append_dataframes
 from app.services.file_service import store_added_file
 from app.services.project_service import (
@@ -32,6 +33,7 @@ from app.services.project_service import (
 )
 from app.utils.logging import get_logger
 from app.utils.pandas_helpers import dataframe_to_response, read_table_safe, save_table_safe
+from app.utils.project_locks import project_write_lock
 from app.utils.security import validate_upload_file
 
 logger = get_logger(__name__)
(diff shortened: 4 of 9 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/api/endpoints/transformations.py:104` Converting transform_project (and analogously save_project, revert_to_checkpoint, undo_last_transformation in projects.py) from async def to plain def removes the incidental serialization that came from blocking a single-threaded event loop. These handlers still do read-file -> mutate -> write-file on project.file_path (see lines 121/126/133) with no per-project lock or optimistic-concurrency check. Previously, a blocking async def call froze the whole event loop, which accidentally prevented two requests for the same project from interleaving; now that the same code runs in a genuine threadpool, two truly concurrent requests against the same project_id (double-submit, two open tabs, a save racing an undo) can interleave and produce a lost update. The PR's own new test only exercises concurrency on a read-only route (get_project_details); nothing covers concurrent writes to the same project.
- **[minor/test]** `dataloom-backend/tests/test_endpoint_concurrency.py:103` test_slow_read_does_not_block_other_requests asserts on wall-clock elapsed time (elapsed < read_delay * concurrency * 0.6, i.e. ~1.2s budget vs a ~2.0s serialized floor). The margin is generous enough that it's unlikely to flake, but timing-based assertions are inherently sensitive to CI runner load and threadpool warmup; worth keeping an eye on if this test ever goes flaky, since a small structural/behavioral check wouldn't depend on wall-clock thresholds.

**Conventions**

This is a well-scoped, well-tested backend concurrency fix: it moves ten route handlers that were doing blocking pandas/file I/O directly inside async def bodies (silently stalling the whole event loop for every other request) to plain def so FastAPI offloads them to its threadpool, and wraps upload_project's two blocking calls in run_in_threadpool since that route must stay async. The new test file pins the contract both structurally (routes are sync; upload offloads correctly) and behaviorally (concurrent requests no longer serialize), and CI (backend lint, backend tests, e2e) is green aside from a pre-existing, unrelated frontend-lint formatting failure. The one thing worth a second look before merge: this same change removes the incidental serialization that was masking a lost-update race when transform_project/save_project/revert_to_checkpoint/undo_last_transformation are hit concurrently for the same project, and no test covers that path -- worth a follow-up (per-project lock or optimistic concurrency) even if it doesn't block this PR. PR title and description are exemplary (clear Conventional Commit title, thorough rationale and scoping notes, explicit red/green testing methodology), and the single commit is clean.

anxkhn:
addressed the race in 09869b3. mutating operations now take a per-project lock across their file and change-log work, while different projects and read-only previews remain concurrent. added a regression test that runs two same-project transforms concurrently and verifies both updates survive.

also rebased onto current main and confirmed the frontend prettier check, frontend lint, backend ruff, and all 648 backend tests pass locally. @ivantha could you please re-review?

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/api/endpoints/projects.py:118` Preview-mode transform_project (unlocked fallback) plus the read-only routes get_project_details (projects.py:118), export_project (projects.py:319), and the four profiling routes (profiling.py) never acquire project_write_lock, yet they all read the same project.file_path that save_project, revert_to_checkpoint, undo_last_transformation, and non-preview transform_project now write to from a genuinely separate threadpool thread. Because file_formats.py's writers (_write_csv, _write_json, _write_xlsx, _write_parquet) truncate and rewrite the destination path in place instead of writing atomically (temp file plus rename), a concurrently running unlocked reader can now observe a partially written file mid-write, which read_table_safe surfaces as a 400 or 500 to the client, and export_project's native-format FileResponse can stream a torn mix of old and new bytes. This race was structurally impossible before this PR, since every blocking call previously ran serialized on the single event-loop thread; it only becomes reachable because this PR moves these handlers onto real OS threads, and none of the PR's new tests exercise a concurrent read against an in-flight write.
- **[minor/quality]** `dataloom-backend/app/api/endpoints/transformations.py:121` Unlike save_project (projects.py:196-210), which scopes project_write_lock tightly around just the file read and checkpoint creation, transform_project pulls its entire body, including pagination slicing and dataframe_to_response construction, inside the lock by calling _transform_project from within the with project_write_lock block; revert_to_checkpoint (projects.py:243) and undo_last_transformation (projects.py:438) do the same via their own helpers. None of that trailing in-memory work touches the shared file or database row, so holding the lock through it needlessly extends the critical section and reduces the concurrency win this PR is aiming for whenever two requests target the same project back to back.
- **[nit/style]** `dataloom-backend/app/api/endpoints/transformations.py:125` The extracted helper _transform_project(project_id, transformation_input, preview, page, page_size, db, project) drops every type hint that transform_project's original FastAPI parameter declarations carried (uuid.UUID, schemas.TransformationInput, bool, int, Session, models.Project). The same pattern repeats in projects.py's new _revert_to_checkpoint (line 247) and _undo_last_transformation (line 442), inconsistent with the rest of the typed codebase, e.g. _unlink_if_exists(path: str) -> None a few lines above in projects.py.

**Conventions**

This PR correctly diagnoses and fixes a real bug: ten data-heavy routes ran blocking pandas I/O directly on the asyncio event loop, so converting them to plain def and offloading upload_project's blocking calls via run_in_threadpool is the right fix, and the new project_write_lock is a necessary, correctly implemented companion change that prevents lost-update races between concurrent writers on the same project. However, the lock only guards the write endpoints: get_project_details, export_project, the four profiling routes, and preview-mode transform_project still read the same working-copy file unlocked, and since file_formats.py's writers truncate and rewrite the file in place rather than atomically, a concurrently running read can now observe a partially written file and return a spurious 400 or 500, a new failure mode this PR's own tests do not cover. Recommend closing that read or write gap (or making the writers atomic via temp file plus rename) before merging. The lock-scope inconsistency between save_project's tight critical section and the other three endpoints holding the lock for their entire body, including pure response construction, is a smaller, non-blocking follow-up.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/api/endpoints/projects.py:124` get_project_details reads project.file_path without taking project_write_lock, while save_project/revert_to_checkpoint/undo_last_transformation/transform_project now hold that lock only while writing the same path from a genuinely concurrent OS thread (routes are plain def, run via FastAPI's threadpool). The four format writers in app/utils/file_formats.py (_write_csv/_write_xlsx/_write_json/_write_parquet) all open the target path directly and truncate it immediately, with no temp-file-plus-atomic-rename, so a concurrent read can now land mid-write and see a truncated or partial file (parse error, or in principle malformed data) — a race window that was effectively impossible before this PR because every blocking call ran to completion on the single event-loop thread before anything else could run. The same unlocked-read-vs-locked-write gap applies to profiling.py's four endpoints (via load_project_df), export_project's read path, and transform_project when preview=True.
- **[minor/quality]** `dataloom-backend/app/api/endpoints/transformations.py:82` transform_project duplicates the call to _transform_project(project_id, transformation_input, preview, page, page_size, db, project) across both branches of `if not preview` (lines 83-85) solely to decide whether project_write_lock is held. A conditional context manager (e.g. contextlib.nullcontext() when preview is true) would express the same branching without repeating the 7-argument call.
- **[nit/test]** `dataloom-backend/tests/test_endpoint_concurrency.py:143` project_write_lock's docstring promises it serializes 'without blocking other projects,' but no test exercises two distinct project_ids concurrently to confirm that. test_slow_read_does_not_block_other_requests reuses a single pid for all 5 concurrent requests, and it calls get_project_details, which never acquires project_write_lock in the first place, so the cross-project-independence claim is untested.

**Conventions**

This PR converts ten routes doing blocking pandas file I/O from async def to plain def so FastAPI runs them in its threadpool instead of the event loop, wraps upload_project's blocking calls in run_in_threadpool, and adds a new per-project threading.Lock (project_write_lock) around the four write endpoints (save, revert, undo, non-preview transform) to prevent lost updates now that those writes can run truly concurrently. The core fix is sound and well tested (structural checks plus a behavioral overlap test), but the new lock only covers writer-vs-writer races: read-only routes (get_project_details, all four profiling endpoints, export, and preview-mode transform) still read the same file path unlocked, and since the format writers truncate-and-write the target path directly rather than via temp-file-plus-rename, a read can now race an in-flight write and see a partial file — a window that was effectively closed before this PR by the single event-loop thread. Recommend closing that read/write gap (e.g., a shared/read variant of the lock, or making saves atomic via temp file + os.replace) before or shortly after merge. Two minor cleanups are also worth a look: transform_project duplicates its dispatch call across both branches of the preview check, and the lock's "doesn't block other projects" guarantee isn't directly tested.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/api/dependencies.py:170` load_project_df now acquires the blocking project_write_lock before every read, but visualizations.py's suggest_charts/get_chart (and project_files.py's preview_add_file) are still `async def` handlers that call it directly on the event loop instead of through the threadpool. If a threadpool-run writer (save_project, transform_project, revert_to_checkpoint, undo_last_transformation, export_project) holds that project's lock, these chart/preview endpoints block lock.acquire() on the event-loop thread itself, freezing request handling for every project and every user until the writer finishes -- a worse stall than the single blocking read that existed pre-PR, and it undercuts the exact goal this PR sets out to achieve. Example: client A's transform on project X holds project_write_lock while it reads+applies+saves; while that write is in flight, any client hitting GET /{project}/charts or /charts/suggest for any project blocks the whole server until A's write finishes.
- **[major/bug]** `dataloom-backend/app/utils/project_locks.py:16` project_locks.py's docstrings claim project_write_lock makes it so 'a threadpool reader cannot observe a torn in-place write,' but that guarantee is incomplete: project_files.py's _append_and_log (used by POST /{project}/files and POST /{project}/files/{id}/append) writes the working copy via save_table_safe(combined, project.file_path) at line 82 without ever acquiring project_write_lock. A concurrent profiling/projects/transformations request that correctly takes the lock before reading project.file_path can still race this unlocked writer and observe a partially written file, since the append writer never registers with the lock at all.
- **[minor/quality]** `dataloom-backend/app/utils/project_locks.py:13` project_write_lock is a plain mutex rather than a reader/writer lock, so independent read-only requests to the same project now fully serialize through the same critical section that guards writes. All four profiling endpoints and get_project_details funnel through this one lock via load_project_df/direct use, so a client opening several profiling panels at once for one project (a normal usage pattern: /profile/summary, /profile/columns, /profile/correlation firing together) gets no more concurrency after this change than before it -- they now queue one at a time instead of running concurrently in the threadpool, muting the latency win for the most common multi-widget-load case. Only cross-project concurrency actually improves.

**Conventions**

This PR correctly converts ten blocking-I/O routes from async def to plain def so FastAPI runs them in a threadpool, and adds a per-project lock to keep concurrent readers/writers of a project's working copy from tearing -- a real and previously-latent bug once true parallelism is introduced. However, the new load_project_df lock is called synchronously from async def handlers in visualizations.py (and project_files.py's preview_add_file) that were not converted, so a write on one project can now stall the entire event loop for all users until it releases -- worse than the pre-PR behavior in that specific path. Separately, project_files.py's append-file write path (_append_and_log) mutates project.file_path without ever taking project_write_lock, so the "no torn writes" guarantee the new lock advertises does not actually hold across the whole codebase. Recommend either extending the lock/threadpool conversion to visualizations.py and project_files.py in this PR, or scoping the docstring claims down to the files actually covered, before merging.

ivantha:
Pushed a fix commit (`61072fa`) addressing the three findings from the last review, plus two related things I found while verifying. Happy to discuss any of it if you'd rather take a different approach.

**Root cause of all three findings:** the lock was a plain mutex shared by readers and writers, which conflates two different guarantees ("don't tear a file mid-write" and "don't lose an update"). Split into a writer-preferring readers-writer lock in `project_locks.py`: `project_read_lock` is shared, `project_write_lock` stays exclusive. Writer preference is deliberate, since the profiling panels poll continuously and a reader-preferring lock would let them starve a save.

**1. Async routes blocking the event loop.** `suggest_charts` / `get_chart` and `reappend_project_file` await nothing, so they're now plain `def`. `preview_add_file` and `add_file` await validation, so they stay async and offload via `run_in_threadpool`, same pattern as `upload_project`.

**2. Unlocked writers.** `_append_and_log` now holds the write lock across its whole read-modify-write, reading through a new unlocked `read_project_df` (the lock isn't reentrant, so `load_project_df` would deadlock there).

**3. Readers serializing.** Read-only paths (`get_project_details`, `export_project`, `load_project_df`, preview transforms) take the shared lock, so concurrent profiling requests for one project overlap again.

**Two extra things:**

- `pipelines.apply_pipeline` was a third unlocked read-modify-write: it read via `load_project_df`, released, then wrote through `apply_pipeline_to_project`. Same lost-update class as `_append_and_log`, now under one exclusive lock.
- `test_concurrent_transforms_for_same_project_do_not_lose_updates` couldn't fail. The fixture read the shared frame *after* its `sleep`, so the "slow" thread picked up fresh data. I confirmed it passed with `project_write_lock` removed from `transform_project` entirely. Snapshotting before the stall makes it detect the race it was guarding.

**On verification:** I mutation-tested each assertion rather than trusting a green run, reverting one fix at a time to confirm the matching test goes red. Two of my own tests initially failed that check for the same reason as the one above. New coverage: shared reads, cross-project independence, writer preference, registry cleanup, the append and pipeline write paths, and an AST sweep that fails if *any* async route reaches blocking I/O or the project lock without offloading (it scans all 10 async routes across every endpoint module, so a new endpoint can't reintroduce this). The overlap tests use `threading.Barrier` rather than wall-clock thresholds, which also addresses the earlier flakiness note: a serializing lock raises `BrokenBarrierError` instead of just running slow.

`ruff check` and `ruff format --check` clean, 757 backend tests pass.

**One design note:** I considered atomic writes (temp file plus `os.replace`), which would let readers skip locking entirely and is the more elegant fix. I ruled it out because `SETUP.md` documents Windows 11 as the local dev platform, and on Windows `os.replace` fails with a sharing violation while a reader holds the file open, so writes would fail intermittently there. If Windows isn't actually a target, atomic writes are worth revisiting as a follow-up.

PR #444, opened 2026-07-14
feat: add data quality assessment feature

### LINKED ISSUE #434: [Feature] Add Data Quality metrics
DataLoom profiles data but can't judge it. No automated detection of duplicates, outliers, type mismatches, or format issues.

A quality service with six detectors, severity-weighted 0–100 score, persisted reports, remediation suggestions, and a QualityPanel UI.

Extending the profiling service directly, I've rejected it to keep the descriptive contract stable and scoring independently tunable.

**Additional context**
Weeks 7 – 8 of GSoC 2026 work. Detectors map to completeness, uniqueness, validity, consistency. Remediations reuse existing transforms.

hanzalahwaheed:
## Description

Adds the Week 7/8 data quality assessment feature: a detection layer that scans a project's working copy for quality issues, scores it 0–100, and suggests fixes that plug into the existing transform flow.

- quality_service with six pure detectors (duplicates, missing values, IQR/z-score outliers, type mismatches, regex pattern rules, format drift), reusing profiling's sentinel coercion so both features agree on missingness
- severity-weighted 0-100 overall score plus per-column health scores; each issue maps to a plain-language suggested fix
- POST /projects/{id}/quality computes the report on demand; nothing is persisted
- Quality workspace feature: results tab plus docked config panel (outlier method/sensitivity, per-column pattern rules) bridged by QualityViewContext, with session-local score history and a stale banner offering a re-run after transforms
- opening the tab auto-runs a default-config assessment; fixable suggestions render an action button that docks the matching transform form (drop duplicates, fill empty, cast type, trim whitespace, replace values)

Fixes #434

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

45 new backend tests (`tests/test_quality.py`) cover every detector (including edge cases: sentinel nulls, constant columns, too-few values, invalid regex/unknown columns → 422), scoring bounds, remediation mapping, JSON safety, and all endpoint behaviors including auth. Manual verification against a live backend + Postgres with a deliberately messy CSV confirmed correct issues, sample row indices, and column scores end-to-end.

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

## Checklist

- [x] Tests pass locally


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/quality_service.py:237` detect_type_mismatches only flags a text column as a date type-mismatch when `TYPE_PARSE_THRESHOLD <= date_rate < 1.0` — a column where every non-null value parses as a date (date_rate == 1.0) is never flagged. This is inconsistent with the numeric branch a few lines above, which explicitly handles `numeric_rate == 1.0` as a castable MEDIUM issue, and contradicts the function's own docstring ('Datetime gets the same mostly-parseable check'). A pure date-as-text column — a very common CSV import pattern — is silently never suggested for casting. No test covers the date_rate == 1.0 boundary; test_mostly_dates_flags_offenders only exercises the 90% case, mirroring the gap.
- **[major/security]** `dataloom-backend/app/services/quality_service.py:276` validate_patterns compiles an arbitrary user-supplied regex (schemas.py allows up to 500 chars, 50 rules/request) and runs `.fullmatch` per row via `series.map(...)`, synchronously, inside the `async def run_quality_assessment` endpoint — no timeout, no complexity check. A catastrophic-backtracking pattern (classic `(a+)+`-style forms) matched against one non-matching value would hang; since this executes on the asyncio event loop rather than being offloaded to a thread/process, it stalls the whole worker for every in-flight request, not just the caller's.
- **[minor/bug]** `dataloom-frontend/src/Components/quality/QualityReportView.tsx:17` OPERATION_ACTION only maps 5 operation names (dropDuplicate, fillEmpty, castDataType, trimWhitespace, stringReplace), but quality_service.py's suggest_remediations (line 371) emits a 6th operation, "delCol", for high-severity missing_values ("mostly empty, consider dropping it"). Since `OPERATION_ACTION['delCol']` is undefined, that remediation's action button silently never renders — the user sees the suggestion text but gets no one-click way to act on it, unlike every other remediation type.
- **[minor/bug]** `dataloom-backend/app/services/quality_service.py:382` suggest_remediations tells apart the three inconsistent_format sub-cases (whitespace / casing / mixed-date-format) by substring-matching the free-text `issue['detail']` message ('whitespace' in detail, then 'casing' in detail) instead of a structured discriminator. Because detail embeds the raw column name, a casing-consistency issue on a column named e.g. `whitespace_count` hits the whitespace branch first and gets the wrong suggestion text plus the wrong operation (trimWhitespace instead of stringReplace).
- **[minor/quality]** `dataloom-frontend/src/context/QualityViewContext.tsx:51` `const { projectId, dataVersion } = useProjectContext() as unknown as {...}` (same pattern in QualityConfigPanel.tsx:52 for `columns`) double-casts through `unknown` rather than relying on ProjectContext's real type. If ProjectContext doesn't actually expose `dataVersion`, it silently reads `undefined` at runtime with no compiler warning, and `stale` (`dataVersion !== versionAtRun`) would evaluate true immediately after the very first run (`undefined !== 0`), permanently showing the 're-run needed' banner. Worth confirming ProjectContext's actual shape before merge.
- **[nit/style]** `dataloom-frontend/src/Components/Table.tsx:662` Fixes indentation on an unrelated `<div>` inside `TablePagination` — harmless, but unconnected to the quality-assessment feature this PR is titled for. Scope creep that makes the diff slightly harder to review/bisect.

**Conventions**

Solid, well-tested implementation: six pure detectors (duplicates, missing values, outliers, type mismatches, pattern rules, format drift) feed a severity-weighted 0-100 score and a remediation layer wired into the existing transform UI, backed by dozens of new backend tests covering detectors, scoring, and endpoint behavior. Two correctness gaps are worth a follow-up: a text column that's 100% parseable as dates is never flagged for casting (unlike the equivalent fully-numeric case), and the "delCol" remediation for mostly-empty columns has no matching entry in the frontend's action-button map, so that specific suggestion never gets a fix button. There's also an unbounded user-supplied-regex concern worth hardening, since patterns run synchronously against every row inside an async endpoint with no complexity or timeout guard. The rest is minor cleanup — a fragile substring-matched remediation branch, a couple of `as unknown as` casts, and a stray formatting fix in an unrelated file.

PR #446, opened 2026-07-17
fix: improve numeric cell editing consistency

### LINKED ISSUE #445: [Bug] Improve numeric cell editing consistency across column types
**Describe the bug**

Editing numeric cells currently results in inconsistent behavior depending on the input.

- Entering a decimal value (e.g. `4.5`) into an integer column does not consistently preserve the user's input.
- Entering an invalid numeric value (e.g. `1200 05`) into a float column causes the entire column to be converted to `string`, changing the displayed values (e.g. `1200` becomes `1200.0`) while preserving the invalid input.

This results in inconsistent typing semantics and unexpected column type changes.

**To Reproduce**

1. Upload a dataset containing an integer or float column.
2. Edit an integer cell and enter a decimal value such as `4.5`.
3. Observe the resulting value and column type.
4. Edit a float cell and enter an invalid numeric value such as `1200 05`.
5. Observe that the column is converted to a string column.

**Expected behavior**

- Editing an integer column with a decimal value should preserve the entered value by promoting the column to `float`.
- Editing a float column with an invalid numeric value should fail validation instead of changing the column type.
- Invalid edits should not modify the dataset or create transformation logs.

**Screenshots**

N/A

**Environment:**

- OS: macOS
- Browser: Chrome
- Node.js version: Project-supported version
- Python version: Project-supported version

**Additional context**

This improves consistency with the application's typed-column model and prevents accidental schema changes caused by invalid user input.

SHIVAM-KUMAR-59:
## Description

Improves numeric cell editing behavior by making type handling more consistent and predictable.

### Changes

- Integer columns are automatically promoted to `float` when a decimal value is entered, preserving the user's input instead of silently truncating it.
- Float columns now perform strict validation and reject invalid numeric values instead of silently converting the entire column to `object`.
- Invalid numeric edits no longer mutate the dataset or unintentionally change the column's dtype.
- Existing boolean and datetime validation behavior remains unchanged.

This prevents unexpected schema changes while preserving valid user edits.

Fixes #445 

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Manual testing included:

- Editing an integer cell with another integer.
- Editing an integer cell with a decimal value and verifying the column is promoted to `float`.
- Editing a float cell with a valid decimal value.
- Editing a float cell with an invalid numeric value and verifying validation is shown.
- Verifying invalid edits do not modify the dataset.
- Verifying boolean and datetime editing behavior remains unchanged.
- Running the existing transformation service test suite.

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

N/A

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -8,6 +8,7 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 
+import numpy as np
 import pandas as pd
 
 from app.schemas import FillStrategy, MeltParams, OperationType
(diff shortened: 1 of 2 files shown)


SHIVAM-KUMAR-59:

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:283` When a decimal value is entered into an integer cell, `df[column_name] = df[column_name].astype(float)` upcasts the ENTIRE column to float64, not just the edited cell. Any other, unedited row in that column holding an integer beyond 2^53 (e.g. 9007199254740993 — the exact value this file's own test uses to verify precision preservation) silently loses precision when rounded to the nearest representable float64. This is new: the old code never rewrote the whole column on a successful edit, so other rows' values were never at risk.
- **[major/test]** `dataloom-backend/app/services/transformation_service.py:318` The `elif pd.api.types.is_datetime64_any_dtype(col_dtype):` branch (~20 lines, including a fragile `len(normalized) == 10` heuristic to pick between date-only and full-timestamp formats) does not exist on main — `change_cell_value` previously had no datetime handling at all, so unvalidated strings flowed straight into `df.at[...]` and relied on pandas' default coercion. The PR description's claim that 'existing boolean and datetime validation behavior remains unchanged' is therefore inaccurate for the datetime half, and this new branch has zero corresponding tests added in `TestChangeCellValue` (only the pre-existing, unrelated `test_cast_to_datetime` for a different function exists in the suite).
- **[minor/quality]** `dataloom-backend/app/services/transformation_service.py:300` In the float-column branch, `value` is reassigned in place (`value = float(normalized)`), so the subsequent finite-number check's error message `f"Cannot interpret {value!r} as a finite number..."` reports the already-parsed float (e.g. `inf` for input '1e500') rather than the user's original input. The int-column branch avoids this by using a separate `float_value` variable so its equivalent error message correctly echoes the raw user string — the two branches are inconsistent.
- **[nit/style]** `dataloom-frontend/src/Components/Table.tsx:662` Pure indentation fix on the pagination container div (2-space to 6-space), unrelated to numeric cell editing and bundled into a PR otherwise scoped to backend type-coercion logic. Harmless but adds unrelated noise to the diff.

**Conventions**
- **Description:** States datetime validation 'remains unchanged,' but change_cell_value has no datetime branch on main at all — it's entirely new, untested logic that the description mischaracterizes as pre-existing.
- _Single well-formed conventional commit; the only hygiene gap is the description inaccurately calling the new datetime-validation branch pre-existing/unchanged._

This PR correctly implements the two behaviors requested in the linked issue: decimal edits into integer columns now promote the column to float instead of silently truncating, and invalid edits into float columns raise instead of demoting the column to object — both paths have solid new test coverage. However, it also introduces an entirely new, untested datetime-validation branch that the description inaccurately calls "unchanged," and the int-to-float column promotion can silently lose precision on other, unedited rows holding integers above 2^53. There's also a minor inconsistency where the float branch's "finite number" error reports the already-parsed value instead of the user's original input, unlike the equivalent int-branch check. Recommend adding test coverage for the datetime path (or splitting it into a separate PR) and reconsidering the column-wide float promotion for large-magnitude integer columns.

SHIVAM-KUMAR-59:
Hi @ivantha ! I've addressed all of the major review comments on this PR.

**Changes made:**

- Fixed the precision loss issue when editing an integer cell with a decimal value by promoting the column to `object` instead of `float64`. This preserves untouched large integers (including values beyond `2^53`) while still allowing decimal edits.
- Added a regression test covering the large integer precision case to ensure this behavior is preserved going forward.
- Removed the unrelated datetime validation changes from this PR to keep the scope focused on numeric cell edit validation, as suggested by the review.

The PR should now be ready for another review. Thanks!

ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `dataloom-backend/app/services/transformation_service.py:302` In the float-column branch, `value` is reassigned to the parsed float on line 294 before the finiteness check. The rejection raised on lines 300-303 then interpolates that already-parsed value (e.g. `inf`, `nan`, `-inf`) instead of the user's original raw input (e.g. 'Infinity', 'NaN', '1e500'), so the error message misreports what was actually typed. The analogous integer-column path avoids this: it keeps the parsed result in a separate `decimal_value` variable, so `value` still holds the original input when its own finiteness-check message is built on line 279.
- **[nit/quality]** `dataloom-backend/app/services/transformation_service.py:288` The float-branch parse -> isfinite-check -> raise sequence (lines 288-303) duplicates the decimal-fallback parse -> isfinite-check -> raise sequence in the integer branch (lines 269-286) almost line for line, differing only in which variable holds the parsed result. That divergence is exactly what produced the value/decimal_value mismatch flagged above. Extracting a small shared helper (e.g. `_parse_finite_float(raw_value, normalized, column_name)`) used by both branches would remove the duplication and the risk of the two paths drifting again.
- **[minor/test]** `dataloom-backend/tests/test_transformations.py:402` `test_float_cell_rejects_non_finite_values` is parametrized over 'nan'/'NaN'/'inf'/'-inf'/'Infinity'/'-Infinity'/'1e500' but only asserts `match="finite number"`. It never checks that the raised message contains the original raw input, so it doesn't catch that the float-column path reports the parsed float (e.g. `inf`) rather than what the user actually typed.

**Conventions**

This PR meaningfully tightens change_cell_value's numeric validation: integer columns now promote to object+float on decimal input instead of silently truncating, and both integer and float columns reject non-finite values (nan/inf) and non-numeric strings with a TransformationError instead of silently demoting the column and storing the raw string; tests confirm invalid edits leave the original dataframe untouched. There is one real message-formatting bug: in the float-column branch, the 'finite number' rejection interpolates the already-parsed float (e.g. `inf`) rather than the user's original input (e.g. 'Infinity'), unlike the equivalent integer-column path, and the parametrized test for that case only checks for the substring 'finite number' so it didn't catch the drift. Recommend fixing the message to reference the pre-parse string and considering a small shared helper for the two nearly-identical parse/validate/raise blocks so they don't diverge again. Otherwise the change is well-tested and the commit/PR hygiene is clean.

PR #450, opened 2026-07-18
test: add unit test coverage for all transformation form components

### LINKED ISSUE #449: [Feature] Add unit test coverage for all transformation form components
The transformation form components under `Components/forms/` (`FillEmptyForm`, `FilterForm`, `MeltForm`, `SampleRowsForm`, `SortForm`, `AdvQueryFilterForm`, `DropDuplicateForm`, `GroupByForm`, `PivotTableForm`, `StringReplaceForm`, `TrimWhitespaceForm`, `CastDataTypeForm`) currently have little to no automated test coverage. These components each drive a distinct data-transformation operation, contain their own validation logic, error handling, and preview/save workflow, but regressions in any of them (e.g., a broken payload shape, a validation check silently no-op'ing, preview mode not disabling correctly) would only be caught through manual QA. This makes refactors and shared-dependency changes (`usePreviewSave`, `ProjectContext`, `ColumnSelect`, `Select`, `ColumnMultiSelect`) risky, since there's no automated signal when a change to a shared component breaks one of the eleven forms silently.

Add a Vitest + React Testing Library test suite for each form component, following the existing pattern. Each suite should cover:
- Default rendering of all form controls and buttons
- Default field values (e.g., default strategy/condition/aggregation function)
- Client-side validation errors when required fields are missing
- Correct `transformProject` payload shape on successful submit
- `enterPreviewMode` being called with the transformation response
- Loading/"Applying..." disabled state while the request is pending
- Backend error messages (with and without a `response.data.detail`) rendering correctly
- Preview mode behavior: Apply disabled, "Save Changes" button appears and triggers `handleSave`, "Saving..." disabled state
- Cancel behavior: `cancelPreview` when in preview mode vs. `onClose` otherwise

This gives consistent, predictable coverage across every form and a template new forms can follow going forward.

- *End-to-end/integration tests only*: Could rely on E2E tests (e.g., Playwright/Cypress) covering the transformation flows instead of unit tests. Rejected for now since unit tests are faster to run, easier to pinpoint failures in, and don't require a full backend/browser environment; E2E coverage can be a complement, not a replacement.
- *Snapshot testing*: Considered relying on snapshot tests for rendering. Rejected because snapshots don't verify behavioral correctness (payload shape, validation logic) and tend to be brittle/rubber-stamped during review.

**Additional context**
N/A

SHIVAM-KUMAR-59:
## Description

Adds unit test coverage for all remaining transformation form components (`FillEmptyForm`, `FilterForm`, `MeltForm`, `SampleRowsForm`, `SortForm`, `AdvQueryFilterForm`, `DropDuplicateForm`, `GroupByForm`, `PivotTableForm`, `StringReplaceForm`, `TrimWhitespaceForm`, `CastDataTypeForm`). Each new suite mocks `api`, `ProjectContext`, `usePreviewSave`, and the shared `ColumnSelect`/`Select`/`ColumnMultiSelect` primitives, while exercising the real `useError`, `FormErrorAlert`, and `Button` components.

Fixes #449 

## Type of Change

- [x] New feature

## How Has This Been Tested?

- [x] Existing tests pass
- [x] New tests added

## Screenshots 
Screenshot 2026-07-18 at 21 00 20

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/test/AdvQueryFilterForm.test.jsx
+++ b/dataloom-frontend/src/test/AdvQueryFilterForm.test.jsx
@@ -1,0 +1,189 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { ADV_QUERY_FILTER } from "../constants/operationTypes";
+import AdvQueryFilterForm from "../Components/forms/AdvQueryFilterForm";
+import { transformProject } from "../api";
+import { useProjectContext } from "../context/ProjectContext";
+import usePreviewSave from "../hooks/usePreviewSave";
+
+vi.mock("../api", () => ({
+  transformProject: vi.fn(),
+}));
+
+vi.mock("../context/ProjectContext", () => ({
+  useProjectContext: vi.fn(),
+}));
+
+vi.mock("../hooks/usePreviewSave", () => ({
+  default: vi.fn(),
+}));
+
+const mockEnterPreviewMode = vi.fn();
+const mockCancelPreview = vi.fn();
+const mockHandleSave = vi.fn();
+
+const renderForm = ({ isPreviewMode = false, onClose = vi.fn(), saving = false } = {}) => {
+  useProjectContext.mockReturnValue({
+    isPreviewMode,
+    enterPreviewMode: mockEnterPreviewMode,
+    cancelPreview: mockCancelPreview,
+  });
+
+  usePreviewSave.mockReturnValue({
+    saving,
+    handleSave: mockHandleSave,
+  });
+
+  return {
+    onClose,
+    ...render(<AdvQueryFilterForm projectId="project-123" onClose={onClose} />),
+  };
+};
+
+describe("AdvQueryFilterForm", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+
+    transformProject.mockResolvedValue({
+      columns: ["amount"],
+      rows: [["100"]],
+      dtypes: { amount: "integer" },
+    });
+  });
+
+  it("renders the query input and buttons", () => {
+    renderForm();
+
+    expect(screen.getByPlaceholderText("e.g., col1 > 10 and col2 < 5")).toBeInTheDocument();
+    expect(screen.getByRole("button", { name: "Submit" })).toBeInTheDocument();
+    expect(screen.getByRole("button", { name: "Cancel" })).toBeInTheDocument();
+  });
+
+  it("requires a query to be entered", () => {
+    renderForm();
+
+    expect(screen.getByPlaceholderText("e.g., col1 > 10 and col2 < 5")).toBeRequired();
+  });
+
+  it("submits the query for preview", async () => {
+    const user = userEvent.setup();
+    renderForm();
+
+    await user.type(screen.getByPlaceholderText("e.g., col1 > 10 and col2 < 5"), "amount > 10");
+    await user.click(screen.getByRole("button", { name: "Submit" }));
+
+    await waitFor(() => {
+      expect(transformProject).toHaveBeenCalledWith(
+        "project-123",
+        {
+          operation_type: ADV_QUERY_FILTER,
+          adv_query: { query: "amount > 10" },
+        },
+        { preview: true },
+      );
+    });
+  });
+
+  it("enters preview mode using the transformation response", async () => {
+    const user = userEvent.setup();
+    const response = { columns: ["amount"], rows: [[100]], dtypes: { amount: "integer" } };
+    transformProject.mockResolvedValue(response);
+
+    renderForm();
+    await user.type(screen.getByPlaceholderText("e.g., col1 > 10 and col2 < 5"), "amount > 10");
+    await user.click(screen.getByRole("button", { name: "Submit" }));
+
+    await waitFor(() => {
+      expect(mockEnterPreviewMode).toHaveBeenCalledWith(
+        response.columns,
+        response.rows,
+        response.dtypes,
+        expect.objectContaining({ projectId: "project-123" }),
+      );
+    });
+  });
+
+  it("disables Submit while the request is pending", async () => {
+    const user = userEvent.setup();
+    let resolveTransform;
+    transformProject.mockImplementation(
+      () =>
+        new Promise((resolve) => {
+          resolveTransform = resolve;
+        }),
+    );
+
+    renderForm();
+    await user.type(screen.getByPlaceholderText("e.g., col1 > 10 and col2 < 5"), "amount > 10");
+    await user.click(screen.getByRole("button", { name: "Submit" }));
+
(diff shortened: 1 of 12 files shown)


SHIVAM-KUMAR-59:

ivantha:
### Automated review summary

**Triage:** test  |  **Recommendation:** needs-revision

**Findings**
- **[major/test]** `dataloom-frontend/src/test/MeltForm.test.jsx:91` In the 'shows an error if fetching columns fails' test, `render(<MeltForm .../>)` is called on line 91 before `useProjectContext.mockReturnValue(...)` is configured (lines 92-96 that follow it). Since `beforeEach` only calls `vi.clearAllMocks()` (which clears call history but not a mock's `mockReturnValue` implementation) and this test bypasses the file's `renderForm()` helper that normally sets the mock first, the initial render here only works by accident of whatever `useProjectContext.mockReturnValue` was left over from the previous test. Run in isolation, reordered, or with the prior test removed, `useProjectContext()` would return undefined on mount and MeltForm would likely throw when destructuring it, failing before the intended 'load failure' assertion is ever reached.
- **[minor/test]** `dataloom-frontend/src/test/PivotTableForm.test.jsx:40` The `ColumnSelect` mock disambiguates the 'Column' vs. 'Value' selects using a module-level counter (`columnSelectRenderCount`) that increments on every render and is never reset in `beforeEach`, so it persists across all `it()` blocks in the file. Correctness of `getColumnSelects()` in every test relies on the unstated, unverified invariant that `PivotTableForm` always renders exactly two `ColumnSelect` instances together, in the same order, on every render pass; if any future test causes an odd number of `ColumnSelect` renders, the parity flips and later tests would silently read/write the wrong select (Column vs Value) with no assertion catching the mix-up. A `data-testid` keyed off the distinguishing `placeholder` prop already passed to the mock would be far more robust than order-based tagging.
- **[nit/other]** `dataloom-frontend/src/Components/Table.tsx:662` This one-line JSX indentation fix inside `TablePagination` is unrelated to the PR's stated scope of adding test coverage for transformation forms, and no test in this diff touches Table.tsx. It's harmless (pure whitespace, no behavioral effect) but is scope creep in an otherwise test-only change; incidental production-code formatting fixes are cleaner split into their own commit.

**Conventions**
- _Hygiene is otherwise clean; the only nit is that the single commit also carries an unrelated Table.tsx formatting fix that would be cleaner split out on its own._

This PR adds 12 new Vitest suites covering previously-untested transformation form components (FillEmptyForm, FilterForm, MeltForm, SampleRowsForm, SortForm, AdvQueryFilterForm, DropDuplicateForm, GroupByForm, PivotTableForm, StringReplaceForm, TrimWhitespaceForm, CastDataTypeForm), plus an incidental one-line JSX indentation fix in Table.tsx. The suites are thorough and consistently structured (shared render helpers, mocked api/context/hooks, preview-mode and error-path coverage), but MeltForm.test.jsx has one test that renders the component before its `useProjectContext` mock is configured, so it currently passes only by accident of test-file execution order — that ordering should be fixed. PivotTableForm.test.jsx also disambiguates its two ColumnSelect mocks via an unreset, shared render-order counter, which is fragile and worth replacing with an explicit signal before the pattern gets copied elsewhere. The unrelated Table.tsx formatting change should ideally be split out of this test-only commit.

SHIVAM-KUMAR-59:
Hi @ivantha ! I've addressed the major review comments on this PR.

**Changes made:**

- Updated the test setup so` useProjectContext` and `usePreviewSave` are mocked before rendering the component, removing the dependency on mock state from previous tests.
- Moved the default mock configuration into `beforeEach`, ensuring every test starts with a consistent, isolated setup and can run independently regardless of execution order.

The tests are now deterministic and the PR should be ready for another review. Thanks!

ivantha:
### Automated review summary

**Triage:** test  |  **Recommendation:** needs-revision

**Findings**
- **[major/test]** `dataloom-frontend/src/test/GroupByForm.test.jsx:98` The rewritten GroupByForm.test.jsx (replacing the pre-existing file, not additive) drops three tests that covered real unmount/cancellation logic in GroupByForm.jsx: 'does not call enterPreviewMode when user closes during loading' (previously main:dataloom-frontend/src/test/GroupByForm.test.jsx:94, exercising the cancelledRef guard at GroupByForm.jsx:60), 'allows re-submit after a cancelled in-flight request' (previously line 125), and 'calls cancelPreview on unmount when in preview mode' (previously line 189, exercising the cleanup effect at GroupByForm.jsx:34-41 that calls cancelPreviewRef.current()). None of the 12 new tests in this file unmount the component, so this real cleanup/race-condition logic is now untested even though the PR's stated purpose is adding coverage.
- **[minor/quality]** `dataloom-frontend/src/test/PivotTableForm.test.jsx:40` PivotTableForm.test.jsx disambiguates the two unlabeled ColumnSelect mocks (Column vs Value) using a module-level `columnSelectRenderCount` counter that is never reset in beforeEach/vi.clearAllMocks, alternating data-testid by parity across the whole file. It happens to work today only because PivotTableForm.jsx (Components/forms/PivotTableForm.jsx:72-84) always renders exactly two ColumnSelect instances, unconditionally, in a fixed order; any future change to that render structure (conditional rendering, reordering, an added instance) would silently point getColumnSelects() at the wrong field instead of failing clearly. Passing an explicit data-testid from PivotTableForm.jsx (as FilterForm.jsx and GroupByForm.jsx already do for their ColumnSelect instances) would be more robust than this counter.
- **[minor/test]** `dataloom-frontend/src/test/AdvQueryFilterForm.test.jsx:69` Running the full new/modified suite (154 tests, all passing) surfaces about 370 React 'An update ... was not wrapped in act(...)' warnings, concentrated in tests that click a submit button and then only waitFor the initial API-call assertion (e.g. 'submits the query for preview') without also waiting for the subsequent state settle (enterPreviewMode/setLoading(false) in the submit handler's then/finally). It's console noise rather than a functional failure, but it's pervasive across effectively all 12 files and indicates the async interactions aren't being fully awaited.

**Conventions**

This PR adds Vitest coverage for all twelve transformation form components, consistently mocking the shared ColumnSelect/ColumnMultiSelect/Select primitives while exercising the real error-handling and preview-save hooks; all 154 tests pass and the changed files are clean under ESLint and Prettier (independently verified by checking out the PR branch and running both). The rewritten GroupByForm.test.jsx replaces the previous test file rather than extending it, and in doing so drops three tests that covered real unmount/cancellation logic in GroupByForm.jsx (the cancelledRef guard and the cancelPreview-on-unmount cleanup effect) without an equivalent replacement - that coverage should be restored before merge. Lower-priority follow-ups: PivotTableForm.test.jsx leans on a fragile, never-reset module-level counter to tell its two ColumnSelect mocks apart (works today, but only by an implicit invariant), and most new tests leave a large number of harmless-but-noisy "not wrapped in act(...)" warnings from not fully awaiting async state settling.

SHIVAM-KUMAR-59:
Hi @ivantha ! I've addressed the review comments on this PR.

**Changes made:**

- Added the missed test cases in `GroupByForm.test.jsx`.

The PR should now be ready for another review. Thank you!

PR #455, opened 2026-07-21
fix: run lint on pull_request and comment via workflow_run

### LINKED ISSUE #454: fix: Lint workflow fails for fork PRs due to actions/checkout@v4 pull_request_target restriction
## Bug Description

The `Lint / Frontend Lint` and `Lint / Backend Lint` required checks fail for **all PRs from forks** with the following error:

```
##[error] Refusing to check out fork pull request code from a 'pull_request_target' workflow.
This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access.
Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities.
To opt in, review the risks at https://gh.io/securely-using-pull_request_target
and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.
```

## Root Cause

The `lint.yml` workflow uses the `pull_request_target` trigger and `actions/checkout@v4` to check out the PR head code. Recent versions of `actions/checkout@v4` now **block fork checkouts by default** in `pull_request_target` workflows as a security measure.

Since `pull_request_target` always reads the workflow file from `main`, contributors **cannot fix this from their PRs** — the fix must be applied on the `main` branch.

## Suggested Fix

Add `allow-unsafe-pr-checkout: true` to both checkout steps in `.github/workflows/lint.yml`:

```yaml
# Frontend Lint job (line ~34)
- uses: actions/checkout@v4
  with:
    repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
    ref: ${{ github.event.pull_request.head.sha || '' }}
    allow-unsafe-pr-checkout: true

# Backend Lint job (line ~74)
- uses: actions/checkout@v4
  with:
    repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
    ref: ${{ github.event.pull_request.head.sha || '' }}
    allow-unsafe-pr-checkout: true
```

This is safe in this context because the lint workflow only runs read-only linting tools (ESLint, Prettier, tsc, Ruff) — it does not publish artifacts, deploy, or use secrets beyond the default token.

## Impact

This blocks **all fork PRs** from passing the 2 required Lint checks, making them unmergeable even when the code is fully clean.

YADUNANDAN-SINGH:
## Summary

`actions/checkout@v4` now refuses fork checkouts under `pull_request_target`, which blocks the `Lint / Frontend Lint` and `Lint / Backend Lint` required checks on **every fork PR**.

My first attempt at this fix set `allow-unsafe-pr-checkout: true`. That was the wrong call and I've replaced it — details below.

## Why overriding the guard was unsafe

The lint jobs don't only run read-only tools. They execute PR-authored code:

- `npm ci` runs `postinstall` / `prepare` lifecycle scripts from the PR's `package.json` and lockfile
- `npx eslint` loads `eslint.config.js` **from the PR** — contributor-authored JS executed in-process
- `uv sync --frozen` can invoke build backends declared in the PR's `pyproject.toml`

Under `pull_request_target` that code ran with `permissions: pull-requests: write` and default-branch cache scope. That is exactly the "pwn request" pattern the checkout guard exists to prevent, so the guard was right and my override would have re-opened the hole.

## The change

Split the pipeline along the trust boundary instead:

- **`lint.yml`** → `pull_request` trigger, `permissions: contents: read`. Fork code still runs, but with a read-only token and no secrets — the same model `ci.yml` and `e2e.yml` already use in this repo. Results are published as a data-only `report.json` artifact.
- **`lint-comment.yml`** (new) → `workflow_run` trigger. Holds the `pull-requests: write` that `lint.yml` gave up, and never checks out PR code. The comment logic is a faithful port: same `<!-- ci-bot -->` marker, same legacy-marker cleanup, same update-in-place and delete-when-clean behaviour.

Two hardenings beyond a straight port, since the artifact originates from a run that executed fork code:

- verifies the reported PR head SHA matches the triggering run's head SHA, so a forged `report.json` cannot make the bot comment on an unrelated PR
- strips backticks and caps the merge-commit list at 20 lines before embedding author-controlled commit subjects in the comment

Workflow name and both job names are unchanged.

## Verification

The new `pull_request` lint ran green on this PR — which is itself a fork PR, so it exercises the exact path that was broken:

- Frontend Lint ✅ · Backend Lint ✅ · Rebase Check ✅ · Report ✅
- https://github.com/c2siorg/dataloom/actions/runs/29833169069

To be upfront about coverage: **`lint-comment.yml` has not executed and cannot until it is on `main`**, because `workflow_run` always loads its workflow file from the default branch. I've tested its logic locally (clean/failure/rebase/squash paths, comment update vs. duplicate, forged-PR-number rejection, markdown-fence sanitisation), but it is unproven in live CI until merge.

## What this needs from a maintainer

1. **This PR requires an admin merge.** The `Lint / ... (pull_request_target)` checks here are produced by the current `main` workflow — the one this PR removes — so they cannot pass on this PR by construction.
2. **Branch protection needs updating at merge time.** Once this lands, `pull_request_target` lint checks will never be produced again. If they remain in the required-contexts list, every future PR will block permanently. The replacements are the `pull_request`-triggered `Frontend Lint` / `Backend Lint`.

Happy to adjust the approach if you'd prefer a different split.

Closes #454

### DIFF
--- a/.github/workflows/lint-comment.yml
+++ b/.github/workflows/lint-comment.yml
@@ -1,0 +1,159 @@
+name: Lint Comment
+
+# Privileged half of the lint pipeline. It is triggered by the completion of
+# Lint (lint.yml) and never checks out or executes pull request code — it only
+# reads the data-only report.json artifact that run produced.
+
+on:
+  workflow_run:
+    workflows: [Lint]
+    types: [completed]
+
+concurrency:
+  group: lint-comment-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  actions: read
+  pull-requests: write
+
+jobs:
+  comment:
+    name: PR Comment
+    if: github.event.workflow_run.event == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download report
+        id: download
+        continue-on-error: true
+        uses: actions/download-artifact@v4
+        with:
+          name: pr-lint-report
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Post or delete combined comment
+        if: steps.download.outcome == 'success'
+        uses: actions/github-script@v7
+        env:
+          HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+        with:
+          script: |
+            const fs = require('fs');
+            const report = JSON.parse(fs.readFileSync('report.json', 'utf8'));
+
+            // report.json comes from a run that executed fork code, so treat
+            // every field as untrusted input.
+            const prNumber = parseInt(report.pr, 10);
+            if (!Number.isInteger(prNumber) || prNumber <= 0) {
+              core.setFailed('Report contains no usable PR number.');
+              return;
+            }
+
+            // Confirm the reported PR really is the one that triggered the run,
+            // so a forged report cannot make the bot comment on someone else's PR.
+            const { data: pr } = await github.rest.pulls.get({
+              owner: context.repo.owner, repo: context.repo.repo,
+              pull_number: prNumber,
+            });
+            if (pr.head.sha !== process.env.HEAD_SHA) {
+              core.setFailed(`PR #${prNumber} head ${pr.head.sha} does not match run head ${process.env.HEAD_SHA}.`);
+              return;
+            }
+
+            const marker = '<!-- ci-bot -->';
+            const oldMarkers = ['<!-- lint-frontend-bot -->', '<!-- lint-backend-bot -->', '<!-- rebase-bot -->'];
+
+            const { data: comments } = await github.rest.issues.listComments({
+              owner: context.repo.owner, repo: context.repo.repo,
+              issue_number: prNumber,
+            });
+
+            // Clean up old individual comments from previous workflow versions
+            for (const old of oldMarkers) {
+              const c = comments.find(c => c.body.includes(old));
+              if (c) {
+                await github.rest.issues.deleteComment({
+                  owner: context.repo.owner, repo: context.repo.repo,
+                  comment_id: c.id,
+                });
+              }
+            }
+
+            // Build combined comment sections
+            const sections = [];
+
+            const eslintFailed = report.eslint === 'failure';
+            const prettierFailed = report.prettier === 'failure';
+            const typecheckFailed = report.typecheck === 'failure';
+            if (eslintFailed || prettierFailed || typecheckFailed) {
+              let s = '### Frontend\n\n';
+              if (eslintFailed) s += '- **ESLint:** Run `npx eslint --fix . --ext js,jsx,ts,tsx` in `dataloom-frontend/`\n';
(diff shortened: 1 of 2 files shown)


YADUNANDAN-SINGH:
Hi @ivantha — following up on my own PR to correct it.

I took a harder look at my `allow-unsafe-pr-checkout: true` suggestion and it was wrong. I'd claimed the lint jobs were safe because they only run read-only tools, but that isn't accurate: `npm ci` executes `postinstall`/`prepare` scripts from the PR's `package.json`, and `npx eslint` loads `eslint.config.js` from the PR. Under `pull_request_target` both run against a `pull-requests: write` token. GitHub's guard was correct and my patch would have re-opened a pwn-request hole.

I've force-pushed a replacement: lint jobs move to `pull_request` (read-only token, no secrets — same as the existing `ci.yml`/`e2e.yml`), and the PR commenting moves to a `workflow_run` workflow that holds the write permission but never checks out PR code. I've updated the PR description with the full rationale.

The new lint ran green on this PR itself: https://github.com/c2siorg/dataloom/actions/runs/29833169069

Two things I can't do from here:

1. The `Lint / ... (pull_request_target)` required checks on this PR are produced by the current `main` workflow — the one this PR removes — so they can't pass here by construction. This will need an admin merge.
2. After it merges, `pull_request_target` lint checks stop being produced altogether. If they're still listed in branch protection's required contexts, every PR will block permanently. They'd need swapping for the `pull_request`-triggered `Frontend Lint` / `Backend Lint`.

Also worth flagging honestly: `lint-comment.yml` can't run until it's on `main` (`workflow_run` always reads the workflow from the default branch), so that half is tested locally but unproven in CI. If it misbehaves after merge I'll fix it straight away.

Happy to restructure differently if you'd prefer another approach.

ivantha:
### Automated review summary

**Triage:** ci  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[minor/quality]** `.github/workflows/lint-comment.yml:29` The Download report step (continue-on-error: true, line 29) is gated only by the binary steps.download.outcome == 'success' check (line 37), so a genuine download error (auth/network) is indistinguishable from 'no artifact was produced because Lint failed early.' Both cases silently skip the comment step with no failed check and no logged warning, which can also leave a stale prior PR comment on screen uncorrected instead of being updated or removed.

**Conventions**

This PR replaces the pull_request_target lint trigger with a pull_request plus workflow_run split so PR checks no longer execute fork code under elevated write permissions, fixing the fork-PR checkout failures from #454. The design matches GitHub's standard mitigation for this class of issue: lint.yml now runs fork code with a read-only token and hands results off via a data-only report.json artifact, while the new lint-comment.yml holds the pull-requests:write permission and never checks out PR code, additionally verifying the reported PR's head SHA against the triggering run before posting to guard against a forged report. The pull_request-triggered path is confirmed passing in live CI on this PR itself; the workflow_run comment path cannot run until merged (a GitHub platform constraint, not an oversight) and was only exercised locally, so it deserves a close look on the first real merge. One minor robustness gap worth a follow-up: the artifact-download step treats all failure modes the same way, so a genuine download error produces no comment and no visible failure signal rather than a distinguishable one.

PR #463, opened 2026-07-25
feat: add formula columns via sandboxed addFormulaCol transformation

hanzalahwaheed:
- addFormulaCol OperationType + FormulaColumnParams, registered in TRANSFORMATION_REGISTRY so apply, preview, and save-replay share one path
- validate_formula_expression() AST allowlist in security.py: arithmetic, comparisons, boolean ops, column refs, and literals only; rejects calls, attribute access, subscripts, and dunders
- add_formula_column() evaluates via DataFrame.eval with backtick-quoted non-identifier columns; never raw eval()
- FormulaColumnForm docked panel with column insert chips, wired into the Data ribbon and the shared preview flow
- unit, endpoint, replay, and component tests

## Description

Adds computed "formula" columns to DataLoom. Users open the docked Formula Column panel from the Data ribbon, name a new column, and enter an arithmetic/logical expression over existing columns (with insert-column chips for convenience). The expression is validated against an AST allowlist and evaluated with `DataFrame.eval` — never raw `eval()` — and flows through the same preview/save-replay path as every other transformation.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Backend unit tests cover the AST allowlist (accepting arithmetic/comparison/boolean forms and rejecting calls, attribute access, subscripts, and dunders), `add_formula_column` evaluation including backtick-quoted non-identifier columns, the transform endpoint, and save-replay. Frontend component tests cover the `FormulaColumnForm` panel. `npm run lint` (ESLint + `tsc --noEmit`) passes clean.

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-backend/app/schemas.py
+++ b/dataloom-backend/app/schemas.py
@@ -45,6 +45,7 @@ class OperationType(StrEnum):
     sample = "sample"
     stringReplace = "stringReplace"
     addFile = "addFile"
+    addFormulaCol = "addFormulaCol"
 
 
 class DropDup(StrEnum):
@@ -211,6 +212,20 @@ class DropNaParams(BaseModel):
         return v
 
 
+class FormulaColumnParams(BaseModel):
+    """Parameters for adding a computed column from a formula expression."""
+
+    column_name: str = Field(min_length=1, max_length=255)
+    expression: str = Field(min_length=1, max_length=500)
+
+    @field_validator("column_name", "expression")
+    @classmethod
+    def must_not_be_blank(cls, v: str) -> str:
+        if not v or not v.strip():
+            raise ValueError("must not be empty")
+        return v.strip()
+
+
 class StringReplaceParams(BaseModel):
     """Parameters for find-and-replace on a string column."""
 
@@ -333,6 +348,7 @@ class TransformationInput(BaseModel):
     melt_params: MeltParams | None = None
     sample_params: SampleParams | None = None
     string_replace_params: StringReplaceParams | None = None
+    formula_col_params: FormulaColumnParams | None = None
 
 
 class BasicQueryResponse(BaseModel):
--- a/dataloom-backend/app/services/transformation_service.py
+++ b/dataloom-backend/app/services/transformation_service.py
@@ -15,7 +15,7 @@ from app.schemas import FillStrategy, MeltParams, OperationType
 from app.services.append_service import append_dataframes
 from app.utils.logging import get_logger
 from app.utils.pandas_helpers import read_table_safe
-from app.utils.security import validate_query_string
+from app.utils.security import prepare_formula_expression, validate_query_string
 
 logger = get_logger(__name__)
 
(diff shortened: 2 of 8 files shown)


ivantha:
### Automated review summary

**Findings**
- **[blocking/security]** `dataloom-backend/app/services/transformation_service.py:875` df.eval() executes `expr`, but validate_formula_expression() (security.py) only AST-checks a differently-derived string `sanitized` (non-identifier columns replaced by placeholders like `_dataloom_col_0`, security.py:165-171) — the two strings are never reconciled before execution. Because quote normalization runs first (line 868: `expression.replace("'", '"')`), a column name containing an apostrophe (e.g. "buyer's price") no longer matches literally inside `expr`, so the backtick-wrap loop (869-871) silently fails to wrap it and its raw, never-AST-validated text reaches eval() directly instead of the placeholder form the validator actually approved. The AST allowlist's safety guarantee does not hold for the string that is actually executed.
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:869` This loop iterates df.columns in natural order and does a plain substring `.replace()`, unlike security.py's validator which explicitly sorts non-identifier columns longest-first (security.py:166) to avoid collisions. If one non-identifier column name is a substring of another (e.g. columns "amount due" and "total amount due") or of a string literal in the formula, the shorter name's `.replace()` corrupts the longer name/literal mid-token wherever the substring appears — even when the user never referenced the shorter column — producing malformed syntax that fails with a confusing "Formula could not be evaluated" error instead of working.
- **[major/bug]** `dataloom-frontend/src/Components/forms/FormulaColumnForm.tsx:43` enterPreviewMode() is called with only 4 arguments, omitting the 5th `paginationInfo` argument every sibling form passes (e.g. FillEmptyForm.tsx forwards `{ total_rows: response.total_rows, total_pages: response.total_pages, page: response.page, page_size: response.page_size }`). ProjectContext.jsx defaults the missing argument to `{}`, which resets `totalRows` to just `previewRows.length` and `totalPages` to `1` regardless of the dataset's real size, even though the backend response already includes the correct totals. On any project larger than the default 50-row page, previewing a formula column will make the pagination controls report far fewer rows/pages than actually exist.
- **[minor/quality]** `dataloom-backend/app/schemas.py:218` column_name and expression have no max_length, unlike comparable string fields elsewhere in this file (e.g. PatternRule.pattern uses Field(min_length=1, max_length=500); project name/description are capped at 255/1000 chars). An arbitrarily long expression is fully regex-scanned against 12 patterns, AST-parsed and walked, and substring-replaced once per non-identifier column before it ever reaches eval() — worth a length cap both for consistency and to bound the cost of a single request.
- **[minor/bug]** `dataloom-backend/app/services/transformation_service.py:870` A column literally named "True" or "False" passes col.isidentifier() (so it's never backtick-wrapped), but Python parses the bare tokens True/False as ast.Constant, not ast.Name — so a formula consisting of just that column's name validates fine and silently evaluates to the literal boolean broadcast to every row instead of the column's actual values, with no error at all. A column named "None" hits the opposite failure: referencing it is rejected with "Formula literals must be numbers, strings, or booleans", a confusing message for someone just trying to use their own column.

**Conventions**

This PR adds an addFormulaCol transformation that lets users create computed columns from arithmetic/boolean expressions, validated by an AST allowlist before running through DataFrame.eval. The security design has a real gap: the string that gets AST-validated (a placeholder-substituted copy) is not the string that actually reaches eval() (an independently-built, unsorted, backtick-wrapped copy), so a column name containing an apostrophe, or one that's a substring of another column name, can desync the two and let unvalidated text reach eval(), or silently corrupt legitimate formulas. Separately, the frontend's preview call drops the pagination metadata every sibling form forwards, so the preview table will misreport total rows/pages on datasets bigger than the default page size. Recommend reconciling the validated and executed strings (ideally by validating the exact string that will run) and restoring pagination forwarding before merge; the registry wiring, preview/save-replay integration, and test suite are otherwise solid.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/utils/security.py:194` prepare_formula_expression() substitutes non-identifier column names by doing a raw substring replace on the expression text (sanitized.replace(col, placeholder)) rather than operating on parsed tokens. If a string literal inside the formula happens to equal the exact text of such a column, that literal gets swapped for the column's placeholder too, and the reconstructed expression ends up silently comparing against a backtick-wrapped value instead of the string the user typed — no exception is raised. Verified directly: prepare_formula_expression('status == "unit cost"', ['status', 'unit cost']) returns 'status == "`unit cost`"', and evaluating that returns all-False on data where the correct comparison is [True, False, True]. This is especially reachable right after a Melt/unpivot, whose value/variable columns hold text equal to other columns' names, and no test in test_formula_columns.py exercises a literal colliding with a column name.
- **[nit/quality]** `dataloom-backend/app/services/transformation_service.py:854` The docstring's Raises section only documents TransformationError, but most of add_formula_column's actual failure paths (empty, dangerous-pattern, unparseable, or unknown-column formulas, all raised inside prepare_formula_expression) surface as a raw fastapi.HTTPException instead — only the duplicate/blank-name check and the eval-time catch raise TransformationError.

**Conventions**
- **Description:** Detailed on what changed and why, but never links an issue or tracker item it closes.
- _Single well-scoped conventional commit and a thorough narrative description; just add the issue link._

Adds a new addFormulaCol transformation for computed columns, backed by an AST-allowlisted DataFrame.eval (arithmetic/comparison/boolean only, no calls/attributes/subscripts) and wired through the existing preview/save-replay path, with good breadth of backend and frontend tests. The column-name placeholder substitution in prepare_formula_expression operates on raw substrings of the expression text rather than parsed tokens, so a string literal that happens to match a space-containing column's exact text is silently corrupted into a backtick-wrapped value instead of erroring out — verified directly with `status == "unit cost"` returning all-False instead of the correct result when "unit cost" is also a column name. That scenario is most reachable right after Melt/unpivot output, and it isn't covered by the new test suite. Recommend making the substitution collision-safe (or detecting and rejecting the collision) and adding a regression test before merging.

PR #465, opened 2026-07-25
feat: add profiling options dropdown to workspace tab bar '+' button

### LINKED ISSUE #464: [Enhancement] Add Profiling options dropdown to workspace tab '+' button
### Is your feature request related to a problem?
Clicking the `+` icon next to the "DataSet" tab does not perform any action or show available options for opening new profiling views.
### Describe the solution you'd like
When a user clicks the `+` icon in the workspace tab bar, a dropdown menu should appear listing all options from the **Profiling** menu:
- **Summary**: Opens the dataset summary tab.
- **Column Profiles**: Toggles column profiles on the table.
- **Charts**: Opens the charts tab and chart builder panel.
- **Quality**: Opens the data quality report tab and configuration panel.
### Additional Context
- The `+` button should stay positioned right next to the open tabs.
- Selecting any option should trigger the action and close the menu.

YADUNANDAN-SINGH:
## Description

Added a dropdown popover menu to the workspace tab bar `+` button directly adjacent to open tabs (`DataSet x [+]`). Clicking the `+` icon now displays all options from the **Profiling** menu:
- **Summary**: Opens the dataset summary tab.
- **Column Profiles**: Toggles inline column profiles on the table.
- **Charts**: Opens the charts tab and docks the chart builder panel.
- **Quality**: Opens the data quality report tab and docks the quality config panel.

Additionally, updated menu navbar tooltips to use clean theme-aware surface styling and fixed z-index popover layering over sticky table headers.

Fixes #464 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Ran unit tests, linter, formatting checks, and local manual verification.

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

## Screenshots 

image

## Checklist

- [x] Tests pass locally

### DIFF
--- a/dataloom-frontend/src/Components/DataScreen.jsx
+++ b/dataloom-frontend/src/Components/DataScreen.jsx
@@ -57,7 +57,7 @@ function WorkspaceContent({ projectId }) {
       <MenuNavbar projectId={projectId} />
       <div className="flex min-h-0 flex-1 overflow-hidden">
         <div className="flex min-h-0 flex-1 flex-col overflow-hidden">
-          <WorkspaceTabBar onAddTab={() => openTab(DATASET_TAB)} />
+          <WorkspaceTabBar />
           <div className="flex min-h-0 flex-1 flex-col overflow-hidden">{renderActiveTab()}</div>
         </div>
         <RightPanel projectId={projectId} />
--- a/dataloom-frontend/src/Components/MenuNavbar.jsx
+++ b/dataloom-frontend/src/Components/MenuNavbar.jsx
@@ -10,7 +10,7 @@ import { usePanel } from "../context/PanelContext";
 import { useWorkspaceTabs } from "../context/WorkspaceTabsContext";
 import { useHistoryRefresh } from "../context/HistoryRefreshContext";
 import { useColumnProfilesView } from "../context/ColumnProfilesContext";
-import { DATASET_TAB } from "./workspace/DataSetTab";
+import { useColumnProfilesAction } from "./workspace/useProfilingMenu";
 import { getFeatureMenu } from "./workspace/featureRegistry";
 
 // Ribbon skeleton: the top tabs and the group order within each. Features and the
@@ -32,7 +32,7 @@ const MenuNavbar = ({ projectId }) => {
   const { activePanel, openPanel, togglePanel, closePanel } = usePanel();
   const { openTab, activeTabId } = useWorkspaceTabs();
   const { refreshLogs, refreshCheckpoints } = useHistoryRefresh();
-  const { showColumnProfiles, toggleColumnProfiles } = useColumnProfilesView();
+  const { showColumnProfiles } = useColumnProfilesView();
 
   const handleMouseEnter = (e, hoverText) => {
     if (!hoverText) return;
@@ -54,12 +54,8 @@ const MenuNavbar = ({ projectId }) => {
     setActiveTooltip(null);
   };
 
-  // Column profiles render inside the DataSet tab, so focus it when turning
-  // them on (re-opens it if the tab was closed).
-  const handleToggleColumnProfiles = () => {
-    if (!showColumnProfiles) openTab(DATASET_TAB);
-    toggleColumnProfiles();
-  };
+  // Shared with the tab bar dropdown so both entry points behave identically.
+  const handleToggleColumnProfiles = useColumnProfilesAction();
 
   const handleSave = () => setIsInputOpen(true);
 
@@ -298,7 +294,7 @@ const MenuNavbar = ({ projectId }) => {
             top: `${activeTooltip.top}px`,
             left: `${activeTooltip.left}px`,
           }}
-          className="fixed -translate-x-1/2 z-40 pointer-events-none px-2.5 py-1 text-xs font-medium text-white bg-slate-900/90 dark:bg-slate-800/95 backdrop-blur-sm rounded-md shadow-md border border-slate-700/50 max-w-xs text-center text-wrap"
+          className="fixed -translate-x-1/2 z-50 pointer-events-none px-2.5 py-1 text-xs text-foreground bg-surface border border-app-border rounded-md shadow-md max-w-xs text-center text-wrap"
         >
           {activeTooltip.text}
         </div>
--- a/dataloom-frontend/src/Components/workspace/SummaryTab.tsx
+++ b/dataloom-frontend/src/Components/workspace/SummaryTab.tsx
@@ -1,8 +1,16 @@
 import { useParams } from "react-router-dom";
-import { useWorkspaceTabs } from "../../context/WorkspaceTabsContext";
+import { useWorkspaceTabs, type WorkspaceTab } from "../../context/WorkspaceTabsContext";
 import { useProjectContext } from "../../context/ProjectContext";
 import useDatasetSummary from "../../hooks/useDatasetSummary";
 import DatasetSummaryPanel from "../profiling/DatasetSummaryPanel";
+
+// eslint-disable-next-line react-refresh/only-export-components
+export const SUMMARY_TAB: WorkspaceTab = {
+  id: "summary",
+  title: "Summary",
+  type: "summary",
+  closeable: true,
+};
 
 /**
  * Summary tab — the dataset-wide overview (row/column counts, dtypes, memory).
--- a/dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx
+++ b/dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx
@@ -1,23 +1,110 @@
+import { useState, useEffect, useRef } from "react";
 import { Plus, X } from "lucide-react";
 import { useWorkspaceTabs } from "../../context/WorkspaceTabsContext";
-
-interface WorkspaceTabBarProps {
-  /** When provided, renders a trailing "+" button that invokes this. */
-  onAddTab?: () => void;
-}
+import { useProfilingMenuItems } from "./useProfilingMenu";
 
 /**
  * VS Code–style tab strip for the project workspace. Reads open tabs from
  * WorkspaceTabsContext; selecting a tab activates it and the × closes it.
+ * The "+" button directly follows open tabs and opens a dropdown menu of the
+ * Profiling actions, derived from the feature registry via useProfilingMenu.
  */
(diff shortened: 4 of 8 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx:69` Closing the closeable 'DataSet' tab while another tab (e.g. Summary, Charts, or Quality) stays active and showColumnProfiles is already true leaves no way to reopen the dataset table. The '+' button used to unconditionally call openTab(DATASET_TAB) (that handler was removed at DataScreen.jsx:60), and the new dropdown's only tab-opening path back to it, 'Column Profiles', is gated by `if (!showColumnProfiles)`, so it no-ops when column profiles are already on. The user is stuck until a full page reload resets tabs to initialTabs.
- **[minor/quality]** `dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx:14` onAddTab is now dead: WorkspaceTabBar has a single call site (DataScreen.jsx:60) and it no longer passes the prop, so the WorkspaceTabBarProps.onAddTab field and the `if (onAddTab) onAddTab();` branch in the '+' click handler (line 147) never fire. Worth removing, or documenting why it's being kept.
- **[minor/quality]** `dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx:150` The '+' button's aria-label="New tab" is now stale: the button no longer adds a tab directly, it opens a 'Profiling options' menu (matching the dropdown panel's own aria-label a few lines down). Screen reader users will hear a description that no longer matches what the control does.
- **[nit/test]** `dataloom-frontend/src/test/WorkspaceTabBar.test.tsx:63` The three new dropdown interaction tests (from 'opens profiling dropdown menu' onward) drive state updates (openTab / toggleColumnProfiles) via userEvent.click without them settling inside act(), so running the suite prints repeated 'not wrapped in act(...)' warnings for WorkspaceTabsProvider, PanelProvider, and ColumnProfilesProvider. Tests still pass; this is just noise worth cleaning up.

**Conventions**

This PR adds a dropdown to the workspace tab bar's '+' button that surfaces the existing Summary, Column Profiles, Charts, and Quality actions, plus small theme and z-index fixes to the MenuNavbar tooltip. It is covered by 8 passing tests and the diff is clean under tsc and eslint. However, removing the previous unconditional 'reopen DataSet tab' behavior of '+' creates a dead end: if a user closes the DataSet tab while Column Profiles is already toggled on and another tab is active, none of the four dropdown options (nor anything else short of a full page reload) gets them back to the data table, so this should be fixed before merge (for example, by adding an explicit 'Table' entry or making 'Column Profiles' always focus the dataset tab). It is also worth a quick cleanup pass on the now-unused onAddTab prop and the stale aria-label="New tab" on the '+' button.

YADUNANDAN-SINGH:
Thanks for the review @ivantha  all three code findings are fixed, and the branch is
rebased so the merge commit is gone.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/DataScreen.jsx:60` Removing the old `onAddTab={() => openTab(DATASET_TAB)}` callback leaves no direct way to reopen the closeable DataSet tab from the tab bar once a user closes it while another tab (Summary, Charts, etc.) stays active, since the empty-workspace fallback screen only appears when zero tabs remain. None of the four new dropdown options is "open DataSet"; the closest substitute, Column Profiles, only opens it as an undocumented side effect of switching that toggle on (and only on the click where the toggle goes from off to on).
- **[major/quality]** `dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx:55` The new `profilingOptions` array hand-encodes the same four actions that already exist once elsewhere: Summary, Charts, and Quality are declared via registerFeature and consumed generically by MenuNavbar through getFeatureMenu(), and Column Profiles already has its own toggle handler (handleToggleColumnProfiles) in MenuNavbar.jsx with the identical conditional. There are now two hand-maintained lists that must stay in sync for the same four actions, which is exactly the fan-out the featureRegistry module's own docstring says it exists to avoid.
- **[minor/test]** `dataloom-frontend/src/test/WorkspaceTabBar.test.tsx:88` The new dropdown tests cover Summary, Charts, and Quality (each a one-line openTab call) but there is no test for Column Profiles, the only option with real branching (it calls openTab(DATASET_TAB) only when showColumnProfiles is currently false, otherwise it just toggles). That conditional is untested.
- **[minor/quality]** `dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx:14` onAddTab is now a dead prop: DataScreen.jsx renders `<WorkspaceTabBar />` without passing it and nothing else in the codebase supplies it either, but the prop, its JSDoc, and the `if (onAddTab) onAddTab();` call at line 147 are all still there.
- **[nit/quality]** `dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx:161` The dropdown sets role="menu" and role="menuitem" but only implements click-outside and Escape-to-close; there is no arrow-key navigation between items and no focus management (focus doesn't move into the menu on open or back to the + button on close), so it doesn't fully back up the ARIA menu semantics it declares.
- **[nit/style]** `dataloom-frontend/src/Components/workspace/features/dataset.tsx:6` This comment ("it's opened by the tab bar '+' and focused by the Profiling column-profiles toggle") is no longer accurate: the + button now opens the Profiling dropdown instead of calling openTab(DATASET_TAB) directly. It isn't part of this diff, but this PR is what made it stale.

**Conventions**

This PR does what issue #464 asked for: clicking the tab bar's "+" now opens a dropdown with the four Profiling actions (Summary, Column Profiles, Charts, Quality), and it also cleans up a tooltip's theming and z-index along the way. The real gap is that removing the old onAddTab handler leaves no direct way back to the DataSet tab once it's closed while other tabs stay open; none of the four options is "open DataSet," and Column Profiles only opens it as an undocumented side effect of turning profiles on. The dropdown also hand-duplicates action wiring that already lives elsewhere (the featureRegistry-driven ribbon items, and the Column Profiles toggle handler in MenuNavbar), so it would be worth deriving it from that instead of maintaining a second list. Test coverage is solid for three of the four options but skips the one with actual branching logic (Column Profiles).

YADUNANDAN-SINGH:
Thanks for the re-review — all six addressed.

**[major/quality] Duplicated action list.** Added
`src/Components/workspace/useProfilingMenu.ts`. `useProfilingMenuItems()`
derives the dropdown from `getFeatureMenu()` filtered to the Profiling ribbon,
so Summary/Charts/Quality come from their feature modules — no second list.
`useColumnProfilesAction()` holds the toggle once; `MenuNavbar` now calls it
instead of its own copy. Column Profiles is still spliced in at `order: 1`
rather than registered, since it drives hook state, not a declarative action.

**[major/bug] No way back to the DataSet tab.** The shared action now calls
`openTab(DATASET_TAB)` unconditionally before toggling, so one click always
restores the table. Two calls I'd rather flag than make silently:

- No fifth "open DataSet" entry — #464 specifies four options, and fixing
  Column Profiles removes the dead end within them. Say the word if you want it
  explicit.
- Ribbon behaviour shifts slightly: toggling profiles *off* now also focuses
  the DataSet tab. Consistent, but a change.

**[minor/test] Column Profiles untested.** Three tests (on, off, and reopening
a closed DataSet tab while profiles are on), plus one asserting the derived
menu order so the registry wiring itself is guarded. Also added a parameterised
test that the menu closes after each of the four options — #464 requires that
explicitly and nothing covered it.

**[minor/quality] Dead `onAddTab`.** Prop, JSDoc, interface, and call removed.

**[nit/quality] Unbacked ARIA menu.** Implemented the pattern: roving
`tabIndex`, focus into the menu on open, Arrow Up/Down with wrap, Home/End,
Escape closes and returns focus to "+", Tab dismisses, Arrow keys open from the
trigger. Six tests. Also fixed two leftovers from your earlier pass —
`aria-haspopup="true"` → `"menu"` and the stale `aria-label="New tab"`.

**[nit/style] Stale comment.** `features/dataset.tsx` updated.

ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `dataloom-frontend/src/Components/workspace/useProfilingMenu.ts:14` COLUMN_PROFILES_ORDER = 1 duplicates the 'order: 1' already hardcoded for the same Column Profiles item in MenuNavbar.jsx's coreItems array (that item is not sourced from getFeatureMenu() like every other Profiling entry). The doc comment two lines above (line 48) claims registry-driven items 'cannot drift apart', but this one still requires the two literals to be kept in sync by hand across two files; an independent edit to either will silently reorder the ribbon relative to the new dropdown.
- **[minor/test]** `dataloom-frontend/src/Components/workspace/WorkspaceTabBar.tsx:38` handleClickOutside (dismiss the dropdown on an outside mousedown) has no dedicated test. The suite thoroughly covers Escape-to-close, select-to-close, and keyboard navigation, but no test clicks outside the trigger/menu while it's open, so a regression in this specific branch would not be caught by CI.
- **[minor/quality]** `dataloom-frontend/src/Components/workspace/useProfilingMenu.ts:41` useColumnProfilesAction now unconditionally calls openTab(DATASET_TAB) on every toggle, including turning Column Profiles off (previously gated by `if (!showColumnProfiles)`, so turning off did not navigate). Since this hook also backs the pre-existing ribbon 'Column Profiles' button in MenuNavbar.jsx, it silently changes that existing entry point too: a user on another tab who just wants to switch the setting off is now always redirected to the DataSet tab. It reads as a deliberate, tested fix for a stuck-toggle edge case (per the comment above), but the behavior change to the existing ribbon control isn't mentioned in the PR description.

**Conventions**

This PR replaces the tab-bar '+' button's single click handler with an accessible dropdown (Summary, Column Profiles, Charts, Quality) sourced from the existing Profiling feature registry, plus a tooltip re-theme and a z-index fix; the accessibility work itself (roving tabindex, arrow/Home/End/Escape handling, focus return) is solid and well tested. One maintainability concern: useProfilingMenu.ts hardcodes COLUMN_PROFILES_ORDER = 1, duplicating a value already hardcoded in MenuNavbar.jsx, so the ribbon and dropdown can silently drift out of order if either is edited independently. The same refactor makes the shared toggle always navigate to the DataSet tab, including when turning Column Profiles off, which changes the pre-existing ribbon button's behavior in a way the PR description doesn't call out. Test coverage is otherwise thorough, though the new click-outside-to-close handler has no dedicated test.

PR #474, opened 2026-07-30
feat: reusable transformation pipelines

### LINKED ISSUE #473: [Feature] Reusable transformation pipelines

Users who clean similar CSV exports again and again must redo the same sequence of transforms by hand each time. DataLoom logs each transform in the project history, but there is no way to save that sequence and replay it on a different dataset. A common case is "clean this month's export the same way as last month's": today that means repeating every filter, sort, and cast step manually, which is slow and easy to get wrong.

Add reusable transformation pipelines.

- Let a user assemble an ordered set of transformation steps — picked from a project's change log, or built from scratch — and save it under a name.
- Let the user replay a saved pipeline on any of their own projects.
- Add a compatibility dry-run that replays the steps in memory and reports the first step that would fail (and why) before anything is applied.
- When a pipeline is applied, log each step to the normal project history, so save, undo, and checkpoint replay all keep working unchanged.
- A step stores the exact `action_type` + `action_details` of a history row and replays through the existing transformation registry, so pipelines add no new transform logic.

- Manual re-entry of each transform on every dataset (the status quo — slow and error-prone).
- A public pipeline gallery with sharing, `is_public`, and tags. Cut, because there is no multi-user or sharing model in the product yet (RBAC is an explicit non-goal).
- Storing per-operation "required columns" metadata to check compatibility. Rejected in favor of the in-memory dry run, which needs no metadata to maintain.

**Additional context**

- `addFile` steps cannot be part of a pipeline, because their stored inventory file is project-scoped.
- Pipelines are owner-scoped; other users' pipelines return an existence-hiding 404.
- Proposed shape: `Pipeline` + `PipelineStep` models; `/pipelines` endpoints for create, list, delete, check, and apply; a Pipelines tab plus a docked step builder on the frontend.
- Part of the GSoC roadmap (Week 10). A draft implementation lives on the `feat/transformation-pipelines` branch.

hanzalahwaheed:
- Pipeline + PipelineStep models with Alembic migration; each step stores `action_type`/`action_details` verbatim, the same pair a `user_logs` row holds
- pipeline_service: persist an ordered step list, dry-run compatibility check (reports the first failing step), and apply via the existing `apply_logged_transformation` registry replay — no new transform logic
- /pipelines endpoints: create, list, delete, check, check-steps, apply; all auth-gated with existence-hiding 404s
- applying a pipeline logs each step to `user_logs` so save, undo, and checkpoint replay keep working unchanged
- Pipelines tab + docked step builder: assemble a draft from logged steps or ones built from scratch (the transform forms are reused in "capture mode"), then reorder / check / name / save; list / check / apply / delete saved pipelines; a shared `PipelineDraftContext` bridges the two surfaces; registered on the Data ribbon (Data ▸ Pipelines)
- backend endpoint/service tests and lean component tests

## Description

Adds reusable transformation pipelines. A user assembles an ordered set of transformation steps once — picked from a project's change log or built from scratch — saves it under a name, and replays it on any of their own projects. A compatibility dry-run replays the steps in memory and reports the first step that would fail before anything is applied. Steps store the exact `action_type` + `action_details` of a history row and replay through the existing transformation registry, so pipelines add no new transform logic.

Closes #473.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Backend: `uv run pytest tests/test_pipelines.py` → **18 passed** (create/order/validation, list + delete, owner isolation 404s, dry-run compatibility, apply-and-log).
- Frontend: `npx vitest run` on the pipeline suites (`PipelineDraftContext`, `pipelineStepText`, `FilterFormCapture`) → **11 passed**.

- [x] Existing tests pass (pipeline suites; full suite runs in CI)
- [x] New tests added
- [ ] Manual testing

N/A

## Checklist

- [x] Tests pass locally


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/workspace/PipelinesTab.tsx:114` HistoryList (the 'From this project's history' step picker in the Pipelines tab) fetches logs once via getLogs and never reads useHistoryRefreshTokens()'s logsToken, so its useEffect has no way to learn that history changed. PipelineCard.handleApply calls refreshLogs() right after applying a pipeline (which logs every step), but since HistoryList never consumes the bumped logsToken, the picker stays stale until the tab is closed and reopened, even though HistoryRefreshContext exists specifically so components like this can stay in sync.
- **[major/bug]** `dataloom-backend/app/services/pipeline_service.py:111` check_steps_compatibility (backing POST /pipelines/check-steps) does not reject addFile steps the way create_pipeline_from_steps does a few lines above in the same file. HistoryList lets a user add any logged step, including an addFile entry, to the draft with no filtering, and apply_add_file only fails if the stored file becomes unreadable, so a draft containing a valid addFile step can pass 'Check against this project' as compatible, then unconditionally 400 on Save ('addFile steps cannot be included in a pipeline'), defeating the point of the eligibility check for that step.
- **[minor/test]** `dataloom-backend/tests/test_pipelines.py:236` TestAuthIsolation verifies cross-user 404s for list, delete, and apply, but has no case for POST /pipelines/{id}/check, the compatibility-check endpoint that relies on the same _get_owned_pipeline ownership guard and deserves the same coverage.

**Conventions**
- _Title, description, and the single commit message all follow Conventional Commits cleanly; the only nit is that a diff this large (backend, frontend, and tests together) landed in one commit, which a future PR this size might consider splitting for easier incremental review._

PR #474 adds a complete reusable transformation pipelines feature: new Pipeline and PipelineStep tables with a migration, a pipeline service covering create, list, delete, check, and apply, six new API endpoints, and a full frontend flow (a draft context, a step builder that reuses the twelve existing transform forms in capture mode, and a Pipelines tab with a saved pipeline library). Ownership checks are consistently applied and the test suite is solid (18 backend tests plus three frontend suites), but the history picker in the Pipelines tab never listens for the app's own history refresh signal, so it goes stale after a pipeline is applied while the tab stays open. There is also a validation gap between the save path, which explicitly rejects addFile steps, and the check-steps dry run path, which does not, letting a draft pass its compatibility check only to fail unconditionally at save. Neither issue is blocking, but both are worth fixing before merge.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/api/endpoints/pipelines.py:114` apply_pipeline's failure-recovery only reverts the on-disk file when log_transformation raises; it never undoes the log_transformation calls that already committed during earlier iterations of the for-loop over pipeline.steps (log_transformation calls db.commit() itself, per project_service.py, so each call is its own durable transaction). For any pipeline with 2+ steps, a failure on the 2nd or later step's log call leaves committed change-log rows for steps the file no longer reflects, since the file was just reverted to its pre-pipeline state. That is the exact partial-log inconsistency the surrounding comment says this code prevents. The only test for this path, test_apply_restores_file_if_log_transformation_fails, forces failure unconditionally, so it always fails on the first step and never exercises the partial-commit case (a 2-step pipeline where only the second log call fails).
- **[minor/quality]** `dataloom-backend/app/api/endpoints/pipelines.py:42` list_pipelines returns db.query(models.Pipeline)...all() without eager-loading the steps relationship. Since PipelineResponse serializes steps for every pipeline and Pipeline.steps is a default lazy SQLAlchemy relationship, listing a user's N pipelines triggers N additional per-pipeline queries. A selectinload(models.Pipeline.steps) would collapse this to two queries total.
- **[nit/quality]** `dataloom-frontend/src/Components/workspace/PipelinesTab.tsx:122` HistoryList's addLog adds any clicked change-log entry straight to the draft, including an addFile entry, even though both check-steps and create_pipeline always reject addFile steps. There is no inline signal at add-time, so a user who skips the explicitly optional Check against this project step only discovers the problem from a generic save-time toast, without being told which draft step is the offending one.

**Conventions**

PR #474 adds a complete, well-tested reusable-pipelines feature: new Pipeline/PipelineStep tables and migration, CRUD plus dry-run-compatibility and apply endpoints that replay steps through the existing transformation registry, and a frontend builder that captures steps from either the project's change log or the existing transform forms running in a new capture mode. The design is clean overall and reuses existing helpers correctly (the transformation error sanitizer was extracted into a shared module instead of duplicated), adding no new transform logic of its own. The main gap is in apply_pipeline's failure recovery: on a logging failure partway through a multi-step pipeline, it reverts the file but not the change-log rows already committed by earlier steps in the same request, leaving the audit trail inconsistent with the file, a scenario the existing test suite does not exercise. list_pipelines also does an unnecessary per-pipeline query for its steps that eager loading would avoid.

hanzalahwaheed:
@ivantha hey Oshan! please re-review this PR. its a big one!
thanks!

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-frontend/src/Components/pipelines/PipelineStepBuilderPanel.tsx:25` The pipeline step builder's "build a step from scratch" catalog is built via `getFeatureMenu().filter(item => item.ribbon === "Data" && item.group === "Transform")`, which only pulls ribbon items from the "Transform" group. But this same PR also wires AdvQueryFilterForm, PivotTableForm, and MeltForm (all registered under the "Data ▸ Query" group) for capture mode with the identical `onCapture`/`captureStep` pattern used by the 10 "Transform" forms. Because the catalog filter excludes the "Query" group, those three forms are never selectable from the step builder's operation list, so their capture-mode code can only ever fire if a user happens to already have a matching action in a project's log history to pick from HistoryList — "build from scratch" silently can't reach them. _(unverified)_
- **[major/bug]** `dataloom-backend/app/services/transformation_service.py:939` `TransformationSpec.reusable` defaults to `True` and only the `addFile` entry overrides it to `False` (reasoned as "the stored file belongs to one project"). But six other registry entries — addRow, delRow, addCol, delCol, changeCellValue, and renameCol — are equally project-bound: their `build_args` extract `row_index`/`col_index`/`col_index` positions (see e.g. change_cell_value and rename_column), and their underlying functions only raise when the index is out of range for the *target* DataFrame, not when it refers to something semantically different than it did in the source project. A pipeline step built from one of these (reachable via picking it off a project's history in HistoryList) can be saved and will report `compatible: true` during the dry-run check, then silently delete/rename/overwrite the wrong row or column when applied to an unrelated, differently-shaped project instead of failing loudly — undermining the exact safety guarantee the compatibility check is meant to provide.
- **[minor/test]** `dataloom-backend/tests/test_pipelines.py:270` The order assertion `assert sorted(log["action_type"] for log in logs) == ["filter", "sort"]` sorts both sides alphabetically before comparing, so it cannot detect a regression where the pipeline replays/logs steps out of order (e.g. sort-then-filter instead of filter-then-sort) — "filter" < "sort" alphabetically regardless of actual execution order. Asserting on the raw (unsorted) sequence, or on `id` order, would actually verify replay order.

**Conventions**
- **Commits (5):** clean, conventional history
- _Commit 5 ("fix: use an icon react-icons actually exports in DownloadImageButton") is an unrelated drive-by fix bundled into this feature branch; harmless, but a separate PR would have kept history cleaner._

This PR adds a complete, well-tested reusable-pipelines feature: new Pipeline/PipelineStep models with an Alembic migration, save/list/delete/check/apply endpoints backed by a pipeline_service that replays steps through the existing transformation registry, and a frontend Pipelines tab with a step builder, draft sequencer, and saved-pipeline library that reuses the existing transform forms in a new "capture mode." Two concrete issues are worth fixing before merge: the step builder's "build from scratch" catalog only pulls the ribbon's "Transform" group, leaving Advanced Query, Pivot Table, and Melt unreachable there despite being wired for capture mode; and the new `reusable` flag only excludes `addFile`, so six position-indexed operations (add/delete row, add/delete column, rename column, change cell value) stay "reusable" even though replaying them against an unrelated, differently-shaped project can silently touch the wrong row or column instead of failing the compatibility check. Aside from those two gaps, the auth/ownership scoping, the compensating-write logging helper, and the shared error-detail redactor all look solid and are exercised by the accompanying backend and frontend tests.

PR #482, opened 2026-08-11
feat: downloadable PDF project reports

### LINKED ISSUE #481: [Feature] Downloadable PDF project reports

- DataLoom works out plenty about a dataset — summary, column profiles, quality score, the transformations you applied — but all of it stays inside the app.
- Handing that work to someone else means exporting a CSV, which carries none of it, or screenshotting panels one at a time.

- A Report button that downloads the project as a PDF.
- Sections the user ticks: dataset overview, column profiles, data quality, provenance.
- Generated on the server from the same numbers the app shows, so the document can't quietly disagree with the screen.
- Generated on demand, not stored — the same call we made for quality assessment.
- Previewed in the app before download, so there are no surprises in the file.

- Printing a React page from the browser: most design freedom, but page breaks and headers stop being ours.
- WeasyPrint: real CSS, but it drags pango and cairo into the image — a lot of deployment risk for a document feature.
- Embedding charts: ours are recharts and client-side, so it needs either a server-side plotting library or images uploaded with the request. Leaving them out for now.

**Additional context**

- Keep "Report" (the document) and "Export" (the data file) as separate words in the UI and the code — users will otherwise conflate them.

hanzalahwaheed:
- `report_service.build_report()`: a reportlab PDF assembled from the sections the user chose — dataset overview, column profiles, quality, provenance — built from the same dicts that feed the in-app views, with user text escaped
- `GET /projects/{id}/report` streams the PDF as an attachment, generated on demand from the working copy; nothing is persisted
- column profiles print one row per column rather than one table each, so a wide dataset stays readable; quality names the detector settings it used, read off the service defaults
- provenance groups applied transformations under the checkpoint that saved them, after the source files, with unsaved work called out; `project_service.collect_provenance` assembles it from the change log
- `TransformationSpec` gains a `label`, so the server can name an operation without borrowing the frontend's `stepLabel()`
- Report tab previews the generated bytes with pdf.js rather than the browser's viewer, so pages render as sheets on the app background; fit to width with a debounced resize re-render, and the scroll position survives a rebuild
- `ReportConfigPanel` docks the section choices beside the preview; the panel is pinned open for the tab's lifetime, since a preview without the chooser and Download beside it is not usable
- a section change rebuilds the preview after a 600ms pause, sequenced so a slow earlier build cannot overwrite a newer one; a data change instead marks the preview stale and offers a rebuild
- renames the in-app quality report to a quality assessment, freeing "report" for the document the user downloads
- service, endpoint, context, and component tests

## Description

Adds downloadable PDF reports. A user opens `File ▸ Report`, chooses which sections the document carries, sees the generated PDF in the workspace, and downloads it. The dataset overview always prints; column profiles, data quality, and provenance are optional.

The preview is the document, not a rendering of it — the server builds the PDF, the tab draws those exact bytes, and Download saves the blob already on screen, so the file cannot drift from what the user approved. Reports are generated on demand from the working copy and never persisted, matching the on-demand-only decision for quality assessment.

Charts are deliberately out of scope: the app's charts are recharts and client-side, so embedding them would mean either a server-side plotting dependency or shipping images up with the request.

Closes #481.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Backend: `uv run pytest` → **744 passed**. New coverage in `tests/test_reports.py` spans each section on and off, provenance grouping (checkpointed vs unsaved vs unapplied), operation labels and parameter summaries, empty and column-less datasets, markup escaping, section query params, and auth/404 paths.
- Frontend: `npm test` → **423 passed** across 60 files. New suites cover the section request contract (including that clearing every section asks for the overview alone rather than defaulting to all), the preview's stale/rebuild and sequencing behaviour, and the config panel.
- `npm run lint` (ESLint + `tsc --noEmit`) and `ruff check` are clean.
- The rendered PDF was inspected page by page for each section combination while building the layout.

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

image

## Checklist

- [x] Tests pass locally


hanzalahwaheed:
@ivantha hi Oshan! please review

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `dataloom-backend/app/services/project_service.py:558` collect_provenance()'s "files" list (which feeds the Report's Provenance → Source files section) is built solely from the project_files table. Per ProjectFile's own class docstring, that table only records files added after the initial upload via the Add File operation — the project's own originating upload is never inserted into it (create_project(), used by POST /projects/upload, creates the Project row directly with no matching ProjectFile row). For any project that never used Add File — the common case — the generated report's Source files list is silently empty, so the Provenance section gives no indication of where the data came from at all.
- **[minor/quality]** `dataloom-backend/app/services/report_service.py:11` This module docstring, and PdfDocumentView.tsx:7, both cite docs/adr/0004-reports-are-server-rendered-pdfs.md as the design rationale for choosing server-rendered PDFs, but no docs/ directory exists anywhere in the repository — the ADR was never added in this PR. Both references are dangling for anyone who goes looking for the rationale.
- **[minor/test]** `dataloom-backend/tests/test_reports.py:51` TestBuildReport and the endpoint tests below it only assert that build_report() returns bytes starting with %PDF and that output length changes between section combinations; nothing extracts the rendered PDF text to check that specific content — a checkpoint message, a column name, or (per the finding above) an uploaded filename — actually appears on the page. A content-level assertion here would have caught the missing-source-files gap directly.

**Conventions**

This PR adds a well-built, thoroughly tested downloadable-PDF-report feature: report_service.py renders sections server-side from plain dicts shared with the in-app views, reproduces the app's brand colors and logo mark exactly, and the frontend draws the identical bytes it will later download via pdf.js rather than re-deriving a separate preview. All 744 backend and 423 frontend tests pass and both lint suites are clean, independently verified and matching the PR's own claims. The one real gap is that collect_provenance() sources the new Provenance section's "Source files" list only from files added after the initial upload, so for any project that never used Add File — the common case — the report silently shows no source file at all. Two code comments also point to a docs/adr/0004-reports-are-server-rendered-pdfs.md file that isn't part of this diff and doesn't exist anywhere in the repository.

PR #514, opened 2026-09-14
fix(frontend): clear table rows when a result set is empty

### LINKED ISSUE #510: [Bug] Table retains stale rows when results become empty
**Describe the bug**
The table keeps previous rows when project data becomes empty.

**To Reproduce**
1. Open a populated dataset.
2. Apply a filter that returns zero rows.
3. Previous rows remain visible.

**Expected behavior**
Show zero rows while retaining column headers.

**Screenshots**
N/A.

**Environment:**
React frontend; reproduced in a component test.

**Additional context**
Derive table data from project context instead of duplicating it in state.

iamtanishqjain:
## Description

Table kept its own copy of rows and columns in local state and filled it from project context inside an effect. The effect only ran when `ctxRows` was non-empty, so a filter or transform returning no rows skipped the update and the old rows stayed on screen.

Derived both from context instead, as suggested in the issue. Empty then falls out on its own and the headers stay. Ends up removing code rather than adding any.

`updateTableData` no longer sets the two removed pieces of state, so it resets the page right away instead of waiting for `refreshProject` to come back. All six callers already call `refreshProject(projectId, 1, pageSize)` immediately after, so page 1 is not a new behaviour, it just stops the row numbers briefly using the old page.

Fixes #510

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Added `src/test/Table.emptyResults.test.tsx`. It renders with rows, re-renders with none, and checks the old rows are gone and the headers are still there.

It fails on `main`:

```
expected document not to contain element, found <div ...>New York</div>
```

Full suite before and after is the same 19 pre-existing failures across 4 files, unrelated to the table. `tsc`, `eslint --max-warnings 0` and `prettier --check` are clean.

- [x] Existing tests pass
- [x] New tests added
- [ ] Manual testing

## Checklist

- [x] Tests pass locally

---

One thing I left alone: when the result is empty the body is just empty, no `EmptyState`. The issue asks for zero rows with the headers kept, and Table does not use `EmptyState` anywhere else, so adding it felt like a separate call. Happy to add it if you want it.


iamtanishqjain:
@ivantha PTAL

ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**

The PR replaces a useState+useEffect pattern in Table.tsx with derived useMemo values for `columns` and `data`, fixing the real bug: the old effect only updated local state when ctxRows was non-empty, so a filter/transform returning zero rows left stale rows on screen. The fix is a net code-reduction and is well-targeted at the described bug (issue #510). A new regression test (Table.emptyResults.test.tsx) directly verifies both that rows clear and that headers persist on an empty result set, and the PR notes it fails on main and passes with the change. One behavioral side effect the author calls out themselves: updateTableData now calls setPaginationData({page:1}) immediately instead of waiting for refreshProject, causing a brief page-1 flash in row numbers; worth a quick manual sanity check but not a blocker given all six callers already refresh right after.

Issue #504, opened 2026-09-01
[Bug] Column dtype not preserved after save/reload (CSV loses schema info)


SHIVAM-KUMAR-59:
**Describe the bug**
Casting a column to `string` and saving does not persist the dtype. After project reload, the column reports `float` instead of `str`. Root cause: CSV format stores values only, not schema. An empty/all-null string column writes to CSV with no distinguishing marker, so on reload `pd.read_csv()` re-infers dtype from raw values and an empty column comes back as `float64`.

**To Reproduce**

1. Create new column `new` (empty, all missing values)
2. Cast `new` to `string` type
3. Call transform API with `preview=false`, response correctly shows `"new": "str"`
4. Save succeeds
5. Call `GET /projects/get/{project_id}` (or refresh project)
6. Response now shows `"new": "float"`

**Expected behavior**
Dtype set by the user (`string`) should persist across save + reload. Column should still report `"new": "str"` after refresh.

**Screenshots**
N/A

**Environment:**
- OS: any
- Browser: any
- Node.js version: any
- Python version: any

**Additional context**
Traced the full pipeline. Frontend is clean (just displays what backend sends). `dataframe_to_response()` is clean (in-memory dataframe still shows `str` right after transform). The break happens specifically at the save → CSV → reload boundary: `save_table_safe()` writes CSV with no dtype metadata, and the next `read_table_safe()` call re-infers dtype from raw values via `pd.read_csv()`.

Not isolated to this column type, same failure mode can affect `boolean` (all-null) and `datetime` (CSV doesn't preserve datetime dtype natively; `_infer_datetime_columns()` exists but is inference, not persistence).

Related: issue #296 (column type metadata system) covers similar ground, previously closed due to GSoC 2026 scope overlap with #76. Both are now closed/wrapped, flagging for context.

Root cause fully traced; happy to take this on if assigned.

SHIVAM-KUMAR-59:
I'd like to work on this, waiting for @ivantha 's approval.

PR #517, opened 2026-09-21
perf(backend): sampled pre-gate for datetime inference

### LINKED ISSUE #515: [Feature] Skip the full-column datetime scan for obviously non-date columns
`_infer_datetime_columns` (`app/utils/pandas_helpers.py`) runs inside `read_table_safe`, which backs `read_project_df` and about ten other call sites, so it runs on every dataset read, not just on upload. For each object column it runs `isinstance`, `.strip()` and up to four regex full-matches on every value before it can reject the column. For names, notes and IDs, that whole walk is wasted and the column is discarded at the 0.8 date-rate check.

Timed on `main` (single column, best of 3, Python 3.13, pandas 3.0.1):

| Column | Rows | Time |
|--------|-----:|-----:|
| Free text | 1,000,000 | 0.481 s |
| Order IDs | 1,000,000 | 0.485 s |
| Real dates (`YYYY-MM-DD`) | 1,000,000 | 0.683 s |
| Free text | 100,000 | 0.052 s |

A 1M-row dataset with ten text columns spends ~5 s in datetime inference on every read, without changing the result.

Pre-screen each column larger than 1,000 values on a fixed-seed random sample of 1,000 non-null values. The gate can only reject; conversion stays with the existing full-column check. It rejects when:

1. any sampled value is not a `str` (the mixed-type guard would refuse the column anyway),
2. the sampled date-like rate is below 0.5 (conversion needs ≥ 0.8; at n = 1,000 that is ~24 standard errors away),
3. the sample shows both DD/MM and MM/DD (`_infer_dayfirst` would return `None` for the full column too).

Columns at or below 1,000 values skip the gate and behave exactly as today. Random rather than head-based so sorted or appended data isn't misjudged; seeded so the same file always gets the same dtypes.

- Head-based sample: wrongly rejects columns whose first rows aren't representative.
- Persisting dtypes (#504): cuts how often inference runs, but first load and upload still pay the cost, so the two are complementary.

**Additional context**
Performance-only: ~60 lines in `pandas_helpers.py` plus tests in `tests/test_dtypes.py`. No threshold changes; the `quality_service.py` `to_datetime` calls and speeding up real date columns are out of scope. Checking first that this doesn't overlap reserved scope before opening the PR.

YADUNANDAN-SINGH:
## Description

`_infer_datetime_columns` runs inside `read_table_safe`, so it executes on every
dataset read, not just on upload. For each object column it runs `isinstance`,
`.strip()` and up to four regex full-matches on **every value** before it can
reject anything. For a column that is plainly not dates, that whole walk is
discarded at the 0.8 date-rate check.

A 1M-row dataset with ten ordinary text columns spends ~5 s in datetime
inference on every request, without changing the result.

Fixes #515 

## Changes

**Backend** (`app/utils/pandas_helpers.py`): added a private helper,
`_sample_rules_out_datetime`, which pre-screens columns larger than 1,000
values on a fixed-seed random sample of 1,000 non-null values. It is called at
one place in `_infer_datetime_columns`, right after the `non_null.empty` check.
Nothing below that point changed.

The gate **can only reject**. It answers "definitely not a date column" or
"maybe", and never converts a column itself. Conversion stays entirely with
the existing full-column check. It rejects when:

| Check | Kind | Why it can't wrongly reject |
|-------|------|-----------------------------|
| A sampled value is not a `str` | exact | The full path's mixed-type guard refuses any column containing a non-string |
| Sampled date-like rate < 0.5 | statistical | Conversion needs ≥ 0.8; at n = 1,000 the standard error is ~1.3 pp, so < 0.5 is ~24 SE away |
| Sample shows both DD/MM and MM/DD | exact | `_infer_dayfirst` returns `None` for the full column too, so it is skipped there as well |

The sample is random, so sorted or appended data isn't misjudged, and it is
seeded, so the same file always yields the same dtypes. Columns of ≤ 1,000
values return immediately and take today's path unchanged.

The two inline `0.8` literals are now one named constant, `_MIN_DATETIME_RATE`,
because the gate's margin is defined relative to it and the two must not drift
apart.

**Benchmarks**: `_infer_datetime_columns` timed directly, single-column
frames, `upstream/main` vs this branch. Best of 3 runs on Apple M5, Python
3.13, pandas 3.0.1.

| Column | Rows | `main` | This branch | Speedup | Dtype |
|--------|-----:|-------:|------------:|--------:|-------|
| Free text | 1,000,000 | 0.481 s | 0.007 s | 69× | `str` both |
| Order IDs | 1,000,000 | 0.485 s | 0.007 s | 69× | `str` both |
| Real dates (`YYYY-MM-DD`) | 1,000,000 | 0.683 s | 0.701 s | 0.97× (noise) | `datetime64[us]` both |
| Free text | 100,000 | 0.052 s | 0.002 s | 26× | `str` both |

The gate costs ~7 ms per column. Real date columns are unaffected, because
their cost is `pd.to_datetime(format="mixed")`, which is out of scope here.

**Out of scope:** the `to_datetime` calls in `quality_service.py`, any
threshold change, and persisting dtypes (#504), which is complementary since
first load and upload still pay this cost.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement (no behaviour change)

## How Has This Been Tested?

- [x] Existing tests pass: `uv run pytest` (800 passed). Every existing
  `TestInferDatetimeColumns` case uses a handful of rows, so they exercise the
  unchanged path.
- [x] New tests added in `tests/test_dtypes.py`, 13 in total, all on 5,000-row
  columns so the gate is actually hit:
  * `TestSampledPreGate`, end-to-end through `_infer_datetime_columns`:
    large text stays text; large dates convert; exactly 80% dates (shuffled)
    still converts; 79% stays text; 1,000 notes followed by 4,000 dates converts
    (proves the sample isn't head-based); a mixed-type column stays object; mixed
    DD/MM + MM/DD stays text; repeated inference gives identical dtypes
  * `TestSampleRulesOutDatetime`, the helper directly: `False` at ≤ 1,000
    values, `True` for text, `False` for dates, `True` with non-strings, `True`
    on conflicting conventions
- [x] Equivalence sweep (throwaway script, not committed): 300 generated
  columns of 1,001–6,000 rows, date shares 0–100% including exactly 0.5 and
  0.8, 7 date formats, 7 junk types, shuffled / dates-first / junk-first
  layouts, with nulls and non-strings mixed in. Gate on vs. off gave identical
  dtypes and values on all 300.
- [x] Manual verification: `uv run ruff check .` and `uv run ruff format --check .`
  are clean. The frontend is untouched but was verified green anyway (ESLint,
  Prettier, `tsc --noEmit`, build, 475 vitest tests), and the Playwright E2E
  suite was run locally.

No timing assertions in tests, since CI timing is noisy. Speed is shown by the
benchmark and correctness by the tests.

N/A. This is a backend-only performance change with no visual change.

## Checklist

- [x] Tests pass locally


ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**

This PR adds a statistical pre-gate (fixed-seed 1,000-row sample) that lets `_infer_datetime_columns` cheaply skip columns on large DataFrames before running the full per-value scan, without changing the existing conversion rules or threshold logic. The one-sided design (the gate can only cause an early skip, never an early conversion) is sound, and the documented margin between the gate's 0.5 rejection threshold and the 0.8 conversion threshold gives a comfortably wide safety margin against sampling error. Test coverage is thorough: 13 new tests exercise the exact-threshold boundary, non-head-based ordering, mixed types, conflicting date conventions, and gate-on/off equivalence. No correctness, security, or quality issues were found in the diff.

